PR #101, opened 2021-03-17
Metric bootstrapper

### LINKED ISSUE #96: Bootstrap wrapper for metrics?
## 🚀 Feature

We should provide ability to compute bootstrapped confidence intervals for metrics.

### Motivation

Confidence intervals are important and we should make it easy for people to increase rigor of their research and model evaluations.

### Pitch

I'm thinking we can have something like this (very high level):
```python
class Bootstrapper(Metric):
   def __init__(self, num_samples, metric):
       self.metrics = nn.ModuleList([deepcopy(metric) for _ in range(num_samples)])

  def update(self, preds, targets):
     for idx in range(self.num_samples):
        preds_sampled, targets_sampled = sample_for_bootstrap(preds, targets)
        self.metrics[i].update(preds_sampled, targets_sampled)
```

which will let people to wrap any metric, have a set of copies of the metric internally updated with different samples of the data, giving us then ability to get a distribution of metric values.

### Alternatives

We can skip it on the class-based metrics side and assume anyone doing bootstrap will load everything in memory and do bootstrap using functional metrics.

SkafteNicki:
# Before submitting

## What does this PR do?
Fixes #96 
Implements a wrapper metric that allows the user to get bootstrapped metric values easily:
```python
metric = BootStrapper(Accuracy(), num_bootstraps=10)
metric.update(preds, target)
mean, std = metric.compute()
```
where mean, std will be the mean and standard deviation calculated over 10 different metrics that have each been updated with resampled (with replacement) versions of `preds, target`.
Efficient way to determine confidence interval for metric values.

## PR review    

## Did you have fun?


docs/source/references/modules.rst
```diff
@@ -269,4 +269,15 @@ R2Score
...
 .. autoclass:: torchmetrics.R2Score
-    :noindex:
\ No newline at end of file
+    :noindex:
+    
+********
+Wrappers
+********
+
+Modular wrapper metrics are not metrics in themself, but instead the take in other metrics and alter
```
tchaton, line 278:
```suggestion
Modular wrapper metrics are not metrics in themself, but instead take a metric and alter 
```

tchaton (approved):
LGTM ! Great addition !

torchmetrics/utilities/__init__.py
```diff
@@ -1,3 +1,4 @@
 from torchmetrics.utilities.data import apply_to_collection  # noqa: F401
 from torchmetrics.utilities.distributed import class_reduce, reduce  # noqa: F401
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_6, _TORCH_GREATER_EQUAL_1_7  # noqa: F401
```
Borda, line 3:
```suggestion
```
lets do not expose them if they are protected

torchmetrics/wrappers/bootstrapping.py
```diff
@@ -0,0 +1,161 @@
...
+            process_group:
+                Specify the process group on which synchronization is called.
+                default: ``None`` (which selects the entire world)
+            dist_sync_fn:
+                Callback that performs the allgather operation on the metric state. When ``None``, DDP
+                will be used to perform the allgather.
+
+        Example::
+            >>> from torchmetrics.wrappers import BootStrapper
+            >>> from torchmetrics import Accuracy
```
Borda, line 100:
```suggestion
            >>> from torchmetrics import Accuracy, BootStrapper
```

Borda (commented):
@justusschock mind check :]

teddykoker (approved):
LGTM

torchmetrics/wrappers/bootstrapping.py
```diff
@@ -0,0 +1,156 @@
...
+        generator: a instance of ``torch.Generator`` that controls the sampling
+
+    Returns:
+        resampled tensor
+
+    """
+    if size is None:
+        size = tensor.shape[0]
+    idx = torch.multinomial(
+        torch.ones(tensor.shape[0], device=tensor.device),
```
maximsch2, line 41: I'm actually not sure if this is completely correct. For real bootstrap, we want to resample with replacement over the entire dataset, not on the batch level. After the correct bootstrap, it's possible that elements from this batch are not present in the result at all or present with more than 1 occurence.

I suspect the right formula will be attempting to model the number of times each individual sample needs to be present [0, inf)

SkafteNicki: @maximsch2 thanks for raising this issue, I had not thought about it. I think it will be hard to implement "real" bootstrapping in batched version as we have no idea when to stop sampling from the current batch. However, I will try to think a bit more about this.
Else it should be stated in the docstring that this only perform "pseudo"-bootstrapping.

SkafteNicki: @maximsch2 I added `sampling_strategy` argument and default to doing `poisson` sampling. As described here https://en.wikipedia.org/wiki/Bootstrapping_(statistics)#Poisson_bootstrap sampling using `Poisson(1)` to determine the number of times each sample should be included, will approximate the true bootstrapped distribution in the limit.

maximsch2: @SkafteNicki, this looks like the right approach, thanks for adding this!

tests/wrappers/test_bootstrapping.py
```diff
@@ -0,0 +1,113 @@
...
+@pytest.mark.parametrize(
+    "metric, sk_metric", [[Precision(average='micro'), precision_score], [Recall(average='micro'), recall_score]]
+)
+def test_bootstrap(sampling_strategy, metric, sk_metric):
+    """ Test that the different bootstraps gets updated as we expected and that the compute method works """
+    _kwargs = {'base_metric': metric, 'mean': True, 'std': True, 'raw': True, 'sampling_strategy': sampling_strategy}
+    if _TORCH_GREATER_EQUAL_1_7:
+        bootstrapper = TestBootStrapper(**_kwargs, quantile=torch.tensor([0.05, 0.95]))
+    else:
+        bootstrapper = TestBootStrapper(**_kwargs)
```
Borda, line 85:
```suggestion
    if _TORCH_GREATER_EQUAL_1_7:
        _kwargs.update(dict(quantile=torch.tensor([0.05, 0.95])))

    bootstrapper = TestBootStrapper(**_kwargs)
```

tests/wrappers/test_bootstrapping.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    def update(self, *args) -> None:
+        self.out = []
+        for idx in range(self.num_bootstraps):
+            size = len(args[0])
+            sample_idx = _bootstrap_sampler(size, sampling_strategy=self.sampling_strategy)
+            new_args = apply_to_collection(args, Tensor, torch.index_select, dim=0, index=sample_idx)
+            self.metrics[idx].update(*new_args)
+            self.out.append(new_args)
+
```
Borda, line 41:
```suggestion

def _check_samples_twice(...):
    # make sure some samples are also sampled twice
    found_one = False
    for os in old_samples:
        cond = os == new_samples
        if cond.sum() > 2:
            found_one = True
            break

```
and use bellow

torchmetrics/wrappers/bootstrapping.py
```diff
@@ -0,0 +1,174 @@
...
+        output = []
+        if self.mean:
+            output.append(computed_vals.mean(dim=0))
+        if self.std:
+            output.append(computed_vals.std(dim=0))
+        if self.quantile is not None:
+            output.append(torch.quantile(computed_vals, self.quantile))
+        if self.raw:
+            output.append(computed_vals)
+        return output
```
Borda, line 174: this variant list can be tricky for interpretation, I would consider fix length and if some are not computed, leave it as None or NaN

torchmetrics/utilities/imports.py
```diff
@@ -11,11 +11,13 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Import utilities"""
+import operator
 from distutils.version import LooseVersion
 from importlib import import_module
 from importlib.util import find_spec
 
-import torch
+import torch  # noqa: F401
```
Borda, line 20:
```suggestion
```
seems unused

torchmetrics/wrappers/bootstrapping.py
Borda: or return as a dictionary so the meaning also clear.. or is there a reason for this floating length?

Borda (approved):

PR #128, opened 2021-03-24
Binned PR-related metrics

### LINKED ISSUE #95: Constant-memory implementation of precision-recall related metrics
## 🚀 Feature

Metrics that depend on precision-recall curve are currently implemented in a way that requires storing all of the predictions and labels in memory, making it's use impractical for large datasets or problem with large label spaces. We should support binning-based metrics implementation to solve this. Prototype is here: https://gist.github.com/maximsch2/2b55bab6deba629a5686258cb8152e53

### Alternatives

Don't do anything and be restricted in the scalability of metrics.

Other options for scaling is making it easier to keep metrics off-GPU.

A possible question is if we want to have both raw and binned implementations of the metrics.

### Additional context

Keras provides binning-based implementation by default: https://keras.io/api/metrics/classification_metrics/#auc-class

maximsch2:
# Before submitting

## What does this PR do?
Fixes #95 .

## PR review    

## Did you have fun?


Borda:
@maximsch2 how is it going here, ready for review?

maximsch2:
@Borda, I need to fix up isort and do a short cleanup, will get to it some time this week.

Borda:
@maximsch2 how is the clean-up doing? ;]

maximsch2:
@Borda , I've moved out the function I wanted to do, feel free to take a look now. Another thing I think I should probably do is to move out test data generation code into a central place, but actually curious about your feedback to see if we should just replace the main one with it? 

The motivation is to get a reasonably-looking model, so that metrics like recall@precision=0.8 can be computed as the random data never actually reaches that precision :)

maximsch2:
@Borda , mind taking a look now?

tests/classification/inputs.py
```diff
@@ -77,3 +77,29 @@
     preds=torch.randint(high=NUM_CLASSES, size=(NUM_BATCHES, BATCH_SIZE, EXTRA_DIM)),
     target=torch.randint(high=NUM_CLASSES, size=(NUM_BATCHES, BATCH_SIZE, EXTRA_DIM))
 )
+
+
+# Generate plausible-looking inputs
+def generate_plausible_inputs_multilabel():
```
Borda, line 83: lets rather have all the constants as arguments with defaults

tests/classification/test_binned_precision_recall.py
```diff
@@ -0,0 +1,162 @@
...
+import pytest
+import torch
+from sklearn.metrics import average_precision_score as _sk_average_precision_score
+from sklearn.metrics import precision_recall_curve as _sk_precision_recall_curve
+
+from tests.classification.inputs import (
+    _input_binary_prob,
+    _input_binary_prob_plausible,
+    _input_multilabel_prob,
+    _input_multilabel_prob_plausible,
```
Borda, line 28:
```suggestion
    _input_binary_prob,
    _input_binary_prob_plausible as _input_binary_prob_ok,
    _input_multilabel_prob as _input_mlb_prob,
    _input_multilabel_prob_plausible as _input_mlb_prob_ok,
```
let's shorten these imports or if you have better shortcuts :]

Borda (commented):
it mainly requires docs work:
- [ ] ref to the metric definition
- [ ] argument/options desc
- [ ] examples

maximsch2:
Sounds good, let me work on that. While I started I've noticed another clean-up opportunity. Also, what's up with circleci? I see bleu failing intermittently even though I'm not touching it

maximsch2:
> it mainly requires docs work:
> 
> * [x]  ref to the metric definition
> * [x]  argument/options desc
> * [x]  examples

One tricky thing is bleu tests, I've added atol there to make them pass stably as they are pretty close anyway. Please take another look.

tests/classification/test_precision_recall.py
```diff
@@ -195,7 +195,7 @@ def test_no_support(metric_class, metric_fn):
 )
 class TestPrecisionRecall(MetricTester):
 
-    @pytest.mark.parametrize("ddp", [False])
+    @pytest.mark.parametrize("ddp", [False, True])
```
justusschock, line 198: @SkafteNicki @Borda do we have multi-gpu tests here to run ddp?

justusschock (approved):

tests/classification/test_precision_recall.py
Borda: yes, we run on the very same cluster as PL

Borda: TM: https://github.com/PyTorchLightning/metrics/blob/935994c8d56a6c9b2c97db6b86dcd224907b55e4/azure-pipelines.yml#L26

PL: https://github.com/PyTorchLightning/pytorch-lightning/blob/aaccbeea2beb3b903428643d2ee520d2dd05b348/azure-pipelines.yml#L26

tests/classification/test_binned_precision_recall.py
```diff
@@ -0,0 +1,163 @@
...
+            NUM_CLASSES,
+        ),
+    ],
+)
+class TestBinnedRecallAtPrecision(MetricTester):
+    @pytest.mark.parametrize("ddp", [True, False])
+    @pytest.mark.parametrize("dist_sync_on_step", [True, False])
+    @pytest.mark.parametrize("min_precision", [0.05, 0.1, 0.3, 0.5, 0.8, 0.95])
+    def test_binned_pr(self, preds, target, sk_metric, num_classes, ddp, dist_sync_on_step, min_precision):
+        self.atol = 0.02
```
SkafteNicki, line 100: Move outside function like here:
https://github.com/PyTorchLightning/metrics/blob/ec239e00ebfad96fc9a0499c7bcd0d74e88dbffa/tests/regression/test_ssim.py#L69-L70
```suggestion

```

docs/source/references/modules.rst
```diff
@@ -126,6 +126,12 @@ AveragePrecision
 .. autoclass:: torchmetrics.AveragePrecision
     :noindex:
 
+BinnedAveragePrecision
```
SkafteNicki, line 129: insert in alphabetical order

docs/source/references/modules.rst
```diff
@@ -198,6 +204,12 @@ PrecisionRecallCurve
 .. autoclass:: torchmetrics.PrecisionRecallCurve
     :noindex:
 
+BinnedPrecisionRecallCurve
```
SkafteNicki, line 207: insert in alphabetical order

docs/source/references/modules.rst
```diff
@@ -198,6 +204,12 @@ PrecisionRecallCurve
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: torchmetrics.BinnedPrecisionRecallCurve
+    :noindex:
+
```
SkafteNicki, line 212: seems like `BinnedRecallAtFixedPrecision` is missing

SkafteNicki (commented):
smaller comments

maximsch2:
Addressed comments. On docs, is your preference to go for strict alphabetical order? I was putting the binned versions close to non-binned ones so that people will notice them together, but I can move the new one to the to the sorted place as well.

Borda:
> I was putting the binned versions close to non-binned ones so that people will notice them together, but I can move the new one to the to the sorted place as well.

that is also an option, I do not have strong preference here :]

torchmetrics/classification/binned_precision_recall.py
```diff
@@ -0,0 +1,303 @@
...
+        tensor([0.0000, 0.5000, 1.0000]),
+        tensor([0.0000, 0.5000, 1.0000]),
+        tensor([0.0000, 0.5000, 1.0000]),
+        tensor([0.0000, 0.5000, 1.0000])]
+    """
+
+    TPs: Tensor
+    FPs: Tensor
+    FNs: Tensor
+    thresholds: Tensor
```
Borda, line 109: is this a counter?
if I have one metric for train other for valid it would mix all together, right?

Borda (changes requested):

torchmetrics/classification/binned_precision_recall.py
maximsch2: This is just type annotation, it doesn't define a new variable and we use instance variables so there won't be overlap.

The cause here is the confusing way of Python type annotation for instance variables:
```python
class A:
   x: int # instance variable
   y: ClassVar[int] # class/static variable
```

https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations

torchmetrics/classification/binned_precision_recall.py
```diff
@@ -0,0 +1,303 @@
...
+            dist_sync_on_step=dist_sync_on_step,
+            process_group=process_group,
+        )
+
+        self.num_classes = num_classes
+        self.num_thresholds = num_thresholds
+        thresholds = torch.linspace(0, 1.0, num_thresholds)
+        self.register_buffer("thresholds", thresholds)
+
+        for name in ("TPs", "FPs", "FNs"):
```
Borda, line 130: ok, see here...

torchmetrics/classification/binned_precision_recall.py
Borda: ok, I see not it is just kind or shadow to make IDE happy, right? https://github.com/PyTorchLightning/metrics/pull/128/files#r612738983
well, I find it confusing... :/
other thoughts? @PyTorchLightning/core-metrics

maximsch2: Yeah, it's for type-checker, I do agree that given that we don't have that in other metrics anyway it's fine to remove from here as well.

Borda: yes, so let's make this PR without it and then open separate PR just by introducing this typing for all metrics... ok?

Borda (approved):

PR #138, opened 2021-03-26
Add AverageMeter implementation

### LINKED ISSUE #129: Include AverageMeter?
One common pattern I've seen copy-pasted across many different projects is a generic `AverageMeter`, which takes the average of things of a quantity. 

This isn't strictly a "metric", but I'm wondering whether you'd be open to having an implementation in this metrics repository -- it's quite common and having it in a centralized place could be helpful. If you are open to it, I'd be happy to contribute an implementation.

```python
class AverageMeter(object):
    """Computes and stores the average and current value"""

    def __init__(self, name, fmt=':f'):
        self.name = name
        self.fmt = fmt
        self.reset()

    def reset(self):
        self.val = 0.
        self.avg = 0.
        self.sum = 0.0
        self.count = 0

    def update(self, val, n=1):
        self.val = float(val.item()) if isinstance(
            val, (np.ndarray, torch.Tensor)) else val
        self.sum += val * n
        self.count += n
        self.avg = self.sum / self.count

    def __str__(self):
        fmtstr = '{name} {val' + self.fmt + '} ({avg' + self.fmt + '})'
        return fmtstr.format(**self.__dict__)
```

alanhdu:
This adds an `AverageMeter`, which is a simple metric that takes the average of a stream of values.

Closes #129


alanhdu:
I am currently struggling a bit with the test suite -- many of them seem to have hardcoded an interface where metrics operate over `predictions` and `target`, which isn't applicable here.  I will continue to study the tests and see how to add it.

Borda:
> I am currently struggling a bit with the test suite

ping us in slack if you need our help :]

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
+from typing import Union
```
SkafteNicki, line 1: include our copyright info at top of file (just copy from another file)

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
...
+
+import torch
+
+from .metric import Metric
```
SkafteNicki, line 5:
```suggestion
from torchmetrics.metric import Metric
```

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
+from typing import Union
+
+import torch
+
```
SkafteNicki, line 4:
```suggestion
from torch import Tensor
```

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
...
+    Forward accepts
+    - ``value`` (scalar): 
+    - ``weight`` (scalar): the weight to give this observation.
+    """
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.add_state("value", torch.zeros(1), dist_reduce_fx="sum")
+        self.add_state("weight", torch.zeros(1), dist_reduce_fx="sum")
+
+    def update(self, value: float, weight: float = 1) -> None:
```
SkafteNicki, line 19: since torchmetrics is intended to be used with pytorch, I definitely think that the value should be a `torch.Tensor`. weight can be both tensor and float in my opinion
```suggestion
    def update(self, value: Tensor, weight: Union[float, Tensor] = 1.0) -> None:
```

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
...
+        self.value += value
+        self.weight += weight
+
+    def compute(self) -> torch.Tensor:
```
SkafteNicki, line 23:
```suggestion
    def compute(self) -> torch.Tensor:
        """ return averaged value """
```

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
...
+from .metric import Metric
+
+class AverageMeter(Metric):
+    """Computes the average of a stream of values.
+
+    Forward accepts
+    - ``value`` (scalar): 
+    - ``weight`` (scalar): the weight to give this observation.
+    """
+    def __init__(self, *args, **kwargs) -> None:
```
SkafteNicki, line 14: please be explicit about these args
```suggestion
    def __init__(self, 
        compute_on_step: bool = True,
        dist_sync_on_step: bool = False,
        process_group: Optional[Any] = None,
        dist_sync_fn: Callable = None,) -> None:
```

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
+from typing import Union
+
+import torch
+
+from .metric import Metric
+
+class AverageMeter(Metric):
+    """Computes the average of a stream of values.
+
```
SkafteNicki, line 9: probably need a bit of formatting
```suggestion

Args:
            compute_on_step:
            Forward only calls ``update()`` and return ``None`` if this is set to ``False``.
        dist_sync_on_step:
            Synchronize metric state across processes at each ``forward()``
            before returning the value at the step
        process_group:
            Specify the process group on which synchronization is called.
            default: ``None`` (which selects the entire world)
        dist_sync_fn:
            Callback that performs the allgather operation on the metric state. When ``None``, DDP
            will be used to perform the allgather

```

torchmetrics/average.py
```diff
@@ -0,0 +1,24 @@
...
+    Forward accepts
+    - ``value`` (scalar): 
+    - ``weight`` (scalar): the weight to give this observation.
+    """
```
SkafteNicki, line 13: could you insert an example at the end of the docstring

SkafteNicki (commented):
For testing, I think it should be quite easily to check that this works as intended by using our current testing interface and just compare against `np.average`.

Also missing:
* add instance to changelog
* add ref in the `docs/source/references/modules.rst`

alanhdu:
Find some time to come back to this. I made a couple different changes:
- I believe I responded to all the style requests
- I changed the interface slightly to allow `AverageMeter` to take in Tensor values/weights along with scalar ones. If it has a tensor value, then we treat each item of the Tensor as a separate observation. Weights is automatically broadcasted to fit the `values`
- I have added tests, although I had to abuse some of the name slightly.

tests/test_average.py
```diff
@@ -0,0 +1,43 @@
...
+    @pytest.mark.parametrize("dist_sync_on_step", [False, True])
+    def test_average_fn(self, ddp, dist_sync_on_step, values, weights):
+        self.run_class_metric_test(
+            ddp=ddp,
+            dist_sync_on_step=dist_sync_on_step,
+            metric_class=AverageMeter,
+            sk_metric=average,
+            # Abuse of names here
+            preds=values,
+            target=weights,
```
alanhdu, line 32: I wasn't really sure what to do here -- looking at the code, it looks like these are special-cased as positional arguments (which is why the differing names don't matter), and that we partition the data across processes using these.

torchmetrics/average.py
```diff
@@ -0,0 +1,101 @@
...
+            process_group=process_group,
+            dist_sync_fn=dist_sync_fn,
+        )
+        self.add_state("value", torch.zeros(()), dist_reduce_fx="sum")
+        self.add_state("weight", torch.zeros(()), dist_reduce_fx="sum")
+
+    # TODO: these should be Union[Tensor, float], but Unions are not picklable
+    # in Python 3.6. https://github.com/python/cpython/pull/6216 fixed this in
+    # Python 3.7+
+    def update(self, value: Tensor, weight: Tensor = 1.0) -> None:
```
alanhdu, line 82: I'm not sure how you want to track this, so I just left it as a code comment for now.

alanhdu:
Intersting... it looks like `torch.broadcast_to` was only added in PyTorch 1.8. Should I add a version check here, or just revert that behavior (and require `weights` to either be a scalar or the same shape as `values`)?

torchmetrics/average.py
maximsch2: Why is pickleability an issue for type annotations? You can probably do `from __future__ import annotations` and they become strings.

torchmetrics/average.py
```diff
@@ -0,0 +1,101 @@
...
+        """Updates the average with.
+
+        Args:
+            value: A tensor of observations (can also be a scalar value)
+            weight: The weight of each observation (automatically broadcasted
+                to fit ``value``)
+        """
+        if not isinstance(value, Tensor):
+            value = torch.as_tensor(value, dtype=torch.float32)
```
maximsch2, line 91: Here you might need to specify a device as well, otherwise it won't work on the GPU. 
Maybe do `dtype = self.value.dtype, device = self.value.device` (and same for weight below).

tests/test_average.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def average(values, weights):
+    return np.average(values, weights=weights)
+
+
+@pytest.mark.parametrize(
+    "values, weights",
+    [
+        (torch.rand(NUM_BATCHES, BATCH_SIZE), torch.ones(NUM_BATCHES, BATCH_SIZE)),
```
maximsch2, line 16: Given that you are supporting both tensors and non-tensors, better test it as well here - e.g. it should catch the `device` issue I'm pointing out when run on GPUs.

torchmetrics/average.py
justusschock: Otherwise you could probably make them strings manually (not shure how this works out with nested typing though)

alanhdu: I think strings manually works here: `from __future__ import annotations` doesn't exist yet on Python 3.6

alanhdu:
I believe I've responded to the code review comments. The tests for the scalar case are a little messy, but I think they should be workable.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added testing for `half` precision ([#77](https://github.com/PyTorchLightning/metrics/pull/77),
     [#135](https://github.com/PyTorchLightning/metrics/pull/135)
 )
+- Added `BootStrapper` to easily calculate confidence intervals for metrics ([#101](https://github.com/PyTorchLightning/metrics/pull/101))
+- Added `__getitem__` as metric arithmetic operation ([#142](https://github.com/PyTorchLightning/metrics/pull/142))
```
SkafteNicki, line 30: duplicated
```suggestion
```

tests/test_average.py
SkafteNicki: looks fine to me :]

SkafteNicki: looks fine to me :]

SkafteNicki (commented):
PR is looking good, if you could look into failing test that would be great :]

It seems something is wrong with the type checking:
`Cannot resolve forward reference in type annotations of "torchmetrics.AverageMeter.update": name 'Union' is not defined`

tests/test_average.py
SkafteNicki: looks fine to me :]

Borda:
@alanhdu @SkafteNicki how are we going here? 🐰

SkafteNicki (approved):

torchmetrics/average.py
```diff
@@ -0,0 +1,111 @@
...
+            dist_sync_fn=dist_sync_fn,
+        )
+        self.add_state("value", torch.zeros(()), dist_reduce_fx="sum")
+        self.add_state("weight", torch.zeros(()), dist_reduce_fx="sum")
+
+    # TODO: need to be strings because Unions are not pickleable in Python 3.6
+    def update(  # type: ignore
+        self,
+        value: "typing.Union[Tensor, float]",
+        weight: "typing.Union[Tensor, float]" = 1.0
```
Borda, line 84: not very happy about this, can we get some better way?

SkafteNicki: it is changed to just `Union[Tensor, float]` now, is that better?
else we can just go with `Any` (which is not completely correct, but should not result in the pickle problem)

torchmetrics/average.py
```diff
@@ -0,0 +1,110 @@
...
+            dist_sync_fn=dist_sync_fn,
+        )
+        self.add_state("value", torch.zeros(()), dist_reduce_fx="sum")
+        self.add_state("weight", torch.zeros(()), dist_reduce_fx="sum")
+
+    # TODO: need to be strings because Unions are not pickleable in Python 3.6
+    def update(  # type: ignore
+        self,
+        value: "Union[Tensor, float]",
+        weight: "Union[Tensor, float]" = 1.0
```
Borda, line 83:
```suggestion
        value: Union[Tensor, float],
        weight: Union[Tensor, float] = 1.0
```

alanhdu: This won't work on Python 3.6 -- see the comment above abuot the pickling limitation.

Borda (approved):

PR #154, opened 2021-04-02
Added changes for Test Differentiability [1/n]

### LINKED ISSUE #26: Test for differentiability
## 🚀 Feature

Add a property that for which the user can determine if a metric is differentiable or not
``` python
@property
def is_differentiable(self):
    return True/False
```
and add appropriate tests. We can take inspiration from what kornia is doing:
https://github.com/kornia/kornia/blob/master/test/color/test_gray.py#L69

### Motivation

Some metrics support differentiability, some does not. Would be great if we were more explicit about it and actually have test for it.

bhadreshpsavani:
# Before submitting

## What does this PR do?
Fixes #26

## PR review    
@SkafteNicki @Borda


tests/classification/test_accuracy.py
```diff
@@ -103,6 +103,11 @@ def test_accuracy_fn(self, preds, target, subset_accuracy):
             },
         )
 
+    def test_differentiability(self, preds, target):
+        metric = Accuracy()
```
SkafteNicki, line 107: to simulate this correctly we need tensors with grad enabled 
```suggestion
        preds.requires_grad = True
        metric = Accuracy()
```

tests/regression/test_mean_error.py
```diff
@@ -96,6 +96,11 @@ def test_mean_error_functional(self, preds, target, sk_metric, metric_class, met
             sk_metric=partial(sk_metric, sk_fn=sk_fn),
         )
 
+    def test_differentiability(self, preds, target):
+        metric = MeanSquaredError()
```
SkafteNicki, line 100:
```suggestion
        preds.requires_grad = True
        metric = MeanSquaredError()
```

SkafteNicki (commented):
A couple of extra comments:
1. If the test looks the same for all metrics maybe we could just add it to the base metric tester class
2. We need to add new section to this file https://github.com/PyTorchLightning/metrics/blob/master/docs/source/pages/overview.rst called something like "Metrics and differentiability" where we describe this property
3. I would add the property to the base `Metric` class and raise a `NotImplementedError`

bhadreshpsavani:
Hi @SkafteNicki,

We may have add this in base tester class but you know there is small difference between differentiable and non differentiable metrics
```python 
# not differentiable
assert metric.is_differentiable != out.requires_grad
# differentiable
assert metric.is_differentiable == out.requires_grad
```

bhadreshpsavani:
Hi @SkafteNicki, 
For adding test cases I thought `MetricTester class` would be better. So made changes in that parent class, 
Please Let me know your suggestions.

torchmetrics/metric.py
```diff
@@ -451,6 +451,10 @@ def __pos__(self):
     def __getitem__(self, idx):
         return CompositionalMetric(lambda x: x[idx], self, None)
 
+    @property
+    def is_differentiable(self):
+        raise NotImplementedError
```
SkafteNicki, line 456: please add test to this file https://github.com/PyTorchLightning/metrics/blob/master/tests/bases/test_metric.py that the error is raised if a custom metric does not implement this property

SkafteNicki:
@bhadreshpsavani I added "[1/n]" to the name of the PR, to indicate that this is the first PR in a series related to this topic. Lets focus on getting the testing interface correct, and then in follow up PRs we can add the property to the remaining codebase

SkafteNicki:
@bhadreshpsavani I added "[1/n]" to the name of the PR, to indicate that this is the first PR in a series related to this topic. Lets focus on getting the testing interface correct, and then in follow up PRs we can add the property to the remaining codebase

tests/classification/test_accuracy.py
```diff
@@ -103,6 +103,14 @@ def test_accuracy_fn(self, preds, target, subset_accuracy):
             },
         )
 
+    def test_metrics_differentiability(self, preds, target):
```
SkafteNicki, line 106:
```suggestion
    def test_metrics_differentiability(self, preds, target, *args):
```

torchmetrics/metric.py
bhadreshpsavani: If i add this,
```python
def test_differentiability():
    with pytest.raises(NotImplementedError):
        metric = DummyMetric()
        metric.is_differentiable
```
Is it seems fine?

tchaton: Yes, I think it is fine.

tests/helpers/testers.py
```diff
@@ -400,6 +400,21 @@ def run_precision_test_gpu(
...
+    ):
+        """Test if a metric is differentiable or not
+        Args:
+            preds: torch tensor with predictions
+            target: torch tensor with targets
+            metric_module: the metric module to test
+        """
+        preds.requires_grad = True
+        out = metric_module(preds, target)
+        assert metric_module.is_differentiable == out.requires_grad
```
tchaton, line 417: Why would `metric_module.is_differentiable == out.requires_grad` ? 
I understand the design logic, but the Metric `differentiable` property should be independent of the tensors
`requires_grad` property @SkafteNicki ?

tchaton (approved):
LGTM !

tests/helpers/testers.py
SkafteNicki: If a metric is differentiable then input that requires grad should result in output that also requires grad.
If a metric is not differentiable then if we give it input that requires grad then the output will not have a grad.

docs/source/pages/overview.rst
```diff
@@ -281,3 +281,31 @@ They simply compute the metric value based on the given inputs.
...
+*****************************
+
+Metrics support backpropagation, if all computations involved in the metric calculation
+are differentiable. All modular metrics have a property that determines if a metric is
+differentible or not.
+
+.. code-block:: python
+
+    @property
+    def is_differentiable(self):
```
Borda, line 297:
```suggestion
    def is_differentiable(self) -> bool:
```

tests/helpers/testers.py
```diff
@@ -430,6 +430,38 @@ def run_precision_test_gpu(
...
+    def run_differentiability_test(
+        self,
+        preds: torch.Tensor,
+        target: torch.Tensor,
+        metric_module: Metric,
+        metric_functional: Callable,
+        metric_args: dict = {},
+    ):
+        """Test if a metric is differentiable or not
+        Args:
```
Borda, line 442:
```suggestion

        Args:
```

torchmetrics/metric.py
Borda: shall we rather have by default None?

Borda (approved):

SkafteNicki:
@bhadreshpsavani just to update you. I took the liberty to make a couple of changes to the PR today such that it could get merged. Feel free to send additional PRs adding the property+test to other metrics :]

bhadreshpsavani:
Sure @SkafteNicki,
I will take this PR as reference and try to make changes for other metrics.

PR #164, opened 2021-04-07
Warning on wrong call order

SkafteNicki:
# Before submitting

## What does this PR do?
Warn user if `compute` is called before `update` as most metrics are meaningless or will produce errors if their states have not been updated. Will return 0.0 in this case.

## PR review    

## Did you have fun?


tests/bases/test_metric.py
```diff
@@ -259,3 +259,25 @@ def test_device_and_dtype_transfer(tmpdir):
...
+    # make sure everything is fine with forward
+    with pytest.warns(None) as record:
+        val = metric(1)
+    assert not record
+
+    metric.reset()
+
+    with pytest.warns(UserWarning, match=r'The ``compute`` method of metric .*'):
+        val = metric.compute()
+        assert val == 0.0
```
Borda, line 276:
```suggestion
    assert val == 0.0
```

torchmetrics/metric.py
```diff
@@ -206,6 +208,15 @@ def _wrap_compute(self, compute):
...
         @functools.wraps(compute)
         def wrapped_func(*args, **kwargs):
+            if not self._update_called:
+                rank_zero_warn(
+                    f"The ``compute`` method of metric {self.__class__.__name__}"
+                    " was called before the ``update`` method which may lead to errors,"
+                    " as metric states have not yet been updated. Will return 0.0 instead.",
+                    UserWarning
+                )
+                return 0.0
```
Borda, line 218: shall this be rather NaN?

SkafteNicki: Fine with me :]
What does lightning do in `self.log` if a `NaN` is passed in?

Borda: @tchaton ^^

tchaton: Great question. Looks like Lightning doesn't handle NaN values. It will store it and then produce a NaN value at reduction time.
@SkafteNicki Can you have a look ? I think we might want to skip saving the result object is NaN is logged.

tchaton: And 0 would actually make the reduction False.

SkafteNicki: let me try to update this on the lightning side

maximsch2: I'm not sure if it's worth it to return 0.0 - some metrics are returning tuples/vectors when computed correctly and there will be code to handle it, in that case, there will be a weird exception down the line due to type-mismatch. Might be better to just raise instead.

In terms of type-mismatch, this return 0 will also break type-checking if it was smart enough to understand all the dynamic wrapping happening in the first place :)

SkafteNicki: @maximsch2 so you propose that we raise an error instead?

Borda:
@SkafteNicki is this ready to go?

Borda:
@maximsch2 mind check comments/questions from @SkafteNicki?

torchmetrics/metric.py
Borda: coming back, how about the NaN instead of 0?

maximsch2: The question is - do we want the training to proceed and do we expect this to happen deep into the training? 

Given that most of the time we will probably get the error right away (on_step=True in training_step and validation will be caught by sanity checking), I think we should throw an exception here. Alternatively - maybe just warn and do nothing? Let `compute` fail if it does? 

Another aspect to think about - say I have a metric that is stateless and the only thing it does is that in `compute()` it queries RAM utilization or CPU load or something else and returns it. Given that it's going to be rare and probably done by more advanced users, asking them to call update() can be fair game I guess.

SkafteNicki: then lets settle for raising an warning to the users. Then it depends on the metric what will happen afterwards:
* metrics using list states will fail because `compute` tries to call `torch.cat` on empty list
* metrics of the form `num/denom` will give `nan` as it in most cases it will be `0/0`
* some metrics will behave nicely (like confusionmatrix with no normalization) because the metric is well defined from the initial state

Borda (approved):
Cool =)

maximsch2:
LGTM with the warning.

PR #172, opened 2021-04-14
Fix metrics not being torch-scriptable due to new is_differentiable property

maximsch2:
# Before submitting

## What does this PR do?
Metrics were made scriptable in  https://github.com/PyTorchLightning/pytorch-lightning/pull/4428 but this was broken with addition of `is_differentiable` property which is triggering TorchScript issue (reported to PyTorch). For now, I'm suggesting we just return None there instead of raising an exception.


Borda:
> but this was broken with addition of `is_differentiable` property which is triggering TorchScript issue

heh can we also add a test for this so prevent it in the future... :]

torchmetrics/metric.py
```diff
@@ -453,7 +455,9 @@ def __getitem__(self, idx):
 
     @property
     def is_differentiable(self):
-        raise NotImplementedError
+        # There is a bug in PyTorch that leads to properties being executed during scripting
+        # To make the metric scriptable, we add property to ignore list and switch to return None here
+        return None
```
Borda, line 460: that was my concern earlier... https://github.com/PyTorchLightning/metrics/pull/154#discussion_r612483225

Borda (approved):

torchmetrics/metric.py
maximsch2: It should hopefully be OK to throw once PyTorch fixes the issue (it's not respecting the ignore on properties!), but that would be a while till it gets to the released version, etc.

maximsch2:
Actually there is a fix on PyTorch side already, but it's not part of the released version yet and given the desire for backwards-compatibility I don't think we'll be able to use it for a while: https://github.com/pytorch/pytorch/pull/52367

We need to use `__jit_ignored_attributes__` there essentially which will be supported in new pytorch, I've updated the PR.

SkafteNicki (approved):

PR #173, opened 2021-04-15
make sure metriccollection is deterministic

### LINKED ISSUE #159: MetricCollection should enforce order when passed a dict
## 🐛 Bug

Not a clear bug yet, but just thinking about distributed metric computation, in here: https://github.com/PyTorchLightning/metrics/blob/53d570158a503497351ae45ec895ca44a0546068/torchmetrics/collections.py#L81
we should make sure to sort the names before the insert so that we insert things in the same order (ModuleDict is already OrderedDict otherwise). If we don't we will get deadlocks when doing distributed metric updates.

Additionally, we might want to enforce sorting when passed list/tuple, but that might be more on the user end.

### To Reproduce

On each of the workers, pass dictionary with same metrics but in different order, try compute and observe deadlock.

maximsch2:
# Before submitting

## What does this PR do?
Fixes #159 

## PR review    

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed behaviour of `confusionmatrix` for multilabel data to better match `multilabel_confusion_matrix` from sklearn ([#134](https://github.com/PyTorchLightning/metrics/pull/134))
 - Updated FBeta arguments ([#111](https://github.com/PyTorchLightning/metrics/pull/111))
 - Changed `reset` method to use `detach.clone()` instead of `deepcopy` when resetting to default ([#163](https://github.com/PyTorchLightning/metrics/pull/163))
+- Metrics passed as dict to `MetricCollection` will now always be in deterministic order ([#173](https://github.com/PyTorchLightning/metrics/pull/173))
 
 ### Deprecated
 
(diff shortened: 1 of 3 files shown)


Borda (approved):

tchaton (approved):
LGTM !

Borda:
@SkafteNicki as some comments :]

SkafteNicki (changes requested):
@maximsch2 we need to enforce the same change to when input is passed in as a list (and test for it)
Also we should probably note in the docstring that output will be sorted alphabetically.

maximsch2:
Metrics in a List don't have names, so no good to sort them. And unlike dicts, lists are ordered, so it will be more on the user in that case IMO. My main concern is that `dict.items()` iteration might be different at different times

SkafteNicki:
Metrics passed in as a list do have names, they are just added automatically. Should therefore have same problem as metrics passed in as dict.

Borda:
@SkafteNicki are we set here? 🐰

SkafteNicki (approved):

maximsch2:
I need to add a fix for list case as @SkafteNicki has suggested, but will do that as separate PR.

SkafteNicki:
@maximsch2 after thinking about it, there is no need to it, as you said yourself the user has to explicitly construct a different list order per process for something to go wrong.

PR #178, opened 2021-04-18
Changes for Test Differentiability [2/n]

### LINKED ISSUE #26: Test for differentiability

bhadreshpsavani:
# Before submitting

## What does this PR do?
Fixes #26 

## PR review    
@SkafteNicki 

Note:
* After My changes in the Test File, Test Cases that I wrote were failing for a few scenarios
* I value of Differentiability Property might not be very accurate. I wrote it according to my understanding and few internet searches


tests/classification/test_auroc.py
```diff
@@ -86,18 +86,18 @@ def _sk_auroc_multilabel_multidim_prob(preds, target, num_classes, average='macr
...
 class TestAUROC(MetricTester):
 
+    @pytest.mark.parametrize(
+        "preds, target, sk_metric, num_classes", [
+            (_input_binary_prob.preds, _input_binary_prob.target, _sk_auroc_binary_prob, 1),
+            (_input_mcls_prob.preds, _input_mcls_prob.target, _sk_auroc_multiclass_prob, NUM_CLASSES),
+            (_input_mdmc_prob.preds, _input_mdmc_prob.target, _sk_auroc_multidim_multiclass_prob, NUM_CLASSES),
+            (_input_mlb_prob.preds, _input_mlb_prob.target, _sk_auroc_multilabel_prob, NUM_CLASSES),
+            (_input_mlmd_prob.preds, _input_mlmd_prob.target, _sk_auroc_multilabel_multidim_prob, NUM_CLASSES)]
+    )
```
SkafteNicki, line 100: move this in front of the class so it parametrize all tests and remove all duplicated
even though `test_auroc_differentiability` does not use the `sk_metric` arg it can still be part of the function definition

tests/classification/test_auroc.py
```diff
@@ -127,6 +127,14 @@ def test_auroc(self, preds, target, sk_metric, num_classes, average, max_fpr, dd
...
         )
 
+    @pytest.mark.parametrize(
+        "preds, target, sk_metric, num_classes", [
+            (_input_binary_prob.preds, _input_binary_prob.target, _sk_auroc_binary_prob, 1),
+            (_input_mcls_prob.preds, _input_mcls_prob.target, _sk_auroc_multiclass_prob, NUM_CLASSES),
+            (_input_mdmc_prob.preds, _input_mdmc_prob.target, _sk_auroc_multidim_multiclass_prob, NUM_CLASSES),
+            (_input_mlb_prob.preds, _input_mlb_prob.target, _sk_auroc_multilabel_prob, NUM_CLASSES),
+            (_input_mlmd_prob.preds, _input_mlmd_prob.target, _sk_auroc_multilabel_multidim_prob, NUM_CLASSES)]
+    )
```
SkafteNicki, line 137:
```suggestion
```

tests/classification/test_auroc.py
```diff
@@ -152,6 +160,39 @@ def test_auroc_functional(self, preds, target, sk_metric, num_classes, average,
...
         )
 
+    @pytest.mark.parametrize(
+        "preds, target, num_classes", [
+            (_input_binary_prob.preds, _input_binary_prob.target, 1),
+            (_input_mcls_prob.preds, _input_mcls_prob.target, NUM_CLASSES),
+            (_input_mdmc_prob.preds, _input_mdmc_prob.target, NUM_CLASSES),
+            (_input_mlb_prob.preds, _input_mlb_prob.target, NUM_CLASSES),
+            (_input_mlmd_prob.preds, _input_mlmd_prob.target, NUM_CLASSES)]
+    )
```
SkafteNicki, line 170:
```suggestion
```

tests/classification/test_auroc.py
```diff
@@ -152,6 +160,39 @@ def test_auroc_functional(self, preds, target, sk_metric, num_classes, average,
...
+    def test_auroc_differentiability(self, preds, target, num_classes, average, max_fpr):
```
SkafteNicki, line 171:
```suggestion
    def test_auroc_differentiability(self, preds, target, sk_metric, num_classes, average, max_fpr):
```

tests/classification/test_average_precision.py
```diff
@@ -61,15 +61,15 @@ def _sk_avg_prec_multidim_multiclass_prob(preds, target, num_classes=1):
...
-@pytest.mark.parametrize(
-    "preds, target, sk_metric, num_classes", [
-        (_input_binary_prob.preds, _input_binary_prob.target, _sk_avg_prec_binary_prob, 1),
-        (_input_mcls_prob.preds, _input_mcls_prob.target, _sk_avg_prec_multiclass_prob, NUM_CLASSES),
-        (_input_mdmc_prob.preds, _input_mdmc_prob.target, _sk_avg_prec_multidim_multiclass_prob, NUM_CLASSES),
-    ]
-)
 class TestAveragePrecision(MetricTester):
 
+    @pytest.mark.parametrize(
```
SkafteNicki, line 66: same as above comments

tests/classification/test_cohen_kappa.py
```diff
@@ -77,20 +77,20 @@ def _sk_cohen_kappa_multidim_multiclass(preds, target, weights=None):
...
-     (_input_mlb.preds, _input_mlb.target, _sk_cohen_kappa_multilabel, 2),
-     (_input_mcls_prob.preds, _input_mcls_prob.target, _sk_cohen_kappa_multiclass_prob, NUM_CLASSES),
-     (_input_mcls.preds, _input_mcls.target, _sk_cohen_kappa_multiclass, NUM_CLASSES),
-     (_input_mdmc_prob.preds, _input_mdmc_prob.target, _sk_cohen_kappa_multidim_multiclass_prob, NUM_CLASSES),
-     (_input_mdmc.preds, _input_mdmc.target, _sk_cohen_kappa_multidim_multiclass, NUM_CLASSES)]
-)
 class TestCohenKappa(MetricTester):
     atol = 1e-5
 
+    @pytest.mark.parametrize(
```
SkafteNicki, line 83: same as above comments

tests/classification/test_auroc.py
bhadreshpsavani: Sure, I will do these changes!

tests/classification/test_auroc.py
bhadreshpsavani: Done

tests/classification/test_average_precision.py
bhadreshpsavani: Done

tests/classification/test_cohen_kappa.py
bhadreshpsavani: Done

torchmetrics/classification/cohen_kappa.py
```diff
@@ -117,3 +117,10 @@ def compute(self) -> Tensor:
         Computes cohen kappa score
         """
         return _cohen_kappa_compute(self.confmat, self.weights)
+
+    @property
+    def is_differentiable(self):
+        """
+        The equation of cohen kappa score is differentiable
+        """
+        return True
```
SkafteNicki, line 126: cohen kappa is not differentiable since the implementation is based on calculating the confusion matrix which in general is not differentiable

bhadreshpsavani: Ohk, I will change it

SkafteNicki:
some tests are failing because this assert cannot handle metrics that returns multiple values.
https://github.com/PyTorchLightning/metrics/blob/c1ae5ace2dc03525720ab3a867ce35fe50a2d276/tests/helpers/testers.py#L465
should be changed to something like:
```python
if isinstance(out, list):
    assert all(metric.is_differentiable == o.requires_grad for o in out)
else:
    assert metric.is_differentiable == out.requires_grad
```

SkafteNicki (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `is_differentiable` property to `AUC`, `AUROC`, `CohenKappa` and `AveragePrecision` ((#178)[https://github.com/PyTorchLightning/metrics/pull/178])
```
SkafteNicki, line 12:
```suggestion
- Added `is_differentiable` property to `AUC`, `AUROC`, `CohenKappa` and `AveragePrecision` ([#178](https://github.com/PyTorchLightning/metrics/pull/178))
```

SkafteNicki:
After looking a bit on the failing tests it seems that:
* AUC should in general be differentiable and the `False` should be changed to `True`
* However, for half of the test cases (those with batch size 4049 https://github.com/PyTorchLightning/metrics/blob/cceced613f4323a1f5124099a969f2cf32a80d7e/tests/classification/test_auc.py#L44) the grad check fails, indicating that the gradient calculation in pytorch is wrong. It makes sense that it only fails for large inputs as we know pytorch is using different implementations in `torch.sort` depending on the size of the input:
https://github.com/pytorch/pytorch/blob/a24163a95edb193ff7b06e98cd69bf7cfd4c0d2f/aten/src/THC/generic/THCTensorSort.cu#L306-L334

@bhadreshpsavani could you add a `pytest.skipif` to the differentiability test of `AUC` such that it is skipped if the input is larger than 2048?

Borda:
@bhadreshpsavani mind check the failing tests?

maximsch2:
For the pytorch bug - it would be great to have a small isolated repro to file as an issue for pytorch.

On AUC - is it really meaningful to treat it as differentiable? What tasks will do that?

bhadreshpsavani:
Hi @borda,
I will make changes according to the suggestion of @SkafteNicki,
Let me know if anything I need to change.

bhadreshpsavani:
Hi @SkafteNicki,
I made changes as per your suggestions
Eventhough i add this line `@pytest.mark.skipif(x.shape[1] > 2048)` in `metrics/tests/classification/test_auc.py` it still fails.

SkafteNicki:
After searching the internet, AUC indeed is described as an non-differentiable function and it must therefore be our specific implementation in pytorch that allows gradient to be backprop. 
I have inserted a `torch.no_grad` to secure that the function is non-differentiable which also fixes the tests.

Borda (approved):
LGTM

Issue #74, opened 2021-03-13
Non-Softmaxed Classification


justusschock:
### Motivation

Support other types for classification metrics. I.e non-softmaxed network outputs. 

For outputs of categorical classification, it does not matter if the output is softmaxed or not. The argmax of these tensors is the same. Can we support those by simply taking an argmax even if the values are out of the 0-1 range?

cc @SkafteNicki Whether we want to support this.

SkafteNicki:
Agree that we should support this. This issue is similar to https://github.com/PyTorchLightning/metrics/issues/60, but we should make it more general than just Accuracy metric.

PR #200, opened 2021-04-26
Allow logit input in classification metrics

SkafteNicki:
# Before submitting

## What does this PR do?
Fixes https://github.com/PyTorchLightning/metrics/issues/74
Fixes https://github.com/PyTorchLightning/metrics/issues/60
Fixes https://github.com/PyTorchLightning/lightning-bolts/issues/587

Allow many classification metrics to now work with raw model predictions/logits/unnormalized scores.
List of metrics that gain support and are tested
* Accuracy
* Precision
* Recall
* F1
* FBeta
* StatsScores
* Hamming
* ConfusionMatrix

## PR review    

## Did you have fun?


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added support for unnormalized scores (e.g. logits) in `Accuracy`, `Precision`, `Recall`, `FBeta`, `F1` metrics ([#200](https://github.com/PyTorchLightning/metrics/pull/200))
```
SkafteNicki, line 12:
```suggestion
- Added support for unnormalized scores (e.g. logits) in `Accuracy`, `Precision`, `Recall`, `FBeta`, `F1`, `StatScore`, `Hamming`, `ConfusionMatrix metrics ([#200](https://github.com/PyTorchLightning/metrics/pull/200))
```

Borda (approved):
lgtm

Borda:
@SkafteNicki mind checks the failing tests, otherwise, it looks fine...
cc: @maximsch2

tests/classification/inputs.py
```diff
@@ -28,6 +28,10 @@
     target=torch.randint(high=2, size=(NUM_BATCHES, BATCH_SIZE))
 )
 
+_input_binary_logits = Input(
+    preds=torch.rand(NUM_BATCHES, BATCH_SIZE), target=torch.randint(high=2, size=(NUM_BATCHES, BATCH_SIZE))
```
maximsch2, line 32: randn to make sure we get negative values as well to test?

maximsch2:
lgtm in general, seems like broken tests are testing old behavior of doing logit checks. 

A high level thought is that this can be getting more confusing for users to keep track of which metrics support logits vs probabilities, we might want to have a table somewhere.

Borda:
> A high level thought is that this can be getting more confusing for users to keep track of which metrics support logits vs probabilities, we might want to have a table somewhere.

Yes, I feel some conflict with logit and probably threshold as argument...

SkafteNicki:
@Borda any suggestions what to do? I kept the 0.5 to be backwards compatible. Setting it to 0.0 (corresponding default threshold when working with logits) does not make much sense to me. Last option is to set it to `None`, forcing users to make a choice (would not be backward compatible).

maximsch2:
I feel like there are two separate usecases:
1. Exclusive multiclass classification (potentially with num_classes=2) produces a list of logits/probabilities/etc and we just pick top one to get the prediction, threshold doesn't matter.
2. Binary classification where we output one value or multilabel classification (where again each value corresponds to just that class) - here threshold is important.

What do you think of splitting those up more explicitly and removing threshold for case 1 + removing support for probabilities/logits for case 2 (users have to threshold themselves)?

Borda:
how about if we preserve the probably expectation in case threshold is a number nad set it as logits for threshold None?

Borda:
> how about if we preserve the proablility expectation in case threshold is a number and set it as logits for threshold None?

@SkafteNicki thoughts? :rabbit:

maximsch2:
For logits we need to have a threshold as well to convert to binary predictions though?

Borda:
another thought shall add bool arg `resticted_probability`?

SkafteNicki:
IMO we keep it simple for now by just adding this functionality by removing the explicit check. It will have the minimal change required for code.

Then we probably should re-discuss the concept of task type (started here https://github.com/PyTorchLightning/metrics/discussions/186 by @maximsch2) such that we can make it more clear when threshold applies, hopefully remove the `multiclass` parameter (which I am not a fan of) ect. When that is clear, we can redo classification for version 0.X (hopefully for the last time)

Borda (approved):

maximsch2 (approved):

PR #202, opened 2021-04-26
Loads metric states from state_dict

hlin09:
The persistent metric states are output in `state_dict`, but cannot be properly loaded through `load_state_dict()`. This PR adds handling to loading persisted metric states from `state_dict` and a unit test.

### DIFF
--- a/torchmetrics/metric.py
+++ b/torchmetrics/metric.py
@@ -324,6 +324,15 @@ class Metric(nn.Module, ABC):
                 destination[prefix + key] = current_val
         return destination
 
+    def _load_from_state_dict(self, state_dict, prefix, local_metadata, strict,
+                              missing_keys, unexpected_keys, error_msgs):
+        for key in self._defaults.keys():
+            name = prefix + key
+            if name in state_dict:
+                setattr(self, key, state_dict.pop(name))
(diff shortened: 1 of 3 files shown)


torchmetrics/metric.py
```diff
@@ -324,6 +324,16 @@ def state_dict(self, destination=None, prefix="", keep_vars=False):
                 destination[prefix + key] = current_val
         return destination
 
+    def _load_from_state_dict(self, state_dict, prefix, local_metadata, strict,
+                              missing_keys, unexpected_keys, error_msgs):
```
ananthsub, line 328: add typehints here as well

ananthsub (commented):
this lgtm - @SkafteNicki what do you think?

torchmetrics/metric.py
SkafteNicki, line 328:
```suggestion
    def _load_from_state_dict(self, state_dict: dict, prefix: str, local_metadata: dict, strict: bool,
                              missing_keys: List[str], unexpected_keys: List[str], error_msgs: List[str]) -> None:
```

SkafteNicki (approved):
LGTM

Borda (approved):

PR #209, opened 2021-04-28
[bugfix] MetricCollection should return metrics with prefix on items(), keys()

tchaton:
# Before submitting

## What does this PR do?

This PR makes sure `items() / keys()` properly returns the right keys with `prefix / postfix`.

Fixes #https://github.com/PyTorchLightning/pytorch-lightning/issues/7235.

## PR review    

## Did you have fun?


itsikad:
Thanks @tchaton 

I'm probably missing something here, but why not changing the clone method such that it adds prefix/postfix the the dictionary keys?

SkafteNicki (approved):
LGTM

Borda (approved):

tests/bases/test_collections.py
```diff
@@ -168,6 +168,25 @@ def test_metric_collection_prefix_postfix_args(prefix, postfix):
...
+        assert 'new_prefix_' not in k
+
+    for k in new_metric_collection.keys(keep_base=True):
+        assert 'new_prefix_' not in k
+
+    repr = 'MetricCollection(' \
+           '\n  prefix=new_prefix_,' \
+           f'\n  postfix={postfix},' \
+           '\n  (DummyMetricSum): DummyMetricSum()' \
+           '\n  (DummyMetricDiff): DummyMetricDiff()\n)'
```
carmocca, line 187: I think this would look nicer if prefix/postfix were at the end

carmocca (approved):
LGTM overall. Not the most familiar with this piece of code.

Disabled auto-merge so you can check out my comment :)

tests/bases/test_collections.py
tchaton: I am using `extra_repr` hook and this adds to the debugging by default.

Borda:
@maximsch2 mind check this one too?

tests/bases/test_collections.py
```diff
@@ -168,6 +168,25 @@ def test_metric_collection_prefix_postfix_args(prefix, postfix):
...
+        assert 'new_prefix_' in k
+
+    for k, _ in new_metric_collection.items(keep_base=True):
+        assert 'new_prefix_' not in k
+
+    for k in new_metric_collection.keys(keep_base=True):
+        assert 'new_prefix_' not in k
+
+    repr = 'MetricCollection(' \
+           '\n  prefix=new_prefix_,' \
```
Borda, line 184:
```suggestion
           f'\n  prefix={prefix},' \
```
why not also this?

torchmetrics/collections.py
```diff
@@ -182,16 +182,38 @@ def persistent(self, mode: bool = True) -> None:
...
         return name
 
+    def keys(self, keep_base: bool = False):
+        r"""Return an iterable of the ModuleDict key.
+        Args:
+            keep_base: Whether to add prefix/postfix on the items collection.
+        """
+        if keep_base:
+            return self._modules.keys()
+        return [self._set_name(k) for k in self._modules.keys()]
```
maximsch2, line 200: I don't like that we now have a subtle difference based on the flag. By default, we'll return a dictview object (https://docs.python.org/3/library/stdtypes.html#dict-views), but if we set a flag we'll get a list instead (which can be accidentally mutated for example, etc).

What's the value of keeping old behavior? Given the prefix was provided in clone, maybe just append it at that time?

maximsch2:
I agree with @itsikad , we are introducing a new behavior there where as making a clean copy/clone will be simpler. I do see that we had prefix/suffix arguments in the constructor before already. Do we know if those are used anywhere?

tests/bases/test_collections.py
tchaton: Done !

tests/bases/test_collections.py
tchaton: Because the prefix was fixed. I added a proper test for repr.

tchaton:
> I agree with @itsikad , we are introducing a new behavior there where as making a clean copy/clone will be simpler. I do see that we had prefix/suffix arguments in the constructor before already. Do we know if those are used anywhere?

This was used in Lightning and causing a bug (https://github.com/PyTorchLightning/pytorch-lightning/issues/7235) as Lightning is using internally `metric.items()`.
I believe the MetricCollection can store internally without prefix but should always return the keys with prefix/postfix.

best,
T.C

torchmetrics/collections.py
tchaton: Good catch ! I changed this.

torchmetrics/collections.py
```diff
@@ -182,16 +183,42 @@ def persistent(self, mode: bool = True) -> None:
...
         return name
 
+    def keys(self, keep_base: bool = False):
+        r"""Return an iterable of the ModuleDict key.
+        Args:
+            keep_base: Whether to add prefix/postfix on the items collection.
+        """
+        if keep_base:
+            return self._modules.keys()
+        return OrderedDict(**{self._set_name(k): v for k, v in self._modules.items()}).keys()
```
carmocca, line 201: Why do you use an `OrderedDict`? The inner `dict` does not enforce the order so you've lost the order already.

Same below

SkafteNicki: since `MetricCollection` inherent from `nn.ModuleDict` which is ordered dict then this should also be

carmocca: My point is we have:

```python
x = {self._set_name(k): v for k, v in self._modules.items()}
return OrderedDict(x).keys()
```

But the dict comprehension to create `x` loses the order guarantee.

So this should be instead:

```suggestion
        return OrderedDict([(self._set_name(k), v) for k, v in self._modules.items()]).keys()
```

tchaton: Updated.

carmocca (approved):

maximsch2:
@tchaton , I agree that we should fix the checkpointing behavior, but my point is that it might be cleaner to just change the `clone` method so that the new collection actually has the keys with prefixes inside, that way we don't have to have hacks that pretend otherwise at run time.

tchaton:
> @tchaton , I agree that we should fix the checkpointing behavior, but my point is that it might be cleaner to just change the `clone` method so that the new collection actually has the keys with prefixes inside, that way we don't have to have hacks that pretend otherwise at run time.

Hey @SkafteNicki. What are your thoughts ? Happy to let someone else change the MetricCollection to save the metrics with prefix and postfix directly.

Best,
T.C

SkafteNicki:
@tchaton I am in favour of this change so the keys are consistent between `.items()`, `.keys()` and the dics returned by `forward` and `.compute()`

awaelchli (approved):

maximsch2:
Actually I now think this is fine - we can just flip the implementation later and just do the in-place modification in case of `keep_base=True`.

lgtm

Borda:
> Actually I now think this is fine - we can just flip the implementation later and just do the in-place modification in case of `keep_base=True`.

so pls make an issue for it for this milestone so we do not forget it... (and link it in this PR)

PR #212, opened 2021-04-29
Remove numpy dependency from the core library

maximsch2:
# Before submitting

## What does this PR do?
Remove numpy dependency that seemed to have been added mostly for one test to make it cleaner. Instead moved the function that needs to support both Tensor and ndarray to test utils and removed the dependency from the core library.

## PR review

## Did you have fun?

### DIFF
--- a/tests/retrieval/helpers.py
+++ b/tests/retrieval/helpers.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from functools import partial
-from typing import Callable, Dict, Tuple, Type, Union
+from typing import Callable, Dict, List, Tuple, Type, Union
 
 import numpy as np
 import pytest
--- a/tests/retrieval/helpers.py
+++ b/tests/retrieval/helpers.py
@@ -30,9 +30,39 @@ from tests.retrieval.inputs import _input_retrieval_scores_mismatching_sizes as 
 from tests.retrieval.inputs import _input_retrieval_scores_mismatching_sizes_func as _irs_mis_sz_fn
 from tests.retrieval.inputs import _input_retrieval_scores_no_target as _irs_no_tgt
 from tests.retrieval.inputs import _input_retrieval_scores_wrong_targets as _irs_bad_tgt
-from torchmetrics.utilities.data import get_group_indexes
 
 seed_all(42)
+
+# a version of get_group_indexes that depends on NumPy is here to avoid this dependency for the full library
+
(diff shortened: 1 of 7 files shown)


Borda:
`numpy` is installed with PyTorch anyway and `numpy` is everywhere so what is the motivation?

tests/retrieval/helpers.py
```diff
@@ -30,10 +30,40 @@
...
 from tests.retrieval.inputs import _input_retrieval_scores_no_target as _irs_no_tgt
 from tests.retrieval.inputs import _input_retrieval_scores_wrong_targets as _irs_bad_tgt
-from torchmetrics.utilities.data import get_group_indexes
 
 seed_all(42)
 
+# a version of get_group_indexes that depends on NumPy is here to avoid this dependency for the full library
+
+
+def get_group_indexes(indexes: Union[Tensor, np.ndarray]) -> List[Union[Tensor, np.ndarray]]:
```
Borda, line 39: Here you still use np

SkafteNicki:
agree with @Borda on this one, no need to remove `numpy` as requirement as it is already installed with `torch`

tests/retrieval/helpers.py
maximsch2: This is in tests, we do use scikit-learn in tests as well, which is fine, I'm only worried about the core library being lean.

maximsch2:
I don't think pytorch requires numpy to run, it's likely for tests as well. In fact, we are running pytorch in builds without numpy so it's for sure not required.

In this particular case, the core library doesn't really require numpy, so I'm thinking we should keep it lean especially given it's a no-cost thing for us. The two current uses for numpy in torchmetrics were:
- np.nan (replaced by float('nan') which is same
- helper function that was made to take np.ndarray so that the test will be used.

Both are not needed for metrics to function.

ananthsub:
@SkafteNicki @Borda one of the concerns is around backwards compatibility for metrics. numpy introduces a large surface area which can make this harder to manage. given how minimal the usage is in torchmetrics now, we'd prefer to keep the library lean and use torch primitives instead to better control for this

Borda:
Ok, I am fine with dropping it 🐰

SkafteNicki:
@maximsch2 and @ananthsub great points, also fine with dropping it, did not really have a strong opinion on this :]

Borda (approved):

SkafteNicki (approved):

PR #215, opened 2021-04-29
Make torchmetrics PEP 561 compatible

### LINKED ISSUE #201: Distribute + package type hints
## 🚀 Feature

Right now, `torchmetrics` mostly uses type hints for internal testing. However, these type annotations are not packaged for type checkers to consume downstream. From [PEP 561](https://www.python.org/dev/peps/pep-0561/):

> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. 

Without this marker file, my IDE takes advantage of these type annotations, but `mypy` makes thinks everything from `torchmetrics` has an `Any` type. 

I believe the solution is to just create an empty `py.typed` file and add it to the `setup.py` (as specified https://www.python.org/dev/peps/pep-0561/#packaging-type-information) -- that is, if we think the type annotations in `torchmetrics` are "ready" to be relied on.

alanhdu:
Closes #201

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,6 +10,9 @@ recursive-include torchmetrics *.md
 # Include the license file
 include LICENSE
 
+# Include marker file for PEP 561
+include torchmetrics/py.typed
+
 exclude *.sh
 exclude *.toml
 exclude *.svg


alanhdu:
I think this is all that's necessary? I admit I'm not super familiar with Python packaging, so I wouldn't be surprised if I missed something.

Borda (approved):

PR #217, opened 2021-04-29
RFC: reduce allgather costs

maximsch2:
# Before submitting

## What does this PR do?
We have a number of metrics that are following a pattern of accumulating (preds, targets) in the state as a list. This leads to a lot of individual AllGather communications at the very end as we end up running them one validation batch at a time. In this diff, I'm suggesting we do a `cat` on those states before we do synchronization.

I'm showing it only for AUROC (sending PR to get CI to run tests, but they do pass locally). If people agree that this is a good idea, we can discuss how to do this - might be we can have a flag on the base `Metric` that will enable this in vanilla `_sync_dist` function (to avoid having to copy/paste this snippet around to PR, AveragePrecision and possible other metrics as well)

## PR review

## Did you have fun?


maximsch2:
Actually implemented a more general solution for dist_reduce_fx="cat" and switched metrics to using it. Interestingly, we didn't have any built-in metrics using "cat" reduction - all of them were doing it manually.

torchmetrics/metric.py
```diff
@@ -180,6 +180,10 @@ def forward(self, *args, **kwargs):
 
     def _sync_dist(self, dist_sync_fn=gather_all_tensors):
         input_dict = {attr: getattr(self, attr) for attr in self._reductions.keys()}
+        for attr, reduction_fn in self._reductions.items():
+            # dim_zero_cat can be applied multiple times, this optimizes the number of gathers
```
SkafteNicki, line 184: just personal preference
```suggestion
            # pre-concatenate metric states that are lists to reduce number of all_gather operations
```

torchmetrics/metric.py
```diff
@@ -180,6 +180,10 @@ def forward(self, *args, **kwargs):
...
+            if reduction_fn == dim_zero_cat and len(input_dict[attr]) > 1:
```
SkafteNicki, line 185: in principal there is nothing preventing the the user from having metric state as a tensor and also set `dist_sync_fx='cat'`. To be sure we do not mess things up we probably have to also add condition that it should be a list:
```suggestion
            if reduction_fn == dim_zero_cat and isinstance(input_dict[attr]) and len(input_dict[attr]) > 1:
```

SkafteNicki (commented):
LGTM, feel free to add change to more metrics!

maximsch2:
Oh, nice, I missed those!

SkafteNicki (approved):

Borda (approved):

Issue #203, opened 2021-04-27
True update() for MetricCollection


IgorHoholko:
## 🚀 Feature
`MetricCollection` class overrides `update `method of parent `nn.ModuleDict` class. Is it possible to add `update` from parent's class to MetricCollection?
```python
 def update_(self, metrics: Union[List[Metric], Tuple[Metric], Dict[str, Metric]]):
        # ... another work for List[Metric], Tuple[Metric] ...
        else:
            return super().update(metrics)
```

### Example of using

```python
metrics = MetricCollection({metric_name: getattr(metrics_obj, metric_name)
                                    for metric_name in self.metrics})

self.train_metrics = metrics.clone(prefix='train_')
self.val_metrics = metrics.clone(prefix='val_')
self.test_metrics = metrics.clone(prefix='test_')
self.test_metrics.update({metric_name: getattr(metrics_obj, metric_name)
                            for metric_name in  ('auroc', 'auroc_macro', 'auroc_weighted')})
```

IgorHoholko:
@SkafteNicki Makes sence. Thanks!

SkafteNicki:
@IgorHoholko you are welcome. Closing the issue. Feel free to re-open if necessary :]

maximsch2:
Should we do something like 
```
def add_metrics(self, metrics):
   super().update(metrics)
```
?

IgorHoholko:
@maximsch2  It would definitely be helpful.

IgorHoholko:
`metrics.add_metricsooks()` looks better and more intuitive than `super(type(metrics), metrics).update()` :)

SkafteNicki:
fine by me :]
@IgorHoholko want to send a PR?

IgorHoholko:
@SkafteNicki Yes, a bit later.

PR #221, opened 2021-05-02
Added add_metrics method to MetricCollection

IgorHoholko:
Fixes [#203](https://github.com/PyTorchLightning/metrics/issues/203)
# Before submitting

## What does this PR do?
`add_metrics()` method for MetricCollection class.

## PR review

## Did you have fun?

### DIFF
--- a/torchmetrics/collections.py
+++ b/torchmetrics/collections.py
@@ -94,46 +94,8 @@ class MetricCollection(nn.ModuleDict):
         postfix: Optional[str] = None
     ):
         super().__init__()
-        if isinstance(metrics, Metric):
-            # set compatible with original type expectations
-            metrics = [metrics]
-        if isinstance(metrics, Sequence):
-            # prepare for optional additions
-            metrics = list(metrics)
-            remain = []
-            for m in additional_metrics:
-                (metrics if isinstance(m, Metric) else remain).append(m)
 
-            if remain:
-                rank_zero_warn(
-                    f"You have passes extra arguments {remain} which are not `Metric` so they will be ignored."
-                )
-        elif additional_metrics:
-            raise ValueError(
-                f"You have passes extra arguments {additional_metrics} which are not compatible"
-                f" with first passed dictionary {metrics} so they will be ignored."
-            )
-
-        if isinstance(metrics, dict):
-            # Check all values are metrics
-            # Make sure that metrics are added in deterministic order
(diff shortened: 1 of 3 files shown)


SkafteNicki (approved):
LGTM

torchmetrics/collections.py
```diff
@@ -185,6 +147,51 @@ def persistent(self, mode: bool = True) -> None:
         for _, m in self.items():
             m.persistent(mode)
 
+    def add_metrics(self, metrics: Union[Metric, Sequence[Metric], Dict[str, Metric]],
```
Borda, line 150:
```suggestion
    def append(self, metrics: Union[Metric, Sequence[Metric], Dict[str, Metric]],
```
would it be a better name?

torchmetrics/collections.py
```diff
@@ -185,6 +147,51 @@ def persistent(self, mode: bool = True) -> None:
...
+                    *additional_metrics: Metric) -> None:
+        """Add new metrics to Metric Collection
+        """
+        if isinstance(metrics, Metric):
+            # set compatible with original type expectations
+            metrics = [metrics]
+        if isinstance(metrics, Sequence):
```
maximsch2, line 157: I understand this is just copying what we had here before, but we should probably have `Iterable` here instead of `Sequence` as we doin't really care about indexing.

torchmetrics/collections.py
maximsch2, line 150: It should be `Mapping[str, Metrics]`, not `Dict` + similar a check below.

maximsch2: I think it's fine to keep the name different. If we say `append` it might imply that it won't work as `extend` while here it does.

torchmetrics/collections.py
Borda: but iterable does not have len, right?

Borda:
@IgorHoholko mind check the comments above?

torchmetrics/collections.py
maximsch2: We are not using len though, we are just directly converting to a list in the next line.

torchmetrics/collections.py
tchaton: `add_metrics` seems more appropriate for a Collection.

tchaton (approved):
LGTM ! Worth to increase the coverage slightly.

PR #224, opened 2021-05-04
Improve AUC numeric stability

SkafteNicki:
# Before submitting

## What does this PR do?
Fixes https://github.com/PyTorchLightning/metrics/discussions/219
In rare case when the input is very large, the `dx` here
https://github.com/PyTorchLightning/metrics/blob/cb6899bb47c7d30f8626d6ef8c28cc29efce82d6/torchmetrics/functional/classification/auc.py#L41-L49
will even for monotone increasing/decreasing (input that is already sorted) result in a mix of positive and negative errors making the algorithm raise the error. This happens when `x[i]` and `x[i+1]` is the same number but due to numerical stability `x[i+1]-x[i]` will randomly either be a small negative or positive number.

This PR solves it by adding a small tolerance to the checks.

I checked with the provided repo/script in the discussion that it correctly solves the error.

## PR review

## Did you have fun?

### DIFF
--- a/torchmetrics/functional/classification/auc.py
+++ b/torchmetrics/functional/classification/auc.py
@@ -31,7 +31,7 @@ def _auc_update(x: Tensor, y: Tensor) -> Tuple[Tensor, Tensor]:
     return x, y
 
 
-def _auc_compute(x: Tensor, y: Tensor, reorder: bool = False) -> Tensor:
+def _auc_compute(x: Tensor, y: Tensor, reorder: bool = False, tol: float = 1e-6) -> Tensor:
     with torch.no_grad():
         if reorder:
             # TODO: include stable=True arg when pytorch v1.9 is released
@@ -40,7 +40,7 @@ def _auc_compute(x: Tensor, y: Tensor, reorder: bool = False) -> Tensor:
 
         dx = x[1:] - x[:-1]
         if (dx < 0).any():
-            if (dx <= 0).all():
+            if (dx <= tol).all():
                 direction = -1.
             else:
                 raise ValueError(
--- a/torchmetrics/functional/classification/auc.py
+++ b/torchmetrics/functional/classification/auc.py
@@ -39,7 +39,7 @@ def _auc_compute(x: Tensor, y: Tensor, reorder: bool = False, tol: float = 1e-6)
             y = y[x_idx]
 
         dx = x[1:] - x[:-1]
-        if (dx+tol < 0).any():
+        if (dx + tol < 0).any():
             if (dx <= tol).all():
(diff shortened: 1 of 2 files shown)


Borda (approved):

torchmetrics/functional/classification/auc.py
```diff
@@ -31,16 +31,16 @@ def _auc_update(x: Tensor, y: Tensor) -> Tuple[Tensor, Tensor]:
...
         if reorder:
             # TODO: include stable=True arg when pytorch v1.9 is released
             x, x_idx = torch.sort(x)
             y = y[x_idx]
 
         dx = x[1:] - x[:-1]
-        if (dx < 0).any():
-            if (dx <= 0).all():
+        if (dx + tol < 0).any():
+            if (dx <= tol).all():
```
maximsch2, line 43: am I reading it correctly that if I have 

dx = [-1e-7, 5e-7, 5e-7, ... 10000 times ...., 5e-7] then we'll deduce incorrect direction here?

Borda: yes, it seems that direction/sign is changed, so shall ve preserve it...

SkafteNicki: maybe I am misunderstanding the question...
the tensor
```dx = [1e-7, 5e-7, 5e-7, ... 10000 times ...., 5e-7]```
implies that `direction=1` right? so if we say that numerical instability leads to a change of sign in the first element
```dx = [-1e-7, 5e-7, 5e-7, ... 10000 times ...., 5e-7]```
then it will still be `direction=1` with the change.

maximsch2: Sorry, let's maybe use this instead:
`dx = [-1.0, 5e-7, 5e-7, ... 10000 times ...., 5e-7]`
Now both `(dx+tol<0).any()` and `(dx <= tol).all()` are true and we'll discover direction as -1.

SkafteNicki: thanks, I see the problem now...
Do you have anyway to solve this?

maximsch2: Numerical issues are usually tricky. I'm suggesting we remove checks for the internal callers of this function, let me quickly show an example.

maximsch2: @SkafteNicki , check out https://github.com/PyTorchLightning/metrics/pull/230

SkafteNicki: @maximsch2 looks good, closing this in favour of yours

PR #241, opened 2021-05-11
Add CRPS metric

### LINKED ISSUE #223: Continuous Ranked Probability Score (CRPS) metric
## 🚀 Feature

The Continuous Ranked Probability Score (CRPS) is a metric for ensemble verification of continuous variables.
It is the continuous version of the  Ranked Probability Score (RPS), which is used for ensemble classification.

A nice description can be found here: https://www.xces.dkrz.de/about/problems/

### Motivation

Sampling from a GAN can be compared to having an ensemble of models. All existing metrics in Pytorch and Pytorch-Lightning do not allow for evaluating the quality of such multiple draws from a GAN, only evaluating each draw individually. 

Use cases would be where any generative model is used to predict a conditioned continuous field or even a single variable where all draws are strongly correlated, for example forecasting weather-patterns or disaster-evaluation. 
It will not be useful in the cases where the draws strongly uncorrelated, like in image-generation of faces and such.

### Pitch

I already have code for the CRPS written in pytorch, which I could provide including some tests.
Before opening a PR I just wanted to check if something like this score is really wanted.

### Alternatives

One could as well also provide the CRP for classification.
The CRPS can also be transformed into a skill score. A skill score would need as additional input a baseline to compare the skill against. This could for example be the condition in the case of a conditional GAN. (I have code for this as well)

### Additional context

For training GANs this score can be used as a loss function since it is 0 for a perfect ensemble and grows for worse ensembles (I have not tried that in practice yet).

tmachnitzki:
I added the class and a functional metric as discussed in #223. The docs and tests are still missing.

I am totally new to this `add_state` thing in the init, so I would be glad if someone with more experience could have a first glance at the code.

closes #223


torchmetrics/functional/regression/crps.py
```diff
@@ -0,0 +1,66 @@
...
+    Computes continuous ranked probability score.
+
+    Args:
+        preds: estimated labels. Second dimension is the number of ensemble members.
+        target: ground truth labels. Has to be the same shape as preds except for the ensemble member dimension, which
+                should be missing for the targets.
+
+    Return:
+        Tensor with CRPS
+    """
```
SkafteNicki, line 64: an example would be nice :]

torchmetrics/functional/regression/crps.py
```diff
@@ -0,0 +1,66 @@
...
+from torch import Tensor
+
+from torchmetrics.utilities.checks import _check_same_shape
+
+
+def _crps_update(preds: Tensor, target: Tensor) -> Tuple[int, Tensor, float, float]:
+    _check_same_shape(preds[:, 0, ], target)  # second dimension of preds should be number of ensemble members
+
+    batch_size = target.size()[0]
+    n_ensemble_members = preds.size()[1]
```
SkafteNicki, line 27: just to be consistent across the codebase
```suggestion
    batch_size = target.shape[0]
    n_ensemble_members = preds.shape[1]
```

SkafteNicki (commented):
Looks like a good starting point. Just some additional mission stuff:
* entry in changelog
* references in docs
* add bindings to `__init__` files

Borda:
@tmachnitzki  how is it going here, still a draft or ready for full review?

tmachnitzki:
Hey, thanks for reminding me. I was a bit busy recently, but I might have time finishing this next week. So at the moment it still is a draft, unfortunately.

Borda:
> I was a bit busy recently, but I might have time finishing this next week. So at the moment it still is a draft, unfortunately.

that would be great to have it landed this week :]

Borda:
@tmachnitzki we are aiming at release by end of this week, do you think we can get this land by that time?

PR #243, opened 2021-05-12
Fix dtype incorrectly being reset

### LINKED ISSUE #187: Precision Loss in ConfusionMatrix
## 🐛 Bug

In Pytorch, `torch.zeros` and `.float()` methods create or convert a tensor to be a `torch.float32` tensor. Compare to the default `np.float64` data type in NumPy, it cannot provide enough precision. In `ConfusionMatrix`, the default values are torch.float32, which leads to wrong calculation (although maybe very little) against ground truth.

I think we only need to change `torch.zeros(num_classes, num_classes)` to `torch.zeros(num_classes, num_classes, dtype=torch.float64)`. I believe this change introduces negligible memory cost.

### To Reproduce

```python
>>> torch.tensor(90170814.).item()
90170816.0
```

### Expected Behaviour

```python
>>> torch.tensor(90170814., dtype=torch.float64).item()
90170814.0
```

### Additional context

`ConfusionMatrix` is the base of many other metrics, such as `IoU`. I found this problem by comparing the results to another NumPy-based metric in a semantic segmentation task using `IoU` as the metric. I think in today's deep learning dataset, float32 precision is not enough for the implementation of metrics.

Based on the above observation, maybe all `PyTorchLightning/metrics` should be implemented based on `torch.float64`

SkafteNicki:
# Before submitting

## What does this PR do?
Fixes #187
Currently, if a module is casted to another dtype, say double
```python
c = ConfusionMatrix(num_classes=3).double()
```
the initial metric states are correctly moved to `double` dtype. However, since currently is not applied to the `_defaults` (which contains the values we reset metric states to) after calling `reset` once the states will be back to whatever dtype they where original initialized with in `self.add_state`.
This PR fixes this and a bug in relation to this inside `ConfusionMatrix` that always forces the input to be float (even if asking for double).

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - AUC can also support more dimensional inputs when all but one dimensions are of size 1 ([#242](https://github.com/PyTorchLightning/metrics/pull/242))
 
-
+- Fixed dtype of modular metrics after reset have been called ([]())
 
 ## [0.3.2] - 2021-05-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - AUC can also support more dimensional inputs when all but one dimensions are of size 1 ([#242](https://github.com/PyTorchLightning/metrics/pull/242))
 
-- Fixed dtype of modular metrics after reset have been called ([]())
+- Fixed dtype of modular metrics after reset have been called ([#243](https://github.com/PyTorchLightning/metrics/pull/243))
 
 ## [0.3.2] - 2021-05-10
 
--- a/torchmetrics/functional/classification/confusion_matrix.py
+++ b/torchmetrics/functional/classification/confusion_matrix.py
@@ -47,8 +47,8 @@ def _confusion_matrix_compute(confmat: Tensor, normalize: Optional[str] = None) 
     allowed_normalize = ('true', 'pred', 'all', 'none', None)
     assert normalize in allowed_normalize, \
         f"Argument average needs to one of the following: {allowed_normalize}"
-    confmat = confmat.float() if not confmat.is_floating_point() else confmat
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - AUC can also support more dimensional inputs when all but one dimensions are of size 1 ([#242](https://github.com/PyTorchLightning/metrics/pull/242))
 
-
+- Fixed dtype of modular metrics after reset have been called ([]())
```
SkafteNicki, line 35:
```suggestion
- Fixed dtype of modular metrics after reset have been called ([#243](https://github.com/PyTorchLightning/metrics/pull/243))
```

Borda (approved):

torchmetrics/functional/classification/confusion_matrix.py
```diff
@@ -48,7 +47,7 @@ def _confusion_matrix_compute(confmat: Tensor, normalize: Optional[str] = None)
     allowed_normalize = ('true', 'pred', 'all', 'none', None)
     assert normalize in allowed_normalize, \
         f"Argument average needs to one of the following: {allowed_normalize}"
-    confmat = confmat.float()
+    confmat = confmat.float() if not confmat.is_floating_point() else confmat
```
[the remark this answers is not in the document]
SkafteNicki: true going to correct it :]

SkafteNicki, line 50:
```suggestion

```

torchmetrics/functional/classification/confusion_matrix.py
```diff
@@ -48,7 +47,7 @@ def _confusion_matrix_compute(confmat: Tensor, normalize: Optional[str] = None)
...
     if normalize is not None and normalize != 'none':
         cm = None
```
SkafteNicki, line 52:
```suggestion
        confmat = confmat.float() if not confmat.is_floating_point() else confmat
        cm = None
```

maximsch2 (approved):

PR #249, opened 2021-05-14
Add RMSE option to MSE code

### LINKED ISSUE #250: Add option to compute root_mean_squared_error
## 🚀 Feature
Allow the user to choose between MSE and RMSE.

### Motivation
In a physical domain the RMSE, which is essentially the mean of distances, may be significantly more intuitive than the MSE. Therefore, it would be nice to have the option to choose the preferd metric.

### Pitch
Similar to the implementation in [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html#sklearn.metrics.mean_squared_error) one could simply pass `squared=False` to the `MeanSquaredError` module or the `mean_squared_error` function.

johannespitz:
## What does this PR do?
Fixes #250 .

## PR review
I'm open for any kind of feedback. This is a very simple fix and I think it might be helpful for others to have this functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for unnormalized scores (e.g. logits) in `Accuracy`, `Precision`, `Recall`, `FBeta`, `F1`, `StatScore`, `Hamming`, `ConfusionMatrix` metrics ([#200](https://github.com/PyTorchLightning/metrics/pull/200))
 
 
+- Added `squared` argument to `MeanSquaredError` for computing `RMSE` ([#249](https://github.com/PyTorchLightning/metrics/pull/249))
+
+
 ### Changed
 
 
--- a/tests/regression/test_mean_error.py
+++ b/tests/regression/test_mean_error.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from collections import namedtuple
 from functools import partial
+import math
 
 import pytest
 import torch
@@ -43,16 +44,18 @@ _multi_target_inputs = Input(
 )
 
 
-def _single_target_sk_metric(preds, target, sk_fn=mean_squared_error):
+def _single_target_sk_metric(preds, target, sk_fn, metric_args):
     sk_preds = preds.view(-1).numpy()
     sk_target = target.view(-1).numpy()
-    return sk_fn(sk_preds, sk_target)
+    res = sk_fn(sk_preds, sk_target)
+    return math.sqrt(res) if (metric_args and not metric_args['squared']) else res
 
 
-def _multi_target_sk_metric(preds, target, sk_fn=mean_squared_error):
+def _multi_target_sk_metric(preds, target, sk_fn, metric_args):
     sk_preds = preds.view(-1, num_targets).numpy()
     sk_target = target.view(-1, num_targets).numpy()
-    return sk_fn(sk_preds, sk_target)
+    res = sk_fn(sk_preds, sk_target)
+    return math.sqrt(res) if (metric_args and not metric_args['squared']) else res
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


SkafteNicki:
Hi @johannespitz,
As RMSE is just sqrt(MSE) the implementation needs to happen in the `compute` function in instead of the `update`. Essentially, it should be:
```python
return sum_squared_error / n_obs if squared else torch.sqrt(sum_squared_error / n_obs)
```
instead of
```python
return sum_squared_error / n_obs
```

maximsch2:
Should this be just a CompositeMetric(MSE, torch.sqrt)?

johannespitz:
Hi,

unfortunately RMSE is not simply the square root of the MSE. Therefore, one can not compose the metric with some arithmetic.

```
import torch

targets = torch.tensor([1.0, 2, 3, 4, -1, -2])
preditions = torch.tensor([0, 1, 5, 2, 0, -2])

lenght = targets.shape[0]

mse = ((targets - preditions) ** 2).sum()
rmse = (((targets - preditions) ** 2) ** 0.5).sum()

print(f"mse {mse / lenght}")
print(f"rmse {rmse / lenght}")
print(f"mse**0.5 {mse ** 0.5 / lenght}")
```

The square root needs to be taken before summing up the individual errors.

maximsch2:
Is that actually true? Wikipedia gives RMSE = sqrt(MSE): https://en.wikipedia.org/wiki/Root-mean-square_deviation

johannespitz:
Oh, my bad. scikit-learn agrees with that definition. 
In this case, I agree that there is no need to modify the existing code to implement RSME.
I still believe that the metric I implemented is more useful for anyone working with physical distances, but I understand 100% if you prefer not to clutter your API with such (apparently niche) additions.
Therefore, please feel free to close the PR and the issue.

SkafteNicki:
I am actually in favour of adding this argument, as RMSE also is considered a standard machine learning algorithm.
The reason I think we should not just do `CompositionalMetric(MSE, torch.sqrt)` as @maximsch2 proposes is that `CompositionalMetric` is really not meant to be used by users (it is not in the documentation) but more as a tool to define metric arithmetics.
@johannespitz would you still be up for making the required changes?

johannespitz:
Sorry, about all those commits. But I have tested it, and now I'm getting the expected result. 
However, `torchmetrics.MeanSquaredError() ** 0.5` is documented and results in the same output...

SkafteNicki (approved):

Borda (approved):

maximsch2:
lgtm to me as well, but agree that the extra value might be slim compared to just `metric ** 0.5`

I see your point about root of each observation. We are getting a few metrics like that recently and it might be worth trying to combine/generalize them. The patter is that we want to have 

`sum(metric(pred_i, target_i))/N`
And so far we have `metric`:
- MSE: `(pred_i - target_i)**2`
- MAE: `torch.abs(pred_i-target_i)`
- What you asked: `torch.sqrt((pred_i - target_i)**2)` - if you look at it, it's actually exactly MAE, we already have it and you can use it! @johannespitz 
- MeanPercentageError: `torch.abs((pred_i-target_i)/target_i)`

I do wonder if we will exhaust all of those or we should make a generic metric implementation that will compute average "distance" between pred and target for a pluggable distance.

johannespitz:
@maximsch2 I like the idea of a generic class which can handle all metrics that fit into this `sum(metric(pred_i, target_i))/N` pattern. All the current metrics could just inherit from this class and specify the function which takes in single predictions and targets.
```
class torchmetrics.MeanSquaredError(AverageMetric):
    def __init__(self):
        super().__init__(
            individual_metric_fn=lambda pred, targ: (pred - targ) ** 2
        )
```
This would also allow me to implement what I wanted (but didn't even implement as you pointed out above) `individual_metric_fn=lambda pred, targ: troch.norm(pred - targ, dim=-1)`

Issue #67, opened 2021-03-13
Offer a dedicated sync() interface on the base Metric class


ananthsub:
## 🚀 Feature
Offer a dedicated `sync()` interface on the base Metric class. This would consolidate state across a provided process group using a given `dist_sync_fn` and would let us deprecate the `dist_sync_on_step` flag on the metric constructor.

### Motivation

The reason we'd like this is to decouple metric computation and global syncing. As a result, we'd be able to inspect both the local metric state separately from the synced state. 
Example scenario:
- We're training on a large number of nodes
- We wish to create a metric to track the local state during training steps, as syncing each step will be incredibly expensive
- At the end of the epoch, we want to sync the state once and log this value.

This interface also enables the training framework to offer higher-level APIs that could automatically call `sync()` for a particular `Metric` at relevant spots in the training loop (e.g. on_step, or on_epoch in Lightning).

cc @maximsch2 

### Pitch

We should be able to re-use most of `_sync_dist()` already. 

### Alternatives
Keep as is

PR #302, opened 2021-06-17
[feat] Add sync_context and sync to nn.Metric

tchaton:
## What does this PR do?

This PR add a more generic `_apply_sync` function to nn.Metric base class.

Fixes: https://github.com/PyTorchLightning/metrics/issues/67

Reasoning:
User might want to perform a reduction but not perform compute.

Current use case:
In Lightning, we are enabling restart in mid-epoch.
To do this, we need to save the synchronised states across process on rank 0. 
Therefore, the compute call is just an overhead and should be skipped.

## Did you have fun?


SkafteNicki:
Hi @tchaton,
Seems related to https://github.com/PyTorchLightning/metrics/issues/67 so maybe this should be a public method?

torchmetrics/metric.py
```diff
@@ -221,6 +222,46 @@ def wrapped_func(*args, **kwargs):
...
+            yield
+            return
+
+        if dist_sync_fn is None:
+            # User provided a bool, so we assume DDP if available
+            dist_sync_fn = gather_all_tensors
+
+        synced = False
+        cache = []
+        if self._to_sync and dist_sync_fn is not None:
```
justusschock, line 250:
```suggestion
        if is_distributed and self._to_sync and dist_sync_fn is not None:
```

justusschock:
```suggestion
```
removed in favor of the changed condition below

torchmetrics/metric.py
```diff
@@ -341,16 +378,17 @@ def persistent(self, mode: bool = False):
...
-            if self._persistent[key]:
-                current_val = getattr(self, key)
-                if not keep_vars:
-                    if torch.is_tensor(current_val):
-                        current_val = current_val.detach()
-                    elif isinstance(current_val, list):
-                        current_val = [cur_v.detach() if torch.is_tensor(cur_v) else cur_v for cur_v in current_val]
-                destination[prefix + key] = current_val
-        return destination
+        with self._apply_sync(dist_sync_fn=self.dist_sync_fn):
```
justusschock, line 381: Should we also do this in `__getstate__`?

tchaton: We can't as it breaks all tests :) __getstate__ is used by `deepcopy` and therefore it will perform reduction and won't create the same clone.

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
```diff
@@ -221,6 +222,46 @@ def wrapped_func(*args, **kwargs):
 
         return wrapped_func
 
+    @contextmanager
+    def _apply_sync(
```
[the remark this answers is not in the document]
tchaton: Yes.

torchmetrics/metric.py
```diff
@@ -221,6 +231,72 @@ def wrapped_func(*args, **kwargs):
...
         return wrapped_func
 
+    def sync(
+        self,
+        dist_sync_fn: Optional[Callable] = None,
+        process_group: Optional[Any] = None,
+        should_sync: bool = True,
+    ) -> Dict[str, Tensor]:
+        """
+        Sync function to control
```
SkafteNicki, line 241:
```suggestion
        Sync function for manually controlling when metrics states should be synced across processes
```

SkafteNicki (approved):
LGTM

justusschock (approved):
LGTM

tests/bases/test_ddp.py
```diff
@@ -116,3 +119,54 @@ def compute(self):
...
+    metric.persistent(True)
+
+    for i in range(5):
+        metric(i)
+        state_dict = metric.state_dict()
+
+        assert state_dict["x"] == sum(range(i + 1)) * 2
+        assert metric.x == sum(range(i + 1))
+
+        assert state_dict["c"] == (i + 1) * 2
```
carmocca, line 151: noob q: the * 2 is for the world size?

tests/bases/test_ddp.py
```diff
@@ -116,3 +119,54 @@ def compute(self):
...
+
+    metric = DummyCatMetric()
+    metric.persistent(True)
+
+    for i in range(5):
+        metric(i)
+        state_dict = metric.state_dict()
+
+        assert state_dict["x"] == sum(range(i + 1)) * 2
+        assert metric.x == sum(range(i + 1))
```
carmocca, line 149: sum(range(n + 1)) = n * (n + 1) / 2

```suggestion
        assert metric.x == steps * (steps + 1) / 2
        assert state_dict["x"] == metric.x * 2
```

tests/bases/test_ddp.py
```diff
@@ -116,3 +119,54 @@ def compute(self):
...
+
+        assert state_dict["c"] == (i + 1) * 2
+        assert metric.c == (i + 1)
```
carmocca, line 152:
```suggestion
        assert metric.c == (i + 1)
        assert state_dict["c"] == metric.c * 2
```

torchmetrics/metric.py
```diff
@@ -181,27 +187,31 @@ def forward(self, *args, **kwargs):
...
             Tensor,
             dist_sync_fn,
-            group=self.process_group,
+            group=process_group or self.process_group,
         )
 
         for attr, reduction_fn in self._reductions.items():
             # pre-processing ops (stack or flatten for inputs)
-            if isinstance(output_dict[attr][0], Tensor):
+            if isinstance(output_dict[attr][0], Tensor) and isinstance(output_dict[attr], Sequence):
```
carmocca, line 214: Maybe switch the order? This way if `output_dict[attr]` is not a sequence, we don't try to access the 0-th index

```suggestion
            if isinstance(output_dict[attr], Sequence) and isinstance(output_dict[attr][0], Tensor):
```

torchmetrics/metric.py
```diff
@@ -221,6 +231,72 @@ def wrapped_func(*args, **kwargs):
 
         return wrapped_func
 
+    def sync(
+        self,
+        dist_sync_fn: Optional[Callable] = None,
+        process_group: Optional[Any] = None,
+        should_sync: bool = True,
```
carmocca, line 238: Do we need `op`?

torchmetrics/metric.py
```diff
@@ -221,6 +231,72 @@ def wrapped_func(*args, **kwargs):
...
+            dist_sync_fn: Function to be used to perform states synchronization
+            process_group:
+                Specify the process group on which synchronization is called.
+                default: None (which selects the entire world)
+            should_sync: Whether to apply to state synchronization.
+
+        Returns:
+            cache: A dictionarry containing the local metric states. The cache will be empty if sync didn't happen.
+        """
+        is_distributed = torch.distributed.is_available() and torch.distributed.is_initialized()
```
carmocca, line 253: Can you move this to a function? Will need to override it for TPUs or other accelerators

torchmetrics/metric.py
```diff
@@ -221,6 +231,72 @@ def wrapped_func(*args, **kwargs):
...
+
+        if dist_sync_fn is None:
+            # User provided a bool, so we assume DDP if available
```
carmocca, line 256: I don't see any bool checked here

torchmetrics/metric.py
```diff
@@ -221,6 +231,72 @@ def wrapped_func(*args, **kwargs):
...
+            dist_sync_fn = gather_all_tensors
+
+        cache = {}
+
+        if is_distributed and should_sync and dist_sync_fn is not None:
```
carmocca, line 261: We know it's not `None` at this point as it would have been overridden a few lines before

```suggestion
        if is_distributed and should_sync:
```

carmocca: Should we early exit if `not should_sync`?

torchmetrics/metric.py
```diff
@@ -363,10 +439,15 @@ def _load_from_state_dict(
...
         """ Loads metric states from state_dict """
+
+        # only global rank 0 should be reloading the values present in the ``state_dict``
+        # as the state contains synced values across all progress_group
         for key in self._defaults:
             name = prefix + key
             if name in state_dict:
-                setattr(self, key, state_dict.pop(name))
+                value = state_dict.pop(name)
+                if os.getenv("GLOBAL_RANK", "0") == "0":
```
carmocca, line 449: Should we put this check inside a function? In case it needs to be overridden

torchmetrics/metric.py
```diff
@@ -341,16 +416,17 @@ def persistent(self, mode: bool = False):
...
-                    elif isinstance(current_val, list):
-                        current_val = [cur_v.detach() if torch.is_tensor(cur_v) else cur_v for cur_v in current_val]
-                destination[prefix + key] = current_val
-        return destination
+        with self.sync_context(dist_sync_fn=self.dist_sync_fn):
+            for key in self._defaults:
+                if self._persistent[key]:
+                    current_val = getattr(self, key)
+                    if not keep_vars:
+                        if torch.is_tensor(current_val):
```
carmocca, line 424: nit: I know it was in the old code but

```suggestion
                        if isinstance(current_val, torch.Tensor):
```

torchmetrics/metric.py
```diff
@@ -341,16 +416,17 @@ def persistent(self, mode: bool = False):
...
+                            current_val = current_val.detach()
+                        elif isinstance(current_val, list):
+                            current_val = [cur_v.detach() if torch.is_tensor(cur_v) else cur_v for cur_v in current_val]
```
carmocca, line 427:
```suggestion
                            current_val = [cur_v.detach() if isinstance(cur_v, torch.Tensor) else cur_v for cur_v in current_val]
```

torchmetrics/metric.py
```diff
@@ -341,16 +416,17 @@ def persistent(self, mode: bool = False):
...
+                    destination[prefix + key] = deepcopy(current_val)
```
carmocca, line 428: Why do we need the deepcopy now?

Borda (approved):

torchmetrics/metric.py
```diff
@@ -363,10 +439,15 @@ def _load_from_state_dict(
...
     ) -> None:
         """ Loads metric states from state_dict """
+
+        # only global rank 0 should be reloading the values present in the ``state_dict``
+        # as the state contains synced values across all progress_group
         for key in self._defaults:
             name = prefix + key
             if name in state_dict:
-                setattr(self, key, state_dict.pop(name))
+                value = state_dict.pop(name)
```
carmocca, line 448: This will modify the passed state dict, do we want that?

torchmetrics/metric.py
```diff
@@ -331,6 +392,20 @@ def _apply(self, fn):
                 )
         return this
 
+    @contextmanager
+    def _apply_persistent(
```
carmocca, line 396: Where will this be used? Don't see it anywhere

tests/bases/test_ddp.py
```diff
@@ -116,3 +119,54 @@ def compute(self):
...
+            self.x += x
+            self.c += 1
+
+        def compute(self):
+            return self.x / self.c
+
+    metric = DummyCatMetric()
+    metric.persistent(True)
+
+    for i in range(5):
```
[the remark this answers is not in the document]
tchaton: Done !

tests/bases/test_ddp.py
tchaton: Yes, replace by world_size

tests/bases/test_ddp.py
tchaton: ahahah :) Done !

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: Yes, good catch !

torchmetrics/metric.py
tchaton: removed !

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: As the state dict is being synced across processes, and we don't want to keep the reference.

torchmetrics/metric.py
tchaton: Yes, as it would fail otherwise with unexpected parameters.

torchmetrics/metric.py
tchaton: Removed 1

torchmetrics/metric.py
tchaton: No need as `reduction_fn` is automatically inferred from `self._reductions`

torchmetrics/metric.py
carmocca: maybe copy the state_dict first then?

Otherwise, the user will be left out with a different `state_dict` than the one he passed

tests/bases/test_ddp.py
```diff
@@ -116,3 +119,56 @@ def compute(self):
...
+            return self.x / self.c
+
+    metric = DummyCatMetric()
+    metric.persistent(True)
+
+    steps = 5
+    for i in range(steps):
+        metric(i)
+        state_dict = metric.state_dict()
+        print(state_dict)
```
carmocca, line 148:
```suggestion
```

torchmetrics/metric.py
```diff
@@ -28,6 +30,10 @@
 from torchmetrics.utilities.imports import _LIGHTNING_AVAILABLE, _compare_version
 
 
+def is_distributed_fn() -> bool:
```
carmocca, line 33: prefer this name but no strong opinion

```suggestion
def distributed_available() -> bool:
```

would need to update the other instances

torchmetrics/metric.py
```diff
@@ -221,6 +235,80 @@ def wrapped_func(*args, **kwargs):
...
+
+        cache = {}
+
+        if is_distributed and should_sync:
+            # cache prior to syncing
+            cache = {attr: getattr(self, attr) for attr in self._defaults.keys()}
+
+            # sync
+            self._sync_dist(dist_sync_fn, process_group=process_group)
+
```
carmocca, line 272:
```suggestion
        is_distributed = is_distributed_fn()
        if not should_sync or not is_distributed:
            return {}

        if dist_sync_fn is None:
            dist_sync_fn = gather_all_tensors
        # cache prior to syncing
        cache = {attr: getattr(self, attr) for attr in self._defaults.keys()}
        # sync
        self._sync_dist(dist_sync_fn, process_group=process_group)
```

torchmetrics/metric.py
```diff
@@ -221,6 +235,80 @@ def wrapped_func(*args, **kwargs):
...
+        Args:
+            dist_sync_fn: Function to be used to perform states synchronization
+            process_group:
+                Specify the process group on which synchronization is called.
+                default: None (which selects the entire world)
+            should_sync: Whether to apply to state synchronization.
+            is_distributed_fn: Function to determine if we are running inside a distributed setting
+
+        Returns:
+            cache: A dictionarry containing the local metric states. The cache will be empty if sync didn't happen.
```
carmocca, line 257:
```suggestion
            cache: A dictionary containing the local metric states. The cache will be empty if sync didn't happen.
```

torchmetrics/metric.py
```diff
@@ -341,16 +414,24 @@ def persistent(self, mode: bool = False):
...
+                            current_val = [
+                                cur_v.detach() if isinstance(cur_v, torch.Tensor) else cur_v for cur_v in current_val
+                            ]
+                    destination[prefix + key] = deepcopy(current_val)
+            return destination
+
+    def _on_load_from_state_dict(self, state_dict, key, name) -> None:
+        value = state_dict.pop(name)
+        if os.getenv("GLOBAL_RANK", "0") == "0":
+            setattr(self, key, value)
```
carmocca, line 434: with my previous comment I meant adding a hook for the

`if os.getenv("GLOBAL_RANK", "0") == "0"`

check, not the whole thing :D

tchaton: I think it is fine :)

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: Done !

torchmetrics/metric.py
tchaton: Done !

carmocca (approved):

torchmetrics/metric.py
carmocca:
```suggestion
                    # the tensors will be synced across processes so deepcopy to drop the references
                    destination[prefix + key] = deepcopy(current_val)
```

Borda:
seems constantly failing on PT 1.6
```
E       Exception: 
E       
E       -- Process 0 terminated with the following error:
E       Traceback (most recent call last):
E         File "/usr/share/miniconda3/envs/test/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 20, in _wrap
E           fn(i, *args)
E         File "/home/runner/work/metrics/metrics/tests/bases/test_ddp.py", line 146, in _test_state_dict_is_synced
E           metric(i)
E         File "/usr/share/miniconda3/envs/test/lib/python3.7/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
E           result = self.forward(*input, **kwargs)
E         File "/home/runner/work/metrics/metrics/torchmetrics/metric.py", line 189, in forward
E           self._forward_cache = self.compute()
E         File "/home/runner/work/metrics/metrics/torchmetrics/metric.py", line 326, in wrapped_func
E           self._computed = compute(*args, **kwargs)
E         File "/home/runner/work/metrics/metrics/tests/bases/test_ddp.py", line 139, in compute
E           return self.x / self.c
E       RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
```

tchaton:
> seems constantly failing on PT 1.6
> 
> ```
> E       Exception: 
> E       
> E       -- Process 0 terminated with the following error:
> E       Traceback (most recent call last):
> E         File "/usr/share/miniconda3/envs/test/lib/python3.7/site-packages/torch/multiprocessing/spawn.py", line 20, in _wrap
> E           fn(i, *args)
> E         File "/home/runner/work/metrics/metrics/tests/bases/test_ddp.py", line 146, in _test_state_dict_is_synced
> E           metric(i)
> E         File "/usr/share/miniconda3/envs/test/lib/python3.7/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
> E           result = self.forward(*input, **kwargs)
> E         File "/home/runner/work/metrics/metrics/torchmetrics/metric.py", line 189, in forward
> E           self._forward_cache = self.compute()
> E         File "/home/runner/work/metrics/metrics/torchmetrics/metric.py", line 326, in wrapped_func
> E           self._computed = compute(*args, **kwargs)
> E         File "/home/runner/work/metrics/metrics/tests/bases/test_ddp.py", line 139, in compute
> E           return self.x / self.c
> E       RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.
> ```

Thanks. Resolved.

Borda:
@tchaton any reason/justification why the tests take more than an extra 5min?

tchaton:
> @tchaton any reason/justification why the tests take more than an extra 5min?

Great question. I am investigating. @SkafteNicki @justusschock Any idea ?

Best,
T.C

tests/bases/test_ddp.py
```diff
@@ -116,3 +119,55 @@ def compute(self):
...
+
+    metric = DummyCatMetric()
+    metric.persistent(True)
+
+    steps = 5
+    for i in range(steps):
+        metric(i)
+        state_dict = metric.state_dict()
+
+        sum = i * (i + 1) / 2
```
awaelchli, line 149: smart! cool test

awaelchli (approved):

torchmetrics/metric.py
```diff
@@ -341,16 +409,23 @@ def persistent(self, mode: bool = False):
...
+                        elif isinstance(current_val, list):
+                            current_val = [
+                                cur_v.detach() if isinstance(cur_v, torch.Tensor) else cur_v for cur_v in current_val
+                            ]
+                    # the tensors will be synced across processes so deepcopy to drop the references
+                    destination[prefix + key] = deepcopy(current_val)
+            return destination
+
+    def _should_load_from_state_dict(self) -> bool:
+        return os.getenv("GLOBAL_RANK", "0") == "0"
```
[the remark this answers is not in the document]
tchaton: This is used for Lightning. We agreed with Justus this would be the cleanest approach to force reload only on global rank 0.

torchmetrics/metric.py
```diff
@@ -221,6 +235,75 @@ def wrapped_func(*args, **kwargs):
 
         return wrapped_func
 
+    def sync(
+        self,
+        dist_sync_fn: Optional[Callable] = None,
+        process_group: Optional[Any] = None,
+        should_sync: bool = True,
```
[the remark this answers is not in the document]
tchaton: `should_sync` means should_sync if possible :) Modified the docstring to reflect this.

tests/bases/test_ddp.py
tchaton: @carmocca doing :)

torchmetrics/metric.py
tchaton: This behaviour was already present in nn.Metric and should be changed in another PR. But I would argue this seems to be a common way to do.

torchmetrics/metric.py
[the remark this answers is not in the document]
awaelchli: It means now that there are two arguments overlapping: `dist_sync_fn` and `should_sync`

You can do this: `should_sync=False` and `dist_sync_fn=mean`

what willl happen now? will it sync or not?
@PyTorchLightning/core-metrics be aware of these cases

maximsch2:
Late to the party here, but I think we can also imagine the future where models are huge and sharded (with FSDP) and metric states are similarly sharded. We are getting away with synchronizing everything on rank0 for now but long-term we might need to have metrics that wont' be able to do that

torchmetrics/metric.py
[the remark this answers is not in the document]
maximsch2: Yeah, I agree. I wonder if the main usage of `should_sync` is just in `sync_context` and maybe we should just decide there if syncing is needed or not? Doing an `if` with context manager is a bit harder and might justify a flag, but for a function, it should be easy for people to just avoid calling it?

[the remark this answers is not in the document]
carmocca: @maximsch2 your argument is to keep the flag for the context manager but remove it from this function, correct?

I think that would be fine.

torchmetrics/metric.py
```diff
@@ -221,6 +234,77 @@ def wrapped_func(*args, **kwargs):
...
+        distributed_available: Optional[Callable] = distributed_available,
```
ananthsub, line 242: n00b question: why does distributed available need to be an argument?

torchmetrics/metric.py
```diff
@@ -221,6 +234,77 @@ def wrapped_func(*args, **kwargs):
...
+            distributed_available=distributed_available
+        )
+
+        yield
+
+        if cache and restore_cache:
+            # if we synced, restore to cache so that we can continue to accumulate un-synced state
+            for attr, val in cache.items():
+                setattr(self, attr, val)
+
```
ananthsub, line 307: What's the use case for this? If we sync, we should assume all metrics are operating off the synced state and not accumulate local changes, right?

torchmetrics/metric.py
```diff
@@ -181,27 +190,31 @@ def forward(self, *args, **kwargs):
...
             Tensor,
             dist_sync_fn,
-            group=self.process_group,
+            group=process_group or self.process_group,
         )
 
         for attr, reduction_fn in self._reductions.items():
             # pre-processing ops (stack or flatten for inputs)
-            if isinstance(output_dict[attr][0], Tensor):
+            if isinstance(output_dict[attr], Sequence) and isinstance(output_dict[attr][0], Tensor):
```
ananthsub, line 217: this check is not safe. we're seeing errors as a result.
```
if isinstance(output_dict[attr], Sequence) and isinstance(output_dict[attr][0], Tensor):

IndexError: list index out of range
```

ananthsub (commented):
@tchaton I had a different idea when proposing `sync`: the Metric subclass should provide the implementation, not the base. The metric states are updated in-place instead of returned to the caller. So one could chain together calls to `update` and `sync` before finally calling `compute` to get the state.

torchmetrics/metric.py
carmocca: Because `torchmetrics` does not know about any distributed platforms other than CUDA

For example TPU, IPUs...

torchmetrics/metric.py
carmocca: Removed in https://github.com/PyTorchLightning/metrics/pull/311

torchmetrics/metric.py
carmocca: This was here already, just moved.

Added in https://github.com/PyTorchLightning/metrics/commit/dd1e744efec0888ddb93abb322d3bc5a4614a540
cc: @SkafteNicki

PR #327, opened 2021-06-29
extend typing: base

Borda:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


torchmetrics/metric.py
```diff
@@ -481,107 +481,108 @@ def __hash__(self):
...
-    def __add__(self, other: Any):
+    def __add__(self, other: "Metric") -> "Metric":
         return CompositionalMetric(torch.add, self, other)
 
-    def __and__(self, other: Any):
+    def __and__(self, other: "Metric") -> "Metric":
         return CompositionalMetric(torch.bitwise_and, self, other)
 
-    def __eq__(self, other: Any):
+    # Fixme: this shall return bool instead of Metric
```
Borda, line 490: @justusschock any idea how we could fix this?

torchmetrics/metric.py
```diff
@@ -92,20 +93,24 @@ def __init__(
...
         self._forward_cache = None
         self._update_called = False
 
         # initialize state
-        self._defaults = {}
-        self._persistent = {}
-        self._reductions = {}
+        self._defaults: dict = {}
+        self._persistent: dict = {}
+        self._reductions: dict = {}
```
justusschock, line 105:
```suggestion
        self._defaults: Dict[str, Union[List, Tensor]] = {}
        self._persistent: Dict[str, bool] = {}
        self._reductions: Dict[str, Union[str, Callable[[Union[List, Tensor]], Tensor]] = {}
```

torchmetrics/metric.py
```diff
@@ -224,10 +229,10 @@ def _sync_dist(self, dist_sync_fn: Callable = gather_all_tensors, process_group:
             reduced = reduction_fn(output_dict[attr]) if reduction_fn is not None else output_dict[attr]
             setattr(self, attr, reduced)
 
-    def _wrap_update(self, update):
+    def _wrap_update(self, update: Callable) -> Callable:
 
         @functools.wraps(update)
-        def wrapped_func(*args, **kwargs):
+        def wrapped_func(*args: Tensor, **kwargs: Any) -> Optional[Any]:
```
justusschock, line 235:
```suggestion
        def wrapped_func(*args: Any, **kwargs: Any) -> Optional[Any]:
```

This probably has to match the signature of `update`

torchmetrics/metric.py
```diff
@@ -305,10 +310,10 @@ def sync_context(
             for attr, val in cache.items():
                 setattr(self, attr, val)
 
-    def _wrap_compute(self, compute):
+    def _wrap_compute(self, compute: Callable) -> Callable:
 
         @functools.wraps(compute)
-        def wrapped_func(*args, **kwargs):
+        def wrapped_func(*args: Any, **kwargs: Any) -> Any:
```
justusschock, line 316:
```suggestion
        def wrapped_func() -> Any:
```

Can also match `compute` signature

torchmetrics/metric.py
```diff
@@ -305,10 +310,10 @@ def sync_context(
             for attr, val in cache.items():
                 setattr(self, attr, val)
 
-    def _wrap_compute(self, compute):
+    def _wrap_compute(self, compute: Callable) -> Callable:
```
justusschock, line 313:
```suggestion
    def _wrap_compute(self, compute: Callable[Any]) -> Callable[Any]:
```

torchmetrics/metric.py
```diff
@@ -450,7 +459,7 @@ def _load_from_state_dict(
             state_dict, prefix, local_metadata, True, missing_keys, unexpected_keys, error_msgs
         )
 
-    def _filter_kwargs(self, **kwargs):
+    def _filter_kwargs(self, **kwargs: Dict[str, Any]) -> Dict[str, Any]:
```
justusschock, line 462:
```suggestion
    def _filter_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
```

kwargs are always of type Dict[str, X] so they are only annotated with X

torchmetrics/metric.py
justusschock: I discussed with @tchaton  in the past. One can argue for bool and for metric (the metric checks the equality of the outputs for each argument then). Equality would probably go via `is` or hashes.

So the current behaviour at least is consistent with the other arithmetic and logic operators.

torchmetrics/collections.py
```diff
@@ -207,7 +207,7 @@ def _to_renamed_ordered_dict(self) -> OrderedDict:
             od[self._set_name(k)] = v
         return od
 
-    def keys(self, keep_base: bool = False):
+    def keys(self, keep_base: bool = False) -> Iterable[str]:
```
justusschock, line 210:
```suggestion
    def keys(self, keep_base: bool = False) -> Iterable[Hashable]:
```

keys aren't necessarily strings

justusschock (approved):
I suggested some changes in typing. Maybe this means that we can get rid of the type ignores

torchmetrics/metric.py
Borda: yep, I see the arithmetic but feel that logic like eq shall return a bool, that is also why mypy complains as we overwrite something which shall return bool but return metric...

justusschock: How strict are we with deprecations in metrics? I'm fine with changing it. But I think checking the equality of metrics is not that easy. Probably it would require some syncing and checking the states.

tchaton (approved):
LGTM !

torchmetrics/metric.py
Borda: we keep 0.1 deprecation loop, but if needed we are fine with breaking changes

awaelchli (approved):

PR #328, opened 2021-06-29
fix ignore_index in the computation of IoU

### LINKED ISSUE #312: Incorrect results with ignore_index in IoU
## 🐛 Bug

The ignore_index option in IoU does not behave as it should

### To Reproduce

Run the above code

#### Code sample

```python
import torch
import torchmetrics

iou = torchmetrics.IoU(3, ignore_index=0)
print(iou(torch.tensor([[1,2,1,1]]), torch.tensor([[2,1,1,0]])))
print(iou(torch.tensor([[1,2,1,0]]), torch.tensor([[2,1,1,0]])))
> tensor(0.1250)
tensor(0.1667)
```

### Expected behavior

Predicting a '0' here should not modify the score, both should give 0.1667

### Environment

 - PyTorch Version (e.g., 1.0): 1.8.1
 - OS (e.g., Linux): Linux & Windows
 - How you installed PyTorch (`conda`, `pip`, source): both
 - Build command you used (if compiling from source):
 - Python version: 3.8
 - CUDA/cuDNN version: 
 - GPU models and configuration:
 - Any other relevant information: 

### Additional context

The ignore_index is implemented like [this](https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/functional/classification/iou.py#L39):
```
if ignore_index is not None and 0 <= ignore_index < num_classes:
    scores = torch.cat([
        scores[:ignore_index],
        scores[ignore_index + 1:],
    ])
```
This is incorrect as it does not provide the right union value. Instead, the confmat should be stripped of its ignore_index column and line.

This can be done with this line provided it is put before the computation of the intersection and union:
`confmat = confmat[torch.arange(confmat.size(0)) != ignore_index].T[[ignore_index != torch.arange(confmat.size(0))]].T`

CSautier:
# Before submitting

## What does this PR do?
Fixes #312

## PR review
The way it does is to remove the line and columns of the ignore_index in the confmat. It should work whenever ignore_index is in the correct range but could be tested further.

There are certainly cleaner ways to do this.


Borda (commented):
nice, can we add a test to verify it works as intended now?

Borda:
@CSautier mind check the failing tests?

Borda:
@CSautier how is it going? :rabbit:

CSautier:
Sorry, it slipped out of my mind.
I verified the tests, and I believe test_iou(True, 'none', 0, torch.Tensor([0.5, 0.5])) does not make sense. I think the result should indeed be (2/3, 1/2) instead of (1/2, 1/2), which would mean that the current version passing this test is wrong. Similarly test_iou_ignore_index([0, 1, 1, 2, 2], [0, 1, 2, 2, 2], 2, 3, 'none', [1, 1 / 2]) should indeed return (1,1) as my version does and not (1, 1/2) as the test expects.

However there is indeed one correct tests that my computation does not pass 
test_iou_ignore_index([0, 1, 1, 2, 2], [0, 1, 2, 2, 2], 1, 3, 'none', [1, 2 / 3])
and it shows a problem in how I remove the ignore_index if it is predicted.

I'm working on it now, and will change my PR if I find a solution.
If you could also verify that I'm correct and test_iou(True, 'none', 0, torch.Tensor([0.5, 0.5])) should indeed return (2/3, 1/2) and test_iou_ignore_index([0, 1, 1, 2, 2], [0, 1, 2, 2, 2], 2, 3, 'none', [1, 1 / 2]) should in fact return (1, 1) that would be great.

CSautier:
I've updated the PR, and included the two modifications I mentioned earlier in the tests.

SkafteNicki (approved):
LGTM :]

Borda (approved):

PR #332, opened 2021-06-30
typing: info retrieval

Borda:
# Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [x] Did you write any new necessary tests?

## What does this PR do?
Fixes # (issue).

## PR review
Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #333, opened 2021-06-30
typing: regression

Borda:
# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/torchmetrics/image/psnr.py
+++ b/torchmetrics/image/psnr.py
@@ -68,6 +68,8 @@ class PSNR(Metric):
         Half precision is only support on GPU for this metric
 
     """
+    min_target: Tensor
+    max_target: Tensor
 
     def __init__(
         self,
@@ -122,8 +124,8 @@ class PSNR(Metric):
         if self.dim is None:
             if self.data_range is None:
                 # keep track of min and max target values
-                self.min_target: Tensor = min(target.min(), self.min_target)
-                self.max_target: Tensor = max(target.max(), self.max_target)
+                self.min_target = min(target.min(), self.min_target)
+                self.max_target = max(target.max(), self.max_target)
 
             self.sum_squared_error += sum_squared_error
             self.total += n_obs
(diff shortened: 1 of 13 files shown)


Borda:
the same also here, most of the `mypy` issues are about
```
error: Signature of "update" incompatible with supertype "Metric"  [override]
```

torchmetrics/image/psnr.py
```diff
@@ -110,7 +110,7 @@ def __init__(
         self.reduction = reduction
         self.dim = tuple(dim) if isinstance(dim, Sequence) else dim
 
-    def update(self, preds: Tensor, target: Tensor) -> None:
+    def update(self, preds: Tensor, target: Tensor) -> None:  # type: ignore
```
justusschock, line 113: this is the same as in #332 right? So contradicting to what I wrote there maybe merge them and do a PR with a followup on all signatures if we found a solution?

torchmetrics/regression/explained_variance.py
```diff
@@ -113,11 +113,11 @@ def update(self, preds: Tensor, target: Tensor) -> None:
...
-        self.n_obs = self.n_obs + n_obs
-        self.sum_error = self.sum_error + sum_error
-        self.sum_squared_error = self.sum_squared_error + sum_squared_error
-        self.sum_target = self.sum_target + sum_target
-        self.sum_squared_target = self.sum_squared_target + sum_squared_target
+        self.n_obs: Tensor = self.n_obs + n_obs
+        self.sum_error: Tensor = self.sum_error + sum_error
+        self.sum_squared_error: Tensor = self.sum_squared_error + sum_squared_error
+        self.sum_target: Tensor = self.sum_target + sum_target
+        self.sum_squared_target: Tensor = self.sum_squared_target + sum_squared_target
```
justusschock, line 120: it's better to add the typing already outside (where you usually define static attributes) and not initialize it then typing lately here. 

So something like 

```python

class MyMetric:
    a: torch.Tensor
    def __init__(self):
        super().__init__()
        self.add_state('a', default=torch.tensor(0.))
```

torchmetrics/image/psnr.py
Borda: yes, want to make another PR to try to resolve this typing over all codebase

justusschock (approved):

PR #334, opened 2021-06-30
[bugfix] Improve sync states management for checkpointing

### LINKED ISSUE #331: Metric with ddp spawn causes script to hang after Trainer.fit()
## 🐛 Bug

Adding a torchmetric as an attribute to the model causes processes to hang when launching with ddp spawn. 

### To Reproduce

1. Install torchmetrics 0.4, PyTorch Lightning 0.3.7
2. Run the script below

#### Code sample

```python
import os

import torch
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer
from pytorch_lightning.metrics import Accuracy


class RandomDataset(Dataset):

    def __init__(self, size, length):
        self.len = length
        self.data = torch.randn(length, size)

    def __getitem__(self, index):
        return self.data[index]

    def __len__(self):
        return self.len


class BoringModel(LightningModule):

    def __init__(self):
        super().__init__()
        self.layer = torch.nn.Linear(32, 2)
        self.metric = Accuracy() # add this to break it all

    def forward(self, x):
        return self.layer(x)

    def training_step(self, batch, batch_idx):
        loss = self(batch).sum()
        self.log("train_loss", loss)
        return {"loss": loss}

    def configure_optimizers(self):
        return torch.optim.SGD(self.layer.parameters(), lr=0.1)


def run():
    train_data = DataLoader(RandomDataset(32, 64), batch_size=2)

    model = BoringModel()
    trainer = Trainer(
        default_root_dir=os.getcwd(),
        limit_train_batches=1,
        max_epochs=1,
        weights_summary=None,
        accelerator="ddp_spawn",
        gpus=2,
    )
    trainer.fit(model, train_dataloader=train_data)


if __name__ == '__main__':
    run()

```

**A few important notes:**
- All you have to do is `self.metric = Accuracy()`, then it will break. You don't have to use the metric, just assign it.
- Installing TM < 0.4 solves the problem

### Expected behavior

No hang.

### Environment

 - PyTorch Version (e.g., 1.0): 1.8.0
 - OS (e.g., Linux): Linux
 - How you installed PyTorch (`conda`, `pip`, source): pip 
 - Build command you used (if compiling from source):
 - Python version: 3.8
 - CUDA/cuDNN version: whatever comes with 1.8.0 torch
 - GPU models and configuration: our grid cluster
 - Any other relevant information: no

### Additional context

trying to prepare PL patch release: https://github.com/PyTorchLightning/pytorch-lightning/pull/8198

tchaton:
# Before submitting

Changes introduce in this PR:

Add `dist_sync_state_dict` argument to the Metric `__init__` object.
Add `should_sync` to state_dict()

Here is the new behaviour on state_dict:
```
If should_sync, then sync the state across processes
     * On rank 0, return the aggregated states 
     * On other ranks, return default
     * add metadata such as rank, world_size, has_synced

else return current state,
     * add metadata such as rank, world_size, has_synced

Here is the new behaviour on load_state_dict:
We support both 1 checkpoint for all ranks, or n checkpoint for n ranks. 

if not has_synced, then check that world_size matches.

if has_synced, then load on all ranks as non 0 ranks contains defaults.
else:
     # this assume 1 checkpoint shared for all. Only rank 0 should reload it
     if the previous rank used is 0
             reload only rank 0
     else:
           # this assume multiple checkpoints generated.
            reload for all 
```

Small caviats: Metadata are saved to tensors as DDP wrapper will try to broadcast them. 

Fixes #331

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?


docs/source/references/modules.rst
```diff
@@ -12,6 +12,25 @@ metrics.
...
+    This behaviour can be blocked in the following ways.
+    We strongly recommend calling :meth:`~torchmetrics.Metric.state_dict` on all processes or this can cause a deadlock.
+
+.. doctest::
+
+    >>> from torchmetrics import Metric
+    >>> metric = Metric(should_sync_state_dict=False)
+    >>> metric.state_dict()
+
+.. doctest::
```
justusschock, line 27: maybe explain the difference here or when to use which way?

justusschock (approved):
LGTM. Just having the same argument twice (once in init and once in state_dict) might be confusing to users without any more detailed  explanation/distinction.

docs/source/references/modules.rst
```diff
@@ -12,6 +12,25 @@ metrics.
 .. autoclass:: torchmetrics.Metric
     :noindex:
 
+.. note::
+
+    Metric states are being synchornized by default in a distributed setting when using :meth:`~torchmetrics.Metric.state_dict`.
```
SkafteNicki, line 17: Maybe it is worth mentioning that the state dict by default is empty because `persistent=False`

SkafteNicki (approved):

tchaton:
@justusschock @SkafteNicki,

Those changes were test against Lightning on this test: https://github.com/PyTorchLightning/pytorch-lightning/commit/189f0ad152cb8bb94e7c64a920ced512861fd0ec#diff-44594fac97ae772bb165301c22fdc1b25cd481589502dd7413c6953289cbb718L1344

Accessible under this Lightning commit: `6c50332`.

Best,
T.C

docs/source/references/modules.rst
tchaton: Done !

docs/source/references/modules.rst
tchaton: Done !

torchmetrics/metric.py
```diff
@@ -415,25 +419,60 @@ def state_dict(
...
+                destination[prefix + "world_size"] = torch.tensor(self.world_size, device=device)  # type: ignore
         return destination
 
-    def _should_load_from_state_dict(self) -> bool:
-        return os.getenv("GLOBAL_RANK", "0") == "0"
+    @property
+    def world_size(self) -> int:
+        if jit_distributed_available():
+            return torch.distributed.get_world_size()
+        return 0
```
awaelchli, line 465: you mean 1? ^^

torchmetrics/metric.py
```diff
@@ -66,6 +66,8 @@ class Metric(nn.Module, ABC):
         dist_sync_fn:
             Callback that performs the allgather operation on the metric state. When `None`, DDP
             will be used to perform the allgather.
+        should_sync_state_dict: Whether states should be synchronized while creating a checkpoint.
+            This would be part of the checkpoint and re-used when reloading the metric state_dict.
```
awaelchli, line 70: `should_sync_state_dict` does not follow the same naming convention as `compute_on_step` and `dist_sync_on_step`.
Does torchmetrics want to have a consistent naming?
RFC @PyTorchLightning/core-metrics

torchmetrics/metric.py
```diff
@@ -447,13 +486,28 @@ def _load_from_state_dict(
     ) -> None:
         """ Loads metric states from state_dict """
 
-        # only global rank 0 should be reloading the values present in the ``state_dict``
-        # as the state contains synced values across all progress_group
+        # assumption: users didn't manually sync the Metric states by default.
+        has_synced = state_dict.pop(prefix + "has_synced", None)
```
awaelchli, line 490: just to double check, is it ok to pop? 
the intention is not to re-use the state dict somewhere else?
could the user want to load the same state dict onto different metrics, like they could do in pytorch with model weights?

awaelchli (approved):

torchmetrics/metric.py
tchaton: ahaha thanks :)

torchmetrics/metric.py
tchaton: Great catch ! Changed to `dist_sync_state_dict`.

torchmetrics/metric.py
tchaton: Yes, you need to pop or it will raise an error.

tchaton: That's the expected behaviour from PyTorch. Checkpoint are being consumed.

docs/source/references/modules.rst
```diff
@@ -12,6 +12,47 @@ metrics.
 .. autoclass:: torchmetrics.Metric
     :noindex:
 
+.. note::
+
+    Metric states are being synchornized by default when ``persistent=True`` in a distributed setting
+    when using :meth:`~torchmetrics.Metric.state_dict`
+    This behaviour can be blocked in the following ways.
+    We strongly recommend calling :meth:`~torchmetrics.Metric.state_dict` on all processes or this can cause a deadlock.
```
carmocca, line 20: I don't think this should be a note, it's an important detail


```suggestion
Metric states are synchornized when ``persistent=True`` in a distributed setting
when calling :meth:`~torchmetrics.Metric.state_dict`
We strongly recommend calling :meth:`~torchmetrics.Metric.state_dict` on all processes, otherwise this can cause a deadlock.
This can also happen when the ``state_dict`` of your :class:`torch.nn.Module` is generated if metric attributes have been registered.
```

docs/source/references/modules.rst
```diff
@@ -12,6 +12,47 @@ metrics.
...
+    ...         super().__init__(dist_sync_state_dict=False)
+    ...         self.add_state("x", torch.tensor(0.0), dist_reduce_fx=torch.sum, persistent=True)
+    ...         self.add_state("total", torch.tensor(0.0), dist_reduce_fx=torch.sum, persistent=True)
+    ...
+    ...     def update(self, x: torch.Tensor):
+    ...         self.x += x
+    ...         self.total += 1
+    ...
+    ...     def compute(self):
+    ...         return self.x / self.total
```
carmocca, line 40: Some whitespace

```suggestion
    ...         return self.x / self.total
    ...
```

docs/source/references/modules.rst
```diff
@@ -12,6 +12,47 @@ metrics.
...
 
+.. note::
+
+    Metric states are being synchornized by default when ``persistent=True`` in a distributed setting
+    when using :meth:`~torchmetrics.Metric.state_dict`
+    This behaviour can be blocked in the following ways.
+    We strongly recommend calling :meth:`~torchmetrics.Metric.state_dict` on all processes or this can cause a deadlock.
+
+
+We recommend to use ``dist_sync_state_dict`` when your metric is nested inside another :class:`~torch.nn.Module`.
```
carmocca, line 23:
```suggestion
The synchronization can be avoided in the following ways.

Set ``dist_sync_state_dict`` when your metric is nested inside another :class:`torch.nn.Module`.
```

docs/source/references/modules.rst
```diff
@@ -12,6 +12,47 @@ metrics.
...
+    ...         self.x += x
+    ...         self.total += 1
+    ...
+    ...     def compute(self):
+    ...         return self.x / self.total
+    >>> metric = AverageMeter()
+    >>> metric.state_dict()
+    OrderedDict(...)
+
+We recommend to use ``state_dict(should_sync=False)`` if you can easily access your :class:`~torchmetrics.Metric`.
```
carmocca, line 45:
```suggestion
Or use ``state_dict(should_sync=False)`` if you can easily access your :class:`~torchmetrics.Metric`.
```

torchmetrics/metric.py
```diff
@@ -415,25 +417,68 @@ def state_dict(
...
+
+        # whether the metric should be synced.
+        should_sync = should_sync if should_sync is not None else self.dist_sync_state_dict
+        is_persistent = False
+
+        with self.sync_context(dist_sync_fn=self.dist_sync_fn, should_sync=should_sync):
+            device = None
             for key in self._defaults:
+
+                # used to track if any defaults is persistent
```
carmocca, line 433:
```suggestion
                # used to track if any defaults are persistent
```

torchmetrics/metric.py
```diff
@@ -415,25 +417,68 @@ def state_dict(
...
+                        for v in current_val:
+                            if device is None:
+                                device = v.device
+
                 # the tensors will be synced across processes so deepcopy to drop the references
                 destination[prefix + key] = deepcopy(current_val)  # type: ignore
+
+            if is_persistent:
+                destination[prefix + "has_synced"] = torch.tensor(should_sync, device=device)  # type: ignore
+                destination[prefix + "rank"] = torch.tensor(self.current_rank, device=device)  # type: ignore
```
carmocca, line 460: Why do we need to save them and why as tensors?

torchmetrics/metric.py
```diff
@@ -447,13 +492,28 @@ def _load_from_state_dict(
...
 
-        # only global rank 0 should be reloading the values present in the ``state_dict``
-        # as the state contains synced values across all progress_group
+        # assumption: users didn't manually sync the Metric states by default.
+        has_synced = state_dict.pop(prefix + "has_synced", None)
+        previous_rank = state_dict.pop(prefix + "rank", 0)
+        previous_world_size = state_dict.pop(prefix + "world_size", self.world_size)
+
+        if has_synced is False:
+            if previous_world_size != self.world_size:
```
carmocca, line 501: These `if`s could be joined together

docs/source/references/modules.rst
tchaton: Done !

torchmetrics/metric.py
tchaton: For DDP. If the LightningModule is wrapped into a DDP wrapped, parameters will be broadcasted for check.
Therefore they need to be tensors.
I plan to add a way to bypass this, but this need to be added to Lightning first.

torchmetrics/metric.py
tchaton: Done !

tchaton:
Closing this pull request as the design choices are wrong.

Issue #372, opened 2021-07-14
SMAPE regression error


hugoperrin:
## 🚀 Feature
Add SMAPE regression error to the regression metrics

### Motivation

There was the presence of MAPE but not SMAPE. As a computer vision research engineer, I'm more used to look at SMAPE instead of MAPE because SMAPE is bounded between 0 and 1 naturally, which makes it easier to interpret and compare.

### Pitch

I implemented the metric in a fork as it is a relatively easy metric to add in, so I figured it would be a quick win to add it in :).
It's actually quite straightforward to implement as it is quite similar to MAPE/MSE and so on!

### Additional context

I have the code ready for a PR with tests already as the code needed is quite light.

hugoperrin:
All tests and docs are running on the code I have, so it should not be a long task from now on. Let me know if I am supposed to create the PR now, or if I need for someone to approve the relevancy of the feature before before creating it :)

SkafteNicki:
Hi @hugoperrin, please feel free to send us your PR, sounds like a great enhancement :]

hugoperrin:
@SkafteNicki Perfect :) The PR has been created, I will have to change a few details regarding the automatic CI, small detail with flake8, will let you know there when it's done (tonight normally :) )

PR #375, opened 2021-07-15
SMAP regression error

hugoperrin:
# Before submitting

## What does this PR do?
This PR adds SMAPE regression metric: https://github.com/PyTorchLightning/metrics/issues/372.

## PR review

## Did you have fun?

### DIFF
--- a/tests/helpers/non_sklearn_metrics.py
+++ b/tests/helpers/non_sklearn_metrics.py
@@ -12,7 +12,7 @@ def symmetric_mean_absolute_percentage_error(
     sample_weight: Optional[np.ndarray] = None,
     multioutput: str = 'uniform_average'
 ):
-    """Symmetric mean absolute percentage error regression loss.
+    r"""Symmetric mean absolute percentage error regression loss.
     <https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error>`_ (SMAPE):
 
     .. math:: \text{SMAPE} = \frac{2}{n}\sum_1^n\frac{max(|   y_i - \hat{y_i} |}{| y_i | + | \hat{y_i} |, \epsilon)}
@@ -48,7 +48,7 @@ def symmetric_mean_absolute_percentage_error(
         Note that we return a large value instead of `inf` when y_true is zero.
 
     """
-    y_type, y_true, y_pred, multioutput = _check_reg_targets(y_true, y_pred, multioutput)
+    _, y_true, y_pred, multioutput = _check_reg_targets(y_true, y_pred, multioutput)
     check_consistent_length(y_true, y_pred, sample_weight)
     epsilon = np.finfo(np.float64).eps
     smape = 2 * np.abs(y_pred - y_true) / np.maximum(np.abs(y_true) + np.abs(y_pred), epsilon)
@@ -56,7 +56,7 @@ def symmetric_mean_absolute_percentage_error(
     if isinstance(multioutput, str):
         if multioutput == 'raw_values':
             return output_errors
-        elif multioutput == 'uniform_average':
+        else:
             # pass None as weights to np.average: uniform mean
             multioutput = None
 
--- a/tests/helpers/non_sklearn_metrics.py
+++ b/tests/helpers/non_sklearn_metrics.py
@@ -1,13 +1,11 @@
 """File for non sklearn metrics that are to be used for reference for tests"""
 
 import numpy as np
-from sklearn.utils.validation import check_consistent_length
 from sklearn.metrics._regression import _check_reg_targets
+from sklearn.utils.validation import check_consistent_length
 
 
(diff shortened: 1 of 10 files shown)


tests/helpers/non_sklearn_metrics.py
```diff
@@ -0,0 +1,52 @@
+"""File for non sklearn metrics that are to be used for reference for tests"""
+
+import numpy as np
+from sklearn.metrics._regression import _check_reg_targets
+from sklearn.utils.validation import check_consistent_length
+
+
+def symmetric_mean_absolute_percentage_error(y_true, y_pred, sample_weight=None, multioutput='uniform_average'):
+    """Symmetric mean absolute percentage error regression loss.
```
SkafteNicki, line 9: could you insert a reference here just to be sure which definition you used, if we need to debug in the future

tests/helpers/non_sklearn_metrics.py
```diff
@@ -0,0 +1,52 @@
+"""File for non sklearn metrics that are to be used for reference for tests"""
+
+import numpy as np
+from sklearn.metrics._regression import _check_reg_targets
+from sklearn.utils.validation import check_consistent_length
+
+
+def symmetric_mean_absolute_percentage_error(y_true, y_pred, sample_weight=None, multioutput='uniform_average'):
```
SkafteNicki, line 8:
```suggestion
def symmetric_mean_absolute_percentage_error(y_true: np.ndarray, y_pred: np.ndarray, sample_weight: Optional[np.ndarray]=None, multioutput: str='uniform_average'):
```

tests/helpers/non_sklearn_metrics.py
```diff
@@ -0,0 +1,52 @@
+"""File for non sklearn metrics that are to be used for reference for tests"""
+
```
SkafteNicki, line 2: for the function typing above
```suggestion
from typing import Optional
```

torchmetrics/functional/regression/symmetric_mean_absolute_percentage_error.py
```diff
@@ -0,0 +1,76 @@
...
+    return sum_abs_per_error, num_obs
+
+
+def _symmetric_mean_absolute_percentage_error_compute(sum_abs_per_error: Tensor, num_obs: int) -> Tensor:
+    return sum_abs_per_error / num_obs
+
+
+def symmetric_mean_absolute_percentage_error(preds: Tensor, target: Tensor) -> Tensor:
+    """
+    Computes symmetric mean absolute percentage error.
```
SkafteNicki, line 46: insert link to wiki page (https://en.wikipedia.org/wiki/Symmetric_mean_absolute_percentage_error) + formula

torchmetrics/functional/regression/symmetric_mean_absolute_percentage_error.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Args:
+        preds: estimated labels
+        target: ground truth labels
+
+    Return:
+        Tensor with SMAPE
+
+    Note:
```
SkafteNicki, line 55: this note does not really make sense to the user without knowing how the epsilon influences the computation. I suggest removing it

SkafteNicki (commented):
Looks great. Two things:
* Remember to add an note in the `Changelog` file
* The function that is compared against takes a `multioutput` argument but the functional/modular implementation of SMAPE does not. Why is that?

Borda (approved):

Borda:
@hugoperrin mind check last comments from @SkafteNicki, mainly related to references

tests/helpers/non_sklearn_metrics.py
hugoperrin: I've added the wiki page and the formula for reference

torchmetrics/functional/regression/symmetric_mean_absolute_percentage_error.py
hugoperrin: I added them :)

torchmetrics/functional/regression/symmetric_mean_absolute_percentage_error.py
hugoperrin: I removed it indeed. I originally kept it, in order to explain the default value used for epsilon, but it's true it does not make much sense since I'm not comparing it to a sklearn implementation

hugoperrin:
@Borda I updated the needed docs, I think this should be good, I'll wait on the CI to finish and fix things if things are needed to be fixed then :) Thank you for your quick feedback :)

Borda:
> PS: should I squash the commits by the way?

no need, well do it with all PR by default: squash + merge

SkafteNicki (approved):
LGTM

PR #378, opened 2021-07-16
fix nDCG can not be called with negative relevance targets

### LINKED ISSUE #377: Functional nDCG can not be called with negative relevance targets
## 🐛 Bug

NDCG does not work with non-binary values because ```_check_retrieval_functional_inputs``` has a wrong check for binary array contents.

### To Reproduce

Call ```torchmetrics.functional.retrieval_normalized_dcg``` with negative relevance targets.

```
torchmetrics_bugreport_ndcg.py", line 6, in <module>
    ndcg_torchmetrics = retrieval_normalized_dcg(score_array, qrel_array, k=5)
torchmetrics/functional/retrieval/ndcg.py", line 49, in retrieval_normalized_dcg
    preds, target = _check_retrieval_functional_inputs(preds, target, allow_non_binary_target=True)
torchmetrics/utilities/checks.py", line 544, in _check_retrieval_functional_inputs
    raise ValueError("`target` must contain `binary` values")
ValueError: `target` must contain `binary` values

```
#### Code sample
```
import torch
from torchmetrics.functional import retrieval_normalized_dcg

qrel_array = torch.tensor([[0, 1, -1, 0], [0, 2, 0, 1], [0, 1, 0, 2]])
score_array = torch.tensor([[0, 0.5, 0, 0.5], [0, 0.5, 0, 0.5], [0, 0.5, 0, 0.5]])
ndcg_torchmetrics = retrieval_normalized_dcg(score_array, qrel_array, k=5)

```
Should work but expects the targets to be in between 0 and 1 and thus throws an exception.

### Environment

 - PyTorch Version: 1.9.0
 - OS (e.g., Linux): Manjaro
 - How you installed PyTorch (`conda`, `pip`, source): conda
 - Build command you used (if compiling from source): None
 - Python version: 3.9
 - CUDA/cuDNN version: 11.4
 - GPU models and configuration: -
 - Any other relevant information: -

paul-grundmann:
# Before submitting
- [x] Did you make sure to update the docs? (No update needed)

## What does this PR do?
Fixes #377

## PR review

## Did you have fun?
Of course I had fun fixing this issue :)


Borda (approved):
can we also add a test to cover this behaviour as nothing was failing till now...

SkafteNicki:
@paul-grundmann could you try changing the `target` line to `target=torch.randint(low=-2, high=4...` to also check for negative values?
https://github.com/PyTorchLightning/metrics/blob/a4d2a7a81ba482f28edd30bd76abc3cffbaca548/tests/retrieval/inputs.py#L35-L39

Borda:
@paul-grundmann mind check the failing tests and https://github.com/PyTorchLightning/metrics/pull/378#issuecomment-881439892

paul-grundmann:
It seems that the random generation of inputs can lead to invalid inputs for the calculation of the DCG (e.g. a single target with 0 relevance). This leads to a division by zero:
```
return _dcg(sorted_target) / _dcg(ideal_target) # _dcg(ideal_target) == 0
```

Should I add a test in the nDCG calculation if the ideal DCG is zero and then return 0.0 as a result?

SkafteNicki:
It seems to me that maybe the ndcg should be changed to account for division by zero. Looking at sklearns implementation:
https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/metrics/_ranking.py#L1458-L1466
they set it to 0 we the denominator is 0. @lucadiliello any opinion here?

lucadiliello:
> It seems that the random generation of inputs can lead to invalid inputs for the calculation of the DCG (e.g. a single target with 0 relevance). This leads to a division by zero:
> 
> ```
> return _dcg(sorted_target) / _dcg(ideal_target) # _dcg(ideal_target) == 0
> ```
> 
> Should I add a test in the nDCG calculation if the ideal DCG is zero and then return 0.0 as a result?

Yes, I think the functional implementation should behave like the `sklearn` version, so it should return `0`. However, if now you are allowing also negative targets, I think [this](https://github.com/paul-grundmann/metrics/blob/5a446c5a8604213ba6b3581b312088e0a52c80e8/torchmetrics/functional/retrieval/ndcg.py#L58) line should be changed.

paul-grundmann:
Ok I had some time to play around since the recent code change introduced a lot more failing tests.
On the one hand it seems that scikit learn also does not handle the edge case of the DCG being zero Typically this happens for ```k=1 and a relevance target of 0 at idx 0```. This of course produces failing tests. 

On the other hand the current architecture of the tests. In the ```helpers.py``` script there are tests for the errors we test against. In case of the nDCG the ```_errors_test_functional_metric_parameters_default``` contains a check for binary values. So I added a new parameter definition which is in fact mostly identical to the default_parameters but without the check for binary targets.
I am not so fluent with pyTest so I don't know whether there might be a more elegant solution for that.

I think the first problem is more drastically because the tests can actually fail randomly. With relevance targets of only -1 instead of -2 I had a lot of successful test runs but one or two failing at some point. Maybe the inputs need to be generated differently or at least with a check if the targets with k=1 contain at least one relevant sample

Borda (commented):
LGTM :rabbit:

SkafteNicki (approved):

Issue #455, opened 2021-08-17
GPU support in BootStrapper


frcnt:
## 🐛 Bug

BootStrapper yields a device error when the predicted and target tensors, along with the metric are on GPU. 

Regardless of the device on which the tensors and the metric are hosted, this [call to `_bootstrap_sampler`](https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/wrappers/bootstrapping.py#L152)` returns indices hosted on the CPU. 
The device error is later thrown when calling `torch.index_select`.

### To Reproduce

1. Run here-below code sample

Stack traces produced

>Traceback (most recent call last):
  File "/home/fco/sentiment-analysis/sentiment_analysis/training/metrics.py", line 110, in <module>
    mae.update(preds, targets)
  File "/home/fco/anaconda3/envs/sentiment/lib/python3.8/site-packages/torchmetrics/metric.py", line 248, in wrapped_func
    return update(*args, **kwargs)
  File "/home/fco/anaconda3/envs/sentiment/lib/python3.8/site-packages/torchmetrics/wrappers/bootstrapping.py", line 153, in update
    new_args = apply_to_collection(args, Tensor, torch.index_select, dim=0, index=sample_idx)
  File "/home/fco/anaconda3/envs/sentiment/lib/python3.8/site-packages/torchmetrics/utilities/data.py", line 197, in apply_to_collection
    return elem_type([apply_to_collection(d, dtype, function, *args, **kwargs) for d in data])
  File "/home/fco/anaconda3/envs/sentiment/lib/python3.8/site-packages/torchmetrics/utilities/data.py", line 197, in <listcomp>
    return elem_type([apply_to_collection(d, dtype, function, *args, **kwargs) for d in data])
  File "/home/fco/anaconda3/envs/sentiment/lib/python3.8/site-packages/torchmetrics/utilities/data.py", line 187, in apply_to_collection
    return function(data, *args, **kwargs)
RuntimeError: Input, output and indices must be on the current device

#### Code sample

```
import torchmetrics as tm

device = "cuda:0"
mae = tm.BootStrapper(tm.MeanAbsoluteError()).to(device)
preds = torch.tensor([-1.5, -.5, .5, 1.5]).to(device)
targets = torch.tensor([-2.0, -1.0, 1.0, 2.0]).to(device)
mae.update(preds, targets)
v = mae.compute()
```

### Expected behavior
The same behaviour as when executing the above piece of code on the CPU.
```
[...]

PR #462, opened 2021-08-18
Device and dtype properties

SkafteNicki:
# Before submitting

## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/455
Issue describes how the `Bootstrapper` metric does not work currently on gpu. Trying to fix this made me realize that we do not have a easy way of getting the device and dtype of a metric. This PR implements the logic from the `DeviceDtypeMixin` class taken from lightning into the core `Metric` class.
https://github.com/PyTorchLightning/pytorch-lightning/blob/38ceb8943ef9b858abead1fbba43ea9a9b4cd93b/pytorch_lightning/core/mixins/device_dtype_mixin.py

Additional, solve the issue using the new property :]

## PR review

## Did you have fun?


torchmetrics/metric.py
```diff
@@ -411,6 +413,101 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+    def half(self) -> "Metric":
+        """Casts all floating point parameters and buffers to ``half`` datatype.
+
+        Returns:
+            Module: self
+        """
+        self._update_properties(dtype=torch.half)
+        return super().half()
+
+    def _update_properties(
```
justusschock, line 502: Just to make sure, this is also applied to the states by the overwritten `_apply` below, right?

Borda (approved):

torchmetrics/metric.py
SkafteNicki: yes, since every call to `super().half()/cpu() ect` looks something like this:
```python
def half(self: T) -> T:
        r"""Casts all floating point parameters and buffers to ``half`` datatype.

        .. note::
            This method modifies the module in-place.

        Returns:
            Module: self
        """
        return self._apply(lambda t: t.half() if t.is_floating_point() else t)
```

torchmetrics/metric.py
```diff
@@ -411,6 +413,100 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+    def dtype(self):
+        return self._dtype
+
+    @dtype.setter
+    def dtype(self, new_dtype) -> None:
+        # necessary to avoid infinite recursion
+        raise RuntimeError("Cannot set the dtype explicitly. Please use module.to(new_dtype).")
+
+    @property
+    def device(self):
```
Borda, line 426: type?

torchmetrics/metric.py
```diff
@@ -411,6 +413,100 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+        if device is None or isinstance(device, int):
+            device = torch.device("cuda", index=device)
+        self._update_properties(device=device)
+        return super().cuda(device=device)
+
+    def cpu(self) -> "Metric":
+        """Moves all model parameters and buffers to the CPU.
+
+        Returns:
+            Module: self
```
Borda, line 457: I feel all these return docstrings with `self` only are quite useless :rabbit:

SkafteNicki: agree, just copy pasted them initially. Have removed them :]

ananthsub (commented):
does this mean metrics used in a lightning module where mixed precision training is used would be converted to use fp16 precision as well? is that always desirable? do people want to compute metrics in fp32 while doing the rest of model training in fp16?

SkafteNicki:
@ananthsub the PR does not actually introduce that kind of change, it has always been the case in TM that if you cast your metric to fp16 the metric states would also be casted (since we have overridden the `self._apply` method):
https://github.com/PyTorchLightning/metrics/blob/689b2189c6f2aff3968d94d8e5fcfdb85dc5b98a/torchmetrics/metric.py#L414-L441
This PR just makes sure that when `half(), cpu(), cuda(), to(...)` is called we have some local properties which tracks this.

If people want fp32 metrics when doing mixed training, I am not sure about. I am not sure that it matters for that many to have the extra precision during training. However, when it comes to testing, it is very clear for me that users should be using fp32.

CHANGELOG.md
```diff
@@ -12,9 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added Learned Perceptual Image Patch Similarity (LPIPS) ([#431](https://github.com/PyTorchLightning/metrics/issues/431))
 
+
 - Added support for float targets in `nDCG` metric ([#437](https://github.com/PyTorchLightning/metrics/pull/437))
 
 
+- Added `device` and `dtype` properties ([462](https://github.com/PyTorchLightning/metrics/pull/462))
```
SkafteNicki, line 19:
```suggestion
- Added `device` and `dtype` properties ([#462](https://github.com/PyTorchLightning/metrics/pull/462))
```

CHANGELOG.md
```diff
@@ -36,6 +40,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed bug where compositional metrics where unable to sync because of type mismatch ([#454](https://github.com/PyTorchLightning/metrics/pull/454))
 
+
+- Fixed `BootStrapper` metrics not working on GPU ([462](https://github.com/PyTorchLightning/metrics/pull/462))
```
SkafteNicki, line 44:
```suggestion
- Fixed `BootStrapper` metrics not working on GPU ([#462](https://github.com/PyTorchLightning/metrics/pull/462))
```

torchmetrics/metric.py
```diff
@@ -411,6 +413,80 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+
+    def to(self, *args: Any, **kwargs: Any) -> "Metric":
+        """Moves and/or casts the parameters and buffers.
+
+        Works similar to nn.Module.to but also updates the metrics device and dtype properties
+        """
+        # there is diff nb vars in PT 1.5
+        out = torch._C._nn._parse_to(*args, **kwargs)
+        self._update_properties(device=out[0], dtype=out[1])
+        return super().to(*args, **kwargs)
```
maximsch2, line 439: will this be enough to move `self._defaults` as well? My guess is not?

torchmetrics/metric.py
```diff
@@ -411,6 +413,80 @@ def __setstate__(self, state: Dict[str, Any]) -> None:
...
+        """Casts all floating point parameters and buffers to ``double`` datatype."""
+        self._update_properties(dtype=torch.double)
+        return super().double()
+
+    def half(self) -> "Metric":
+        """Casts all floating point parameters and buffers to ``half`` datatype."""
+        self._update_properties(dtype=torch.half)
+        return super().half()
+
+    def _update_properties(
```
maximsch2, line 481: Given that we mostly just update one thing, will it be easier to just directly write `self._dtype=dtype` in those functions instead of calling this?

torchmetrics/text/bert.py
```diff
@@ -97,7 +97,7 @@ def __init__(
         self.all_layers = all_layers
         self.num_threads = num_threads
         self.batch_size = batch_size
-        self.device = device
+        self.embedding_device = device
```
maximsch2, line 100: This is a bit counter-intutitive now. If I create this metric with `device=...` in params then it will not actually change the `metric.device`

maximsch2:
On mixed-precision case - I agree with Ananth that this is potentially a concern (especially for metrics with accumulations - fp16 will overflow at ~64k so having a 100k sample dataset and doing fp16 training will break even simple metrics like Accuracy), but not new thing introduced by this diff. Let's file an issue and track it? I'm assuming people will get `nan/inf` as a result of metric in case of fp16 overflow so at least we are not going to silently screw them up.

Worth it to have an example of how to do mixed-precision metric calculation in the docs once this question comes up though.

tchaton (approved):
Great !

leezu:
@ananthsub
> does this mean metrics used in a lightning module where mixed precision training is used would be converted to use fp16 precision as well? is that always desirable? do people want to compute metrics in fp32 while doing the rest of model training in fp16?

It's not desirable, but this behavior was already introduced by accident in cda5dbd. I opened https://github.com/PyTorchLightning/metrics/issues/484 for tracking.

SkafteNicki:
@ananthsub, @maximsch2 please see PR #493 that should fix the problems with auto cast. `half`, `double`, `float` is getting disabled for now and will not change the dtype of the metric states, which should hopefully fix the problems with mixed precision training.

Issue #459, opened 2021-08-18
Don't import all modules from torchmetrics root


ananthsub:
## 🚀 Feature

@SkafteNicki could we not import all metrics at the root module? https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/__init__.py

### Motivation

Reasons against this:
- This slows down import times
- Not all requirements might be installed. For instance, [torch-fidelity is required for FID under torchmetrics.image](https://github.com/PyTorchLightning/metrics/blob/00f0256db37406df09c18ddc18f4eca99ef7d59e/torchmetrics/image/fid.py#L26-L27). But torchmetrics itself doesn't require these dependencies to be installed. 

### Pitch
Keep metric implementation imports at the submodule level, and ensure the root torchmetrics is used for only the essential classes (e.g. `Metric`)

Borda:
> * This slows down import times

do you have any benchmark, how much slower it is?

> * Not all requirements might be installed.

this can be easily solved by conditional import... :rabbit:

SkafteNicki:
Small benchmark executing the statement "import torchmetrics as tm" 10 times
On master: 1.8266461610794067s+-0.033780458858975175 s
PR #463: 1.5197294712066651s+-0.07895099137983734 s

ananthsub:
@Borda it's even more pronounced when using python with tons of shared libraries through a PAR or XAR file

awaelchli:
For comparison: sklearn has also all metrics importat at the top level at sklearn.metrics. They also rely on numpy like we do on pytorch. 
On the other hand, as the inventory of metrics grow, it is possible that there can be collisions in different domains where metrics have the same name but are defined very differently.

carmocca:
I think this is worth it considering the vast majority of users import 1-3 metrics at most. Loading the entire package for this is overkill

tchaton:
I would align with Ananth there. 

As the project grows, I believe there is need to provide better organizations and namespace, especially it saves loading time.

A benefit from this approach might be to incentivise users to check out other namespace and discover new metrics.

Best,
T.C

Borda:
> As the project grows, I believe there is need to provide better organizations and namespace, especially it saves loading time.

I don't see any speed degradation, @ananthsub can you provide some benchmark?
@tchaton this is breaking chnage with any back compatibility for all users so this shall be well justified =)

akihironitta:
> still don't see why top level imports could be problem if there is only torch dependency....

[...]

PR #463, opened 2021-08-18
Change import pattern in TM

SkafteNicki:
# Before submitting

## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/459
Changes the import pattern from
```python
from torchmetrics import Accuracy
```
to
```python
from torchmetrics.classification import Accuracy
```

## PR review

## Did you have fun?


torchmetrics/__init__.py
```diff
@@ -11,117 +11,31 @@
...
+from torchmetrics import (
+    audio,
+    classification,
+    functional,
+    image,
+    regression,
+    retrieval,
+    text,
+    wrappers
+)  # noqa: E402
```
Borda, line 27: this is redundant, it is already the existing package structure

ananthsub (approved):
Thanks @SkafteNicki !

justusschock (approved):

Borda (changes requested):
pls provide justification of the speed-up
this is a breaking change without any back compatibility!

torchmetrics/__init__.py
```diff
@@ -62,7 +62,7 @@
     RetrievalPrecision,
     RetrievalRecall,
 )
-from torchmetrics.text import WER, BERTScore, BLEUScore, ROUGEScore  # noqa: E402
+from torchmetrics.text import WER, BLEUScore  # noqa: E402
```
Borda, line 65: I think all text metrics are free now...

SkafteNicki: Is the change then necessary anymore?

Borda: I think that we would exclude only the Image metrics which has any 3rd party dependency...

Borda:
@SkafteNicki so the conclusion here is just to remove import Metrics that have a dependency on 3rd party integration and include only PyTorch native (our own) implementations...

torchmetrics/detection/map.py
```diff
@@ -190,6 +190,42 @@ class MAP(Metric):
             Callback that performs the allgather operation on the metric state. When ``None``, DDP
             will be used to perform the allgather
 
+    Example:
+        >>> from torchmetrics.detection.map import MAP
```
SkafteNicki, line 194: @Borda can you help here?
I added the example to showcase the new import path but this makes some test fail due to not having the correct version of torchvision installed that is nessesary for this metric

torchmetrics/image/__init__.py
```diff
@@ -11,9 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from torchmetrics.image.fid import FID  # noqa: F401
-from torchmetrics.image.inception import IS  # noqa: F401
-from torchmetrics.image.kid import KID  # noqa: F401
-from torchmetrics.image.lpip_similarity import LPIPS  # noqa: F401
```
Borda, line 17: I think this is fine to keep :rabbit:

Borda (approved):

Borda:
@SkafteNicki that is an interesting failer, that shall happen with the original PR not here...
the case is that MAP is asking for some TorchVision version which is not minimal for TM :rabbit:

torchmetrics/detection/map.py
Borda: seems I found some workaround if the output is different, but it does not work in the method raises an exception...
https://stackoverflow.com/questions/42708809/python-doctest-skip-a-test-conditionally

Issue #53, opened 2020-07-08
Feature request: Add Mean Average Precision (mAP) metric


ternaus:
The main metric for object detection tasks is the Mean Average Precision, implemented in PyTorch, and computed on GPU.

It would be nice to add it to the collection of the metrics.

The example implementation using `numpy`:

https://github.com/ternaus/iglovikov_helper_functions/blob/master/iglovikov_helper_functions/metrics/map.py

jspaezp:
@Borda I can give this a go.

But I am wondering what input format would be the most consistent with the rest of the package ... 
Most of the stuff that it implemented in the package takes directly tensors of shape [N,C, ...] ... and what scope would you like for it ... 

So directly ...

1. Should it support only pre-aligned inputs?
    -  `(preds =  [N,C] , targets =  [N,C] )`
2. Should it support bounding boxes ? if so in what format (xxyy, xyxy, xwyh, xyhw ...)? Should each update be for a single "group" (each image for instance) or should it take a grouping argument?
    - `(preds: float =  [N1,C] , targets: long/bool =  [N2,C] ,  pred_boxes = [x,x,y,y], target_boxes = [x,x,y,y], pred_grouping: long = [N1], target_grouping: long = [N2], iou_cutoff = 0.5)`;  where the values in the grouping are in the range [0, NUM_GROUPS] so only boxes in the same group are compared with each other. 
3. Should it support masks?
    - Option 1: (preds: float =  [N,C] , targets: long/bool =  [N,C] ,  pred_masks = [N,H,W], target_masks = [N,H,W], grouping = [N], iou_cutoff = 0.5, mask_threshold = 0.5)
    - Option 2: (preds: float = [N1,C,H,W] , targets: long/bool = [N2,C,H,W], pred_grouping: long = [N1], target_grouping: long = [N2], iou_cutoff = 0.5, mask_threshold = 0.5)

On the other hand ... the COCO implementation has for each image a set of positive and negative labels in each image and ignores the rest ... should that be supported?

lmk what you think
Kindest wishes,
Sebastian

lucadiliello:
mAP was added to PL some months ago and is now part of the `torchmetrics` package.

Borda:
added with 60c0eec4954b8a9a44f437d312fd0698a75f3630

SkafteNicki:
reopening as the current implementation in torchmetrics is for information retrieval and here is being asked for an implementation for object detection.

jspaezp:
[...]

PR #467, opened 2021-08-19
Add mean average precision metric for object detection

tkupek:
# Mean Average Precision (mAP) for object detection

New metric for object detection.

## What does this PR do?

This PR introduces the commonly used mean average precision metric for object detection.
As there are multiple different implementations, and even different calculations, the new metric wraps the `pycocotools` evaluation, which is used as a standard for several academic and open-source projects for evaluation.

This metric is actively discussed in issue [https://github.com/PyTorchLightning/metrics/issues/53](https://github.com/PyTorchLightning/metrics/issues/53)

## Note

This is my first contribution to the PyTorchLightning project. Please review the code carefully and give me hints on how to improve and match your guidelines.

### DIFF
--- a/torchmetrics/image/map.py
+++ b/torchmetrics/image/map.py
@@ -14,26 +14,35 @@
 import os
 import sys
 from dataclasses import dataclass
-
 from typing import Any, Optional, List
 
 import torch
-from torch import Tensor
 from pycocotools.coco import COCO
 from pycocotools.cocoeval import COCOeval
+from torch import Tensor
 
 from torchmetrics.metric import Metric
+
+COCO_STATS_MAP_VALUE_INDEX = 0
+COCO_STATS_MAR_VALUE_INDEX = 8
 
 
 @dataclass
 class MAPMetricResults():
+    """
+        Dataclass to wrap the final mAP results
+    """
     map_value: Tensor
     mar_value: Tensor
     map_per_class_value: List[Tensor]
     mar_per_class_value: List[Tensor]
 
 
-class hide_prints:
+class _hide_prints:
+    """
+        Internal helper context to suppress the default output of the pycocotools package
+    """
+
     def __enter__(self):
         self._original_stdout = sys.stdout
         sys.stdout = open(os.devnull, 'w')
@@ -44,6 +53,29 @@ class hide_prints:
 
 
 class MAP(Metric):
+    """
+        Computes the Mean-Average-Precision (mAP) and Mean-Average-Recall (mAR) for object detection predictions.
+        Optionally, the mAP and mAR values can be calculated per class.
+
+        https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Mean_average_precision
+
+        Boxes and targets have to be in COCO format with the box score at the end
+        (x-top left, y-top left, width, height, score)
+
+        Note:
+            This metric is a wrapper for the pycocotools, which is a standard implementation for the mAP metric
+            for object detection.
+            https://github.com/cocodataset/cocoapi/tree/master/PythonAPI/pycocotools
+
+        Note:
+            As the pycocotools library cannot deal with tensors directly, all results have to be transfered
+            to the CPU, this might have an performance impact on your training
+
+        Args:
+            num_classes:
+                Number of classes, required for mAP values per class. default: 0 (deactivate)
+        """
+
     def __init__(
             self,
             num_classes: int = 0,
@@ -68,14 +100,15 @@ class MAP(Metric):
             self.add_state(f'ar_{class_id}', default=torch.tensor(data=[], dtype=torch.float),
                            dist_reduce_fx='mean')
 
-    def update(self, preds, target, class_names):
-        coco_target = COCO()
-        coco_target.dataset = get_coco_target(target=target, class_names=class_names)
+    def update(self, preds: list, target: list):
+        assert (len(preds[0]) == len(target[0]), 'preds and targets need to be of the same length')
 
-        coco_preds = COCO()
-        coco_preds.dataset = get_coco_preds(preds=preds, coco_target=coco_target)
+        coco_target, coco_preds = COCO(), COCO()
 
-        with hide_prints():
+        coco_target.dataset = self.get_coco_format(input=target)
+        coco_preds.dataset = self.get_coco_format(input=preds, is_pred=True)
+
+        with _hide_prints():
             coco_target.createIndex()
             coco_preds.createIndex()
             coco_eval = COCOeval(coco_target, coco_preds, 'bbox')
@@ -86,14 +119,17 @@ class MAP(Metric):
 
         self.average_precision = torch.cat(
             (self.average_precision,
-             torch.tensor([stats[0]], dtype=torch.float, device=self.average_precision.device)))
+             torch.tensor([stats[COCO_STATS_MAP_VALUE_INDEX]], dtype=torch.float,
+                          device=self.average_precision.device)))
 
         self.average_recall = torch.cat(
-            (self.average_recall, torch.tensor([stats[8]], dtype=torch.float, device=self.average_recall.device)))
+            (self.average_recall,
+             torch.tensor([stats[COCO_STATS_MAR_VALUE_INDEX]], dtype=torch.float, device=self.average_recall.device)))
 
+        # if class mode is enabled, evaluate metrics per class
         for class_id in range(self.num_classes):
             coco_eval.params.catIds = [class_id]
-            with hide_prints():
+            with _hide_prints():
                 coco_eval.evaluate()
                 coco_eval.accumulate()
                 coco_eval.summarize()
@@ -101,12 +137,14 @@ class MAP(Metric):
 
             current_value = getattr(self, f'ap_{class_id}')
             class_map_value = torch.cat(
-                (current_value, torch.tensor([stats[0]], dtype=torch.float, device=current_value.device)))
+                (current_value,
+                 torch.tensor([stats[COCO_STATS_MAP_VALUE_INDEX]], dtype=torch.float, device=current_value.device)))
             setattr(self, f'ap_{class_id}', class_map_value)
 
             current_value = getattr(self, f'ar_{class_id}')
             class_map_value = torch.cat(
-                (current_value, torch.tensor([stats[8]], dtype=torch.float, device=current_value.device)))
+                (current_value,
+                 torch.tensor([stats[COCO_STATS_MAR_VALUE_INDEX]], dtype=torch.float, device=current_value.device)))
             setattr(self, f'ar_{class_id}', class_map_value)
 
     def compute(self):
@@ -120,51 +158,30 @@ class MAP(Metric):
                                    mar_per_class_value=mar_per_class_value)
         return metrics
 
+    def get_coco_format(self, input: list, is_pred: bool = False):
+        images = []
+        annotations = []
+        annotation_id = 1  # has to start with 1, otherwise COCOEval results are wrong
 
-def get_coco_target(target, class_names):
-    image_list = []
-    annotation_list = []
-    annotation_id = 1
+        for i, (boxes, labels) in enumerate(zip(input[0], input[1])):  # TODO, what is the default bounding box / label format
+            boxes = boxes.cpu().tolist()
+            labels = labels.cpu().tolist()
+            images.append({'id': i})
+            for box, label in zip(boxes, labels):
+                annotation = {
+                    'id': annotation_id,
+                    'image_id': i,
+                    'category_id': label,
+                    'area': box[2] * box[3],
+                    'iscrowd': 0
+                }
+                if is_pred:
+                    annotation['bbox'] = box[:4]
+                    annotation['score'] = box[-1]
+                else:
+                    annotation['bbox'] = box
+                annotations.append(annotation)
+                annotation_id += 1
 
-    for i, (boxes, labels) in enumerate(zip(target['boxes'], target['labels'])):
-        boxes = boxes.cpu().tolist()
-        labels = labels.cpu().tolist()
-        image_list.append({'id': i})
-        for box, label in zip(boxes, labels):
-            box = box
-            annotation_list.append({
-                'id': annotation_id,
-                'image_id': i,
-                'bbox': box,
-                'category_id': label,
-                'area': box[2] * box[3],
-                'iscrowd': 0
-            })
-            annotation_id += 1
-    classes_list = [{'id': i, 'name': val} for i, val in enumerate(class_names)]
-    return {'images': image_list, 'annotations': annotation_list, 'categories': classes_list}
-
-
-def get_coco_preds(preds, coco_target):
-    image_list = [img for img in coco_target.dataset['images']]
-    annotation_list = []
-    annotation_id = 1
-    for i, (boxes, labels) in enumerate(preds):
-        boxes_scores = boxes.cpu().tolist()
-        boxes = [box[:4] for box in boxes_scores]
-        scores = [box[-1] for box in boxes_scores]
-        labels = labels.cpu().tolist()
-        for box, label, score in zip(boxes, labels, scores):
-            box = box
-            annotation_list.append({
-                'id': annotation_id,
-                'image_id': i,
-                'category_id': label,
-                'bbox': box,
-                'score': score,
-                'area': box[2] * box[3],
-                'iscrowd': 0
-            })
-            annotation_id += 1
-
-    return {'images': image_list, 'annotations': annotation_list, 'categories': coco_target.dataset['categories']}
+        classes = [{'id': i, 'name': str(i)} for i in range(self.num_classes)]
+        return {'images': images, 'annotations': annotations, 'categories': classes}
--- a/torchmetrics/image/map.py
+++ b/torchmetrics/image/map.py
@@ -1,0 +1,170 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import sys
+from dataclasses import dataclass
+
+from typing import Any, Optional, List
+
+import torch
+from torch import Tensor
+from pycocotools.coco import COCO
+from pycocotools.cocoeval import COCOeval
+
+from torchmetrics.metric import Metric
+
+
+@dataclass
+class MAPMetricResults():
+    map_value: Tensor
+    mar_value: Tensor
+    map_per_class_value: List[Tensor]
+    mar_per_class_value: List[Tensor]
+
+
+class hide_prints:
+    def __enter__(self):
+        self._original_stdout = sys.stdout
+        sys.stdout = open(os.devnull, 'w')
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        sys.stdout.close()
+        sys.stdout = self._original_stdout
+
+
+class MAP(Metric):
+    def __init__(
+            self,
+            num_classes: int = 0,
+            compute_on_step: bool = True,
+            dist_sync_on_step: bool = False,
+            process_group: Optional[Any] = None,
+    ) -> None:
+        super().__init__(
+            compute_on_step=compute_on_step,
+            dist_sync_on_step=dist_sync_on_step,
+            process_group=process_group,
+        )
+
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.add_state('average_precision', default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx='mean')
+        self.add_state('average_recall', default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx='mean')
+        self.num_classes = num_classes
+
+        for class_id in range(num_classes):
+            self.add_state(f'ap_{class_id}', default=torch.tensor(data=[], dtype=torch.float),
+                           dist_reduce_fx='mean')
+            self.add_state(f'ar_{class_id}', default=torch.tensor(data=[], dtype=torch.float),
+                           dist_reduce_fx='mean')
+
+    def update(self, preds, target, class_names):
+        coco_target = COCO()
+        coco_target.dataset = get_coco_target(target=target, class_names=class_names)
+
+        coco_preds = COCO()
+        coco_preds.dataset = get_coco_preds(preds=preds, coco_target=coco_target)
+
+        with hide_prints():
+            coco_target.createIndex()
+            coco_preds.createIndex()
+            coco_eval = COCOeval(coco_target, coco_preds, 'bbox')
+            coco_eval.evaluate()
+            coco_eval.accumulate()
+            coco_eval.summarize()
+            stats = coco_eval.stats
+
+        self.average_precision = torch.cat(
+            (self.average_precision,
+             torch.tensor([stats[0]], dtype=torch.float, device=self.average_precision.device)))
+
+        self.average_recall = torch.cat(
+            (self.average_recall, torch.tensor([stats[8]], dtype=torch.float, device=self.average_recall.device)))
+
+        for class_id in range(self.num_classes):
+            coco_eval.params.catIds = [class_id]
+            with hide_prints():
+                coco_eval.evaluate()
+                coco_eval.accumulate()
+                coco_eval.summarize()
+                stats = coco_eval.stats
+
+            current_value = getattr(self, f'ap_{class_id}')
+            class_map_value = torch.cat(
+                (current_value, torch.tensor([stats[0]], dtype=torch.float, device=current_value.device)))
+            setattr(self, f'ap_{class_id}', class_map_value)
+
+            current_value = getattr(self, f'ar_{class_id}')
+            class_map_value = torch.cat(
+                (current_value, torch.tensor([stats[8]], dtype=torch.float, device=current_value.device)))
+            setattr(self, f'ar_{class_id}', class_map_value)
+
+    def compute(self):
+        map_per_class_value = [torch.mean(getattr(self, f'ap_{class_id}')) for class_id in
+                               range(self.num_classes)]
+        mar_per_class_value = [torch.mean(getattr(self, f'ar_{class_id}')) for class_id in
+                               range(self.num_classes)]
+        metrics = MAPMetricResults(map_value=torch.mean(self.average_precision),
+                                   mar_value=torch.mean(self.average_recall),
+                                   map_per_class_value=map_per_class_value,
+                                   mar_per_class_value=mar_per_class_value)
+        return metrics
+
+
+def get_coco_target(target, class_names):
+    image_list = []
+    annotation_list = []
+    annotation_id = 1
+
+    for i, (boxes, labels) in enumerate(zip(target['boxes'], target['labels'])):
+        boxes = boxes.cpu().tolist()
+        labels = labels.cpu().tolist()
+        image_list.append({'id': i})
+        for box, label in zip(boxes, labels):
+            box = box
+            annotation_list.append({
+                'id': annotation_id,
+                'image_id': i,
+                'bbox': box,
+                'category_id': label,
+                'area': box[2] * box[3],
+                'iscrowd': 0
+            })
+            annotation_id += 1
+    classes_list = [{'id': i, 'name': val} for i, val in enumerate(class_names)]
+    return {'images': image_list, 'annotations': annotation_list, 'categories': classes_list}
+
+
+def get_coco_preds(preds, coco_target):
+    image_list = [img for img in coco_target.dataset['images']]
+    annotation_list = []
+    annotation_id = 1
+    for i, (boxes, labels) in enumerate(preds):
+        boxes_scores = boxes.cpu().tolist()
+        boxes = [box[:4] for box in boxes_scores]
+        scores = [box[-1] for box in boxes_scores]
+        labels = labels.cpu().tolist()
+        for box, label, score in zip(boxes, labels, scores):
+            box = box
+            annotation_list.append({
+                'id': annotation_id,
+                'image_id': i,
+                'category_id': label,
+                'bbox': box,
+                'score': score,
+                'area': box[2] * box[3],
+                'iscrowd': 0
+            })
+            annotation_id += 1
+
+    return {'images': image_list, 'annotations': annotation_list, 'categories': coco_target.dataset['categories']}
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 numpy>=1.17.2
 torch>=1.3.1
+pycocotools>=2.0.2
 packaging
--- a/tests/image/test_map.py
+++ b/tests/image/test_map.py
@@ -1,0 +1,101 @@
+# Copyright The PyTorch Lightning team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import namedtuple
+
+import torch
+
+from tests.helpers import seed_all
+from tests.helpers.testers import MetricTester
+from torchmetrics.image.map import MAP
+
+seed_all(42)
+
+Input = namedtuple('Input', ["preds", "target"])
+
+# _input_size = (NUM_BATCHES, BATCH_SIZE, 32, 32)
+# _inputs = [
+#     Input(
+#         preds=torch.randint(n_cls_pred, _input_size, dtype=torch.float),
+#         target=torch.randint(n_cls_target, _input_size, dtype=torch.float),
+#     ) for n_cls_pred, n_cls_target in [(10, 10), (5, 10), (10, 5)]
+# ]
+
+
+class TestMapMetric(MetricTester):
+    def generate_predictions_targets(self, batch_size):
+        box1 = [0.0, 0.0, 1.0, 1.0, 0.8]  # TP_class0
+        box2 = [1.0, 1.0, 2.0, 2.0, 0.9]  # TP_class0
+        box3 = [2.0, 2.0, 3.0, 3.0, 0.7]  # false class (FN for class_0, FP for class_1?)
+        box4 = [3.0, 3.0, 4.0, 4.0, 1.0]  # FN (missing box for class_0)
+        box5 = [4.0, 4.0, 5.0, 5.0, 0.6]  # FP (detection - but no GT for class_0)
+        box6 = [5.0, 5.0, 6.0, 6.0, 1.0]  # TP_class_2 --> to check if we get precision one for class_2
+
+        targets = {
+            'boxes': [torch.FloatTensor([box1[0:4],
+                                         box2[0:4],
+                                         box3[0:4],
+                                         box4[0:4],
+                                         box6[0:4]])] * batch_size,
+            'labels': [torch.LongTensor([0, 0, 1, 0, 2])] * batch_size
+        }
+        predictions = []
+        for i in range(batch_size):
+            predictions.append(((torch.tensor([box1, box2, box3, box5, box6]), torch.tensor([0, 0, 0, 0, 2]))))
+        # predictions = [(torch.tensor([box1, box2, box3, box5, box6]),
+        #                torch.tensor([0, 0, 0, 0, 2]))*batch_size]
+        # How to calculate expected_values:
+        # sorted by score both classes:
+        # box 2 TP
+        # box 1 TP
+        # box 3 FN
+        # box 5 FN
+
+        # sorted by score class_0:
+        # box 2 TP
+        # box 1 TP
+        # box 5 FN
+
+        # sorted by score class_1:
+        # box 3  missing
+
+        # sorted by score class_2:
+        # box 6 ok --> 1 gt/1 prediction --> 100%
+
+        return predictions, targets
+
+    def test_mAP_average(self):
+        numerical_correction = (201 / 202)  # (1-100/101)/2 --> 101 coming from interpolation)
+        expected_value = (1 / 3 * 1 + 1 / 3 * 0 + 1 / 3 * 2 / 3 * numerical_correction)
+
+        expected_value_class0 = (2 / 3) * numerical_correction
+        expected_value_class1 = (0 / 1)
+        expected_value_class2 = (1 / 1)
+
+        for batch_size in [2, 10]:
+            predictions, targets = self.generate_predictions_targets(batch_size)
+
+            mAP = MAP(num_classes=3)
+            mAP.update(predictions, targets, ['class_0', 'class_1', 'class_2'])
+            mAP.update(predictions, targets, ['class_0', 'class_1', 'class_2'])
+            mAP.update(predictions, targets,
+                       ['class_0', 'class_1', 'class_2'])  # dist_reduce_fx='mean'--> should stay the same value
+            metrics = mAP.compute()
+            map_value = metrics.map_value
+            map_values_class = metrics.map_per_class_value
+            # test if reduction in the Metric is set to mean
+            # self.assertAlmostEqual(map_value.item(), expected_value, delta=1e-05)
+            # self.assertAlmostEqual(map_values_class[0].item(), expected_value_class0, delta=1e-05)
+            # self.assertAlmostEqual(map_values_class[1].item(), expected_value_class1, delta=1e-05)
+            # self.assertAlmostEqual(map_values_class[2].item(), expected_value_class2, delta=1e-05)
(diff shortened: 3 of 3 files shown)


torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.add_state("average_precision", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+        self.add_state("average_recall", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+        self.num_classes = num_classes
+
+        for class_id in range(num_classes):
+            self.add_state(f"ap_{class_id}", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+            self.add_state(f"ar_{class_id}", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
```
SkafteNicki, line 100: is it correct that the `dist_reduce_fx` is `mean`? 
In the update method the values gets concatenated from different batches so I would somewhat assume this also need to happen when we have multiple devices.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def update(self, preds: list, target: list):
```
SkafteNicki, line 102: definitely need docstring here describing the expected format of preds and target.
```suggestion
    def update(self, preds: List, target: List) -> None:
```

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Boxes and targets have to be in COCO format with the box score at the end
+    (x-top left, y-top left, width, height, score)
+
+    Note:
+        This metric is a wrapper for the pycocotools, which is a standard implementation for the mAP metric
+        for object detection.
+        https://github.com/cocodataset/cocoapi/tree/master/PythonAPI/pycocotools
+
+    Note:
```
SkafteNicki, line 71:
```suggestion
    .. note::
```

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+        As the pycocotools library cannot deal with tensors directly, all results have to be transfered
+        to the CPU, this might have an performance impact on your training
+
+    Args:
+        num_classes:
+            Number of classes, required for mAP values per class. default: 0 (deactivate)
+    """
```
SkafteNicki, line 78:
```suggestion
        num_classes:
            Number of classes, required for mAP values per class. default: 0 (deactivate)
        compute_on_step:
            Forward only calls ``update()`` and return None if this is set to False. default: True
        dist_sync_on_step:
            Synchronize metric state across processes at each ``forward()``
            before returning the value at the step. default: False
        process_group:
            Specify the process group on which synchronization is called. default: None (which selects the entire world)
            
    Raises:
        RuntimeError:
            If ``pycocotools`` is not installed
        ValueError:
            If ``num_classes`` is not an integer larger or equal to 0
        
    """
```

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+        super().__init__(
+            compute_on_step=compute_on_step,
+            dist_sync_on_step=dist_sync_on_step,
+            process_group=process_group,
+        )
+
+        super().__init__(dist_sync_on_step=dist_sync_on_step)
+        self.add_state("average_precision", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+        self.add_state("average_recall", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+        self.num_classes = num_classes
```
SkafteNicki, line 96:
```suggestion
        if not isinstance(num_classes, int) and num_classes < 0:
            raise ValueError("Expected argument `num_classes` to be a integer larger or equal to 0")
        self.num_classes = num_classes
```

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+
+        for class_id in range(num_classes):
+            self.add_state(f"ap_{class_id}", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+            self.add_state(f"ar_{class_id}", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+
+    def update(self, preds: list, target: list):
+        assert (len(preds[0]) == len(target[0]), "preds and targets need to be of the same length")
```
SkafteNicki, line 103: better to use raise
```suggestion
        if len(preds[0]) != len(target[0]):
            raise ValueError("preds and targets need to be of the same length")
```

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        sys.stdout.close()
+        sys.stdout = self._original_stdout
+
+
+class MAP(Metric):
+    """Computes the Mean-Average-Precision (mAP) and Mean-Average-Recall (mAR) for object detection predictions.
+    Optionally, the mAP and mAR values can be calculated per class.
+
+    https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Mean_average_precision
```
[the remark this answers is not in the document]
tkupek: Not from Wikipedia. There is a medium article that describes is nicely:
https://jonathan-hui.medium.com/map-mean-average-precision-for-object-detection-45c121a31173
What do you think?

torchmetrics/image/map.py
```diff
@@ -0,0 +1,201 @@
...
+    from pycocotools.cocoeval import COCOeval
+else:
+    COCO, COCOeval = None, None
+
+COCO_STATS_MAP_VALUE_INDEX = 0
+COCO_STATS_MAR_VALUE_INDEX = 8
+
+
+@dataclass
+class MAPMetricResults:
```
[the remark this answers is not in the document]
tkupek: Comes from our internal code guideline. The benefit is that you don't need to know the exact keys or index and autocompletion works.

torchmetrics/image/map.py
tkupek: I agree, is there any standard format you are using for object detection? We should go with that.

tkupek: I implemented the tensorflow format, as described in https://github.com/PyTorchLightning/metrics/issues/53

torchmetrics/image/map.py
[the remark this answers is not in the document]
SkafteNicki: Yes lets add that one :]

torchmetrics/image/map.py
tkupek: The average precision is always calculated per sample. So for batches it's the mean over all samples. If we have multiple devices it's therefore the mean of the mean, which should be correct?

torchmetrics/image/map.py
```diff
@@ -0,0 +1,321 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        sys.stdout.close()
+        sys.stdout = self._original_stdout
+
+
+class MAP(Metric):
+    """Computes the Mean-Average-Precision (mAP) and Mean-Average-Recall (mAR) for object detection predictions.
+    Optionally, the mAP and mAR values can be calculated per class.
+
+    https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Mean_average_precision
```
tkupek, line 73:
```suggestion
    https://jonathan-hui.medium.com/map-mean-average-precision-for-object-detection-45c121a31173
```

tkupek:
OK, so if I get it correctly, we should separate the two steps and execute the second part (after matching) in the `compute()` step of the metric?

mibaumgartner:
Hi, IMO there are two options (maybe I missed something):

1) Cache predictions in `update` (in tensor format to ensure that they are transferred correctly in DDP trainings) and compute metrics with wrapped COCO in `compute`.
Pro: easy to implement and fairly save since it uses the wrapped COCO tools.
Contra: might lead to high CPU consumption. I'm not sure how this compared to the original COCO tools evaluation since they load the results/predictions without a cache, too. Instance Segmentation should be cached in a compressed format though (i.e. Run-Length Coding as in COCO tools).

2) Compute matching in `update` and compute metric in `compute`.
Pro: might have a better memory footprint in case of Instance Segmentation.
Contra: Much more complicated since COCO tool functions need to be called externally, take care of syncing in DDP training (since COCO caches in numpy), ... 

Best,
Michael

tkupek:
@mibaumgartner thanks for your hints. I will do some experiments with a training on a large dataset (regarding memory and CPU consumption) to decide which way to go.

Borda:
@tkupek mind resolve conflicts?
@SkafteNicki mind review again?

tkupek:
@mibaumgartner @SkafteNicki 
I did some experiments regarding the issue @mibaumgartner mentioned.

The proposed solution 1 (https://github.com/PyTorchLightning/metrics/pull/467#issuecomment-907594310) has the issue that it's not easy to cache all bounding boxes from multiple samples in a single tensor (because the number of bounding boxes varies, and therefore tensors are of different sizes). IIRC if we cache the objects without a tensor format we get problems with DDP training. Additionally, we can then only accept predictions in `tensor` format and would have to rewrite several parts of the current code.

To check if the current solution is working well enough, I tested how big the difference is between both calculation methods.

Test setup:
- detection model with 3 classes
- test set with `1830` samples and `x` gt boxes
- single GPU with DP mode

In the first run, I tested the current implementation with batch evaluation, the second run caches the predictions and evaluates the full test set at once.
Turned out, the evaluation methods are different by up to 10%. So the current solution is not releasable 😞

mibaumgartner:
Thanks for the update and testing!

I'm not an expert regarding the DDP sync operations (rarely train on multiple GPUs 😅 ) but I think as long as the prediction format is a Tensor it can be wrapped into different containers (i.e. List[Tensor] should work). Only things like np.ndarray, List[np.ndarray], .. (COCO uses np.ndarrays internally) don't work.
So caching the predictions and GT in a List[Tensor] format should be a viable option for solution 1.

I think additional input regarding the DDP sync would be helpful, maybe @SkafteNicki can help?

Best,
Michael

tkupek:
A `List[Tensor]` would work just fine. Are we sure this can work for DDP? Do I need to use the `self.add_state()` method for that?

I can test the final implementation, as I have access to a multi-GPU cluster.

tkupek:
I pushed a draft on the caching solution. However, in DDP mode with 4 GPUs the metric does not seem to get reduced correctly (I get four separate results in my model test step). Need some help here.

Borda:
@SkafteNicki mind have look? :rabbit:

tkupek:
OK, there remains an issue with the syncing of tensors in DDP mode.

I assume this format of the caching state:
```
self.groundtruth_boxes = [
torch.Tensor of size (n,4) for each sample
]
self.groundtruth_classes= [
torch.Tensor of size (n,1) for each sample
]
```
where `n` is the number of detections per sample.

However, in DDP mode with multiple GPUs, it is a list of size `num_gpus` including the results in the format above.
Is this intended? Is there a best practice to merge the items of this list?

Borda:
> OK, there remains an issue with the syncing of tensors in DDP mode.

Do we have a test for this case? the GPU test is using 2 GPUs, is it enough?
Mind also check typing of the added code? (seems to be failing with mypy)

Borda:
> > Do we have a test for this case? the GPU test is using 2 GPUs, is it enough?
> 
> 2 GPUs should be sufficient to test. In general the test quality is not good enough, I will try to extend them this week.

cool, pls ping us when you finish it :]

tkupek:
To test this properly I adapted the demo from `pycocotools`.
I created a subset of [their sample results file](https://github.com/cocodataset/cocoapi/blob/master/results/instances_val2014_fakebbox100_results.json) (first 10 detections) and ran the tool locally to get the results. They are now the basis of the MAP test.

Is there a good place where we can document the json files the test is based on? For reproducibility?

The test for the metric is improved now and checks against pycocotools results.
But: I am currently not able to test with the `run_class_metric_test` method (required for further test parametrization ddp, etc.), as it only supports tensors as input and output. I think this method needs to be adjusted in order to make it work.

@Borda can you help here? I drafted the implementation of the test already in `test_map.py`.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,364 @@
...
+        raise ValueError("Expected all groundtruth_classes in `target` to be of type torch.Tensor")
+
+
+class MAP(Metric):
+    r"""
+    Computes the `Mean-Average-Precision (mAP) and Mean-Average-Recall (mAR)`_ for object detection predictions.
+    Optionally, the mAP and mAR values can be calculated per class.
+
+    Predicted boxes and targets have to be in COCO format with the box score at the end
+    (x-top left, y-top left, width, height, score). See the `update` function for more information
```
senarvi, line 96: It seems that this description is outdated.

tkupek: Thanks for the finding. Fixed it.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,364 @@
...
+        self.add_state("average_recall", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+        if not (isinstance(num_classes, int) and num_classes >= 0):
+            raise ValueError("Expected argument `num_classes` to be a integer larger or equal to 0")
+        self.num_classes = num_classes
+
+        self.add_state("boxes", default=[], dist_reduce_fx="cat")
+        self.add_state("scores", default=[], dist_reduce_fx="cat")
+        self.add_state("labels", default=[], dist_reduce_fx="cat")
+        self.add_state("boxes", default=[], dist_reduce_fx="cat")
+        self.add_state("labels", default=[], dist_reduce_fx="cat")
```
ORippler, line 151: Why don't provide suffixes/prefixes for the preds/targets and track them separately ?

torchmetrics/image/map.py
```diff
@@ -0,0 +1,364 @@
...
+        _input_validator(preds, target)
+
+        for item in preds:
+            self.boxes.append(item["boxes"])
+            self.scores.append(item["scores"])
+            self.labels.append(item["labels"])
+
+        for item in target:
+            self.boxes.append(item["boxes"])
+            self.labels.append(item["labels"])
```
ORippler, line 211: Same here

torchmetrics/image/map.py
```diff
@@ -0,0 +1,364 @@
...
+                scores = self.scores[i]
+                if len(boxes) != len(scores):
+                    raise ValueError(
+                        f"Input boxes and scores of sample {i} have a different"
+                        f" length (expected {len(boxes)} scores, got {len(scores)})"
+                    )
+            else:
+                boxes = self.boxes[i]
+                classes = self.labels[i]
+                scores = None
```
ORippler, line 321: And here.

ORippler (commented):
Why don't we keep track of `preds` and `target` in separate lists ?
Is this done internally by `Metrics` class ? Or is this actually a severe bug.

Asking since I am currently hacking this class for segmentation support.

tkupek:
@ORippler thanks for your hint! This is actually a bug introduced by refactoring.
The tests catched it but I haven't seen it so far.

It's fixed.

Would be happy if you add your segmentation support after we finish this PR.

torchmetrics/image/map.py
```diff
@@ -303,21 +303,21 @@ def _get_coco_format(self, is_pred: bool = False) -> Dict:
         annotations = []
         annotation_id = 1  # has to start with 1, otherwise COCOEval results are wrong
 
-        for i, _ in enumerate(self.boxes):
+        for i, _ in enumerate(self.groundtruth_boxes):
```
ORippler: this assumes that `len(self.groundtruth_boxes) == len(self.detection_boxes)`, which is quite unlikely to happen.

Suggestion:

```
iterable =  self.detection_boxes if is_pred else self.groundtruth_boxes
for i, _ in enumerate(iterable):
```

Alternatively, you could also jointly iterate over the corresponding variables directly by zipping them.

ORippler (changes requested):

torchmetrics/image/map.py
tkupek: IMHO it should always be the same length, because its a `List[List[boxes]]`, first list length is the amount of images, which is checked in `_input_validator (if len(preds) != len(targets))`. 

Each iteration represents the results for a single sample.

I'd prefer the zip, but that does not work with `scores = None` for groundtruth data.

ORippler: Ahh okay, then I misread/misunderstood the API of the metric, thanks for clarifying !

tkupek: No problem, thanks for double checking!

tkupek:
I adapted the `MetricTester` AFAIK.

We should now see if the metric works for DDP, too and where the problems lie.

tkupek:
I added a fix for DDP and tested the metric on

- CPU (`dp`)
- single GPU (`ddp`)
- multi GPU (`ddp`)

The results are _almost equal_. CPU and single GPU are exactly the same. I get a difference of `~0.1%`, in DDP mode which is caused by 1-2 results that are duplicated when collecting over all GPUs. I suppose this is because of an odd number of samples that are distributed to the GPUs equally. Not sure if this is an issue or has to be solved in the metric.

From my point of view, this is ready for review.

tkupek:
For reference: This is the script used to get original `pycocotools` results for the unit tests.

[torchmetrics-map-unittest-pycocotools.zip](https://github.com/PyTorchLightning/metrics/files/7201791/torchmetrics-map-unittest-pycocotools.zip)

torchmetrics/image/map.py
```diff
@@ -0,0 +1,364 @@
...
+        compute_on_step: bool = True,
+        dist_sync_on_step: bool = False,
+        process_group: Optional[Any] = None,
+    ) -> None:
+        super().__init__(
+            compute_on_step=compute_on_step,
+            dist_sync_on_step=dist_sync_on_step,
+            process_group=process_group,
+        )
+
```
SkafteNicki, line 140:
```suggestion
        )
        if not _PYCOCOTOOLS_AVAILABLE:
            raise ValueError("`MAP` metric requires that `pycocotools` installed. Please install"
                                 " with `pip install pycocotools` or `pip install torchmetrics[image]`")

```

SkafteNicki (approved):
LGTM :]

Borda:
@ORippler @senarvi mind update your reviews? :rabbit:

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+        annotation_id = 1  # has to start with 1, otherwise COCOEval results are wrong
+
+        for i, _ in enumerate(self.groundtruth_boxes):
+            images.append({"id": i})
+
+            if is_pred:
+                boxes = self.detection_boxes[i]
+                classes = self.detection_labels[i]
+                scores = self.detection_scores[i]
+                if boxes.size()[0] != scores.size()[0]:
```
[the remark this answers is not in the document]
ORippler: Why are we validating sizes here again explicitply ?

This should be done in the `update` method (at least the docstrings claim this). Though in `input_validator` the length of the individual elements are never checked. Suggestion: Shift this check to the `input_validator` function and leave the docstrings as is

ORippler (changes requested):
The PR is starting to look quite good ! :)

Apart from the listed comments, I have 1 major and 1 minor change (which is more of a suggestion).

Major: Currently, the `MAP` class and related methods live inside the `image` metrics. However, according to the docstrings, image metrics should judge the image quality (e.g. FID, SSID etc.).
https://torchmetrics.readthedocs.io/en/latest/references/modules.html#image-metrics

I therefore propose to shift the MAP metric to a new "object detection" metric type, and clearly separate metrics for image quallity from object detection metrics.


The minor suggestion:

I would suggest to refactor the codes for calculating and setting `average_precision` & `recall` to make use of a general loop structure (aka following the DRY principle).

One could fore example do something like

```
for score, stats_idx in zip(("average_precision", "average_recall"), (COCO_STATS_MAP, COCO_STATS_MAR):

new_state = torch.cat(gettattr(self, score), torch.tensor ...)
setattr()
```

I am also still advocating for the joint, direct iteration over predictions/boxes/scores if possible.
Also, is the current API for `preds, targets` in line with the structures returned & used for training of Object Detection methods ?

SkafteNicki:
IMO this should still live within the `image` module, but the top description should either be removed or changed. It is already wrong now as it contains metrics like `PSNR` which are not really used to judge image quality.

ORippler:
> IMO this should still live within the `image` module, but the top description should either be removed or changed. It is already wrong now as it contains metrics like `PSNR` which are not really used to judge image quality.

Not to be a know-it-all, but PSNR is explicitly used to judge image quality in image compression/coding [source: Wikipedia](https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio) and furthermore heavily used in the Superresolution domain of CV to judge quality of generated (refer TPAMI survey [here](https://arxiv.org/pdf/1902.06068.pdf). Same with SSIM. So therefore all metrics listed in image currently do in fact judge image quality and nothing else, which is the reason for my suggestion ;)

But this decision is not my call to make since I am not project owner and you seem to favor leaving `image quality` and `object detection` metrics combined as a joint `image`metrics :)

justusschock:
@SkafteNicki I actually agree with @ORippler here. On the other hand it is mainly defined on images. The benefit I see in creating a separate subpackage here is the dependencies. Users not doing object detection don't want to install cocotools and vice versa. But since we actually import all Metrics on the subpackage root, this would be a hard requirement (We could change that to checks upon instantiation like in flash, but that's a different discussion).

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+        """Transforms and returns all cached targets or predictions in COCO format.
+
+        Format is defined at https://cocodataset.org/#format-data
+        """
+
+        images = []
+        annotations = []
+        annotation_id = 1  # has to start with 1, otherwise COCOEval results are wrong
+
+        for i, _ in enumerate(self.groundtruth_boxes):
```
senarvi, line 311: Have you considered changing the function so that instead of taking ``is_pred``, it would take the boxes and labels that should be converted:

    def _get_coco_format(self, boxes, labels, scores=None)

It would be called like this:

    coco_target.dataset = self._get_coco_format(self.groundtruth_boxes, self.groundtruth_labels)
    coco_preds.dataset = self._get_coco_format(self.detection_boxes, self.detection_labels, self.detection_scores)

Then this loop could be written in a somewhat more compact form:

    for image_id, (image_boxes, image_labels) in enumerate(zip(boxes, labels)):
        for annotation_id, (box, label) in enumerate(zip(image_boxes, image_labels):
            ...
            if scores is not None:
                annotation["score"] = scores[image_id][annotation_id].cpu()

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+
+class _hide_prints:
+    """Internal helper context to suppress the default output of the pycocotools package."""
+
+    def __init__(self) -> None:
+        self._original_stdout = None
+
+    def __enter__(self) -> None:
+        self._original_stdout = sys.stdout  # type: ignore
+        sys.stdout = open(os.devnull, "w")
```
senarvi, line 53: If we want to have the possibility of seeing the pycocotools messages, maybe we could set sys.stdout to an object of our custom logger class, instead of devnull:

    sys.stdout = WriteToLog()

The class would look something like this:

    log = logging.getLogger(__name__)

    class WriteToLog:
       def write(self, buf):
          for line in buf.rstrip().splitlines():
             log.debug(line.rstrip())

SkafteNicki:
> 
> 
> > IMO this should still live within the `image` module, but the top description should either be removed or changed. It is already wrong now as it contains metrics like `PSNR` which are not really used to judge image quality.
> 
> Not to be a know-it-all, but PSNR is explicitly used to judge image quality in image compression/coding [source: Wikipedia](https://en.wikipedia.org/wiki/Peak_signal-to-noise_ratio) and furthermore heavily used in the Superresolution domain of CV to judge quality of generated (refer TPAMI survey [here](https://arxiv.org/pdf/1902.06068.pdf). Same with SSIM. So therefore, all metrics listed in `image` currently do in fact judge image quality, which is the reason for my suggestion ;)
> 
> But this decision is not my call to make since I am not project owner and you seem to favor leaving `image quality` and `object detection` metrics combined as a joint `image`metrics :)

I actually have no strong opinions on this, and the reasons @justusschock mentions are good reasons for doing the splitting. Maybe we can still in the docs have something like:
```
Image
- Image quality
    - PSNR
    - FID
    - ...
- Object dectection
    - MAP
    - ...
```
just because we have ordered some of the other fields based on their data domain.

senarvi:
I would perhaps place this metric in a separate ``detection`` package, because Torchvision and Lightning Bolts also have a separate ` ``detection`` subpackage for object detection models. The arguments of this metric follow the conventions used in those models and it would be logical that metrics used by those models are in ``detection``. But no strong opinions from me either.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+        self.add_state("detection_scores", default=[])
+        self.add_state("detection_labels", default=[])
+        self.add_state("groundtruth_boxes", default=[])
+        self.add_state("groundtruth_labels", default=[])
+
+        for class_id in range(num_classes):
+            self.add_state(f"ap_{class_id}", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+            self.add_state(f"ar_{class_id}", default=torch.tensor(data=[], dtype=torch.float), dist_reduce_fx="mean")
+
+    def update(self, preds: Any, target: Any) -> None:  # type: ignore
```
senarvi, line 162: The type of preds and target is ``Dict[str, Tensor]``, right?

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+
+COCO_STATS_MAP_VALUE_INDEX = 0
+COCO_STATS_MAR_VALUE_INDEX = 8
+
+
+@dataclass
+class MAPMetricResults:
+    """Dataclass to wrap the final mAP results."""
+
+    map_value: Tensor
```
senarvi, line 39: Should we shorten these to ``map``, ``mar``, ``map_per_class``, ``mar_per_class``?

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:  # type: ignore
+        sys.stdout.close()
+        sys.stdout = self._original_stdout  # type: ignore
+
+
+def _input_validator(preds: List[Dict[str, torch.Tensor]], targets: List[Dict[str, torch.Tensor]]) -> None:
+    """Ensure the correct input format of `preds` and `targets`"""
+
+    if not isinstance(preds, list):
```
senarvi, line 63: I suggest checking that preds and targets are ``Sequence``s, instead of ``list``s, or ``(list, tuple)`` if ``Sequence`` is not available. Requiring that an input argument is a ``list`` can cause unnecessary trouble in Python.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+                    if type(score) != float:
+                        raise ValueError(
+                            f"Invalid input score of sample {i}, element {k}"
+                            f" (expected value of type float, got type {type(score)})"
+                        )
+                    annotation["score"] = score
+                annotations.append(annotation)
+                annotation_id += 1
+
+        classes = [{"id": i, "name": str(i)} for i in range(self.num_classes)]
```
senarvi, line 368: Apparently ``num_classes`` is required for mAP values per class, and if it's set to zero, the code should return only overall MAP and MAR values. However, if the "categories" defined here is empty, pycocotools will issue the warning

    Please run evaluate() first

and all the AP and AR values will be -1. It would be great if the user didn't have to know the number of classes. What if we set the (default) ``num_classes`` automatically:

    torch.cat(tuple(x["labels"] for x in preds + target)).unique()

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+            raise ValueError("Expected argument `num_classes` to be a integer larger or equal to 0")
+        self.num_classes = num_classes
+
+        self.add_state("detection_boxes", default=[])
+        self.add_state("detection_scores", default=[])
+        self.add_state("detection_labels", default=[])
+        self.add_state("groundtruth_boxes", default=[])
+        self.add_state("groundtruth_labels", default=[])
+
+        for class_id in range(num_classes):
```
senarvi, line 158: If I'm not mistaken, a state variable can also be a list. In that case you could create two state variables, ``ap_per_class`` and ``ar_per_class``, with ``num_classes`` entries, and you don't have to use ``getattr()`` / ``setattr()``.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+from torchmetrics.metric import Metric
+from torchmetrics.utilities.imports import _PYCOCOTOOLS_AVAILABLE
+
+if _PYCOCOTOOLS_AVAILABLE:
+    from pycocotools.coco import COCO
+    from pycocotools.cocoeval import COCOeval
+else:
+    COCO, COCOeval = None, None
+
+COCO_STATS_MAP_VALUE_INDEX = 0
```
senarvi, line 31: We're only returning MAP and MAR, but apparently the library computes a whole bunch of other values:

     Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ]
     Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ]
     Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ]
     Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ]
     Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ]
     Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ]
     Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ]
     Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ]
     Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ]
     Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ]
     Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ]
     Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ]

I would be interested in the other values too. Since they are computed anyway, hopefully we could return them too.

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+
+        new_average_precision = torch.cat(
+            (
+                self.average_precision,
+                torch.tensor(
+                    [stats[COCO_STATS_MAP_VALUE_INDEX]], dtype=torch.float, device=self.average_precision.device
+                ),
+            )
+        )
+        setattr(self, "average_precision", new_average_precision)
```
senarvi, line 254: Why not ``self.average_precision = new_average_precision``?

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+            coco_preds.createIndex()
+            coco_eval = COCOeval(coco_target, coco_preds, "bbox")
+            coco_eval.evaluate()
+            coco_eval.accumulate()
+            coco_eval.summarize()
+            stats = coco_eval.stats
+
+        new_average_precision = torch.cat(
+            (
+                self.average_precision,
```
senarvi, line 248: I don't understand the purpose of concatenating the previous AP here. IIUC, ``update()`` is supposed to accumulate the statistics from each batch - in this case the detections and ground truths - and ``compute()`` is supposed to calculate the metric from the statistics that were accumulated from a test epoch. Now it seems that ``compute()`` is calculating the average metric from all previous test epochs.

twsl: Calculating the mean should be done on `x_epoch_end` by using `x_step_outputs`, not inside the metric imo

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+            Specify the process group on which synchronization is called. default: None (which selects the entire world)
+
+    Raises:
+        ValueError:
+            If ``pycocotools`` is not installed
+        ValueError:
+            If ``num_classes`` is not an integer larger or equal to 0
+    """
+
+    def __init__(
```
senarvi, line 128: Can we simplify this:

    def __init__(self, num_classes: int = 0, **kwargs) -> None:
        super().__init__(**kwargs)

torchmetrics/image/map.py
```diff
@@ -0,0 +1,369 @@
...
+            class_map_value = torch.cat(
+                (
+                    current_value,
+                    torch.tensor([stats[COCO_STATS_MAP_VALUE_INDEX]], dtype=torch.float, device=current_value.device),
+                )
+            )
+            setattr(self, f"ap_{class_id}", class_map_value)
+
+            current_value = getattr(self, f"ar_{class_id}")
+            class_map_value = torch.cat(
```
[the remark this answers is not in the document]
tkupek: Thanks. Fixed it!

torchmetrics/image/map.py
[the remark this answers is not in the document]
tkupek: Neat, did not know of `.size(0)`.

I moved the checks to the `input_validator`.

torchmetrics/image/map.py
tkupek: Nice and clean, I like it and refactored the method. Thanks

tkupek: Had to adjust the `annotation_id` a bit, as it has to be unique across all images.

torchmetrics/image/map.py
tkupek: Included this. May I ask you to check the correct implementation please?

torchmetrics/image/map.py
tkupek: Correct. I think the reason for typing this `Any` was of `mypy`. Typed it correctly now.

torchmetrics/image/map.py
tkupek: Agreed, much nicer.

torchmetrics/image/map.py
tkupek: Agreed, that would be much nicer. But what would be the correct 'dist_reduce_fx' for this?
In `DDP` mode we would have to calculate the mean over each value.

torchmetrics/image/map.py
tkupek: Done. I would leave exception and docstring as they are, fine for you?

torchmetrics/image/map.py
tkupek: It does. The reason I only added the two is that I assume they are the most relevant (usually only a single mAP value is given, as in [TF2](https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/tf2_detection_zoo.md)).

Adding the others would require some boilerplate code. Do you think they are relevant or just nine to have?

torchmetrics/image/map.py
tkupek: Well spotted, that is indeed a problem.

See the other thread why I'm unsure about removing the state variables.

What about using your suggested "unique classes workaround" only when `num_classes` is 0? Would that solve the problem?

torchmetrics/image/map.py
tkupek: I had to use the setattr for the `class_map_values` so I thought it would be nicer to have this aligned.

No personal preference. Do you prefer `self.average_precision = new_average_precision` ?

torchmetrics/image/map.py
tkupek: Maybe I have a misinterpretation of the metric functionality here.

I thought metrics are collecting predictions over multiple epochs and return the metric over all the collected predictions on `compute()`. Resetting the metric with `reset()` after each epoch.

If this is not the case, please explain me a bit further how `compute()` should work.

torchmetrics/image/map.py
tkupek: LGTM.

tkupek:
@ORippler @senarvi thank you very much for your feedback.
I went through the suggestions and implemented most of them. Some need further clarification, I added comments.

@SkafteNicki I don't have a preference for the `image` / `detection` package decision and I feel like that is something you should have the last word on. Could you help me move this to the correct place once you made your decision please?

torchmetrics/image/map.py
tkupek: Maybe this problem related to the discussion in https://github.com/PyTorchLightning/metrics/pull/467#discussion_r714596707

torchmetrics/image/map.py
tkupek: OK, now I get it. The state variables are not required and a leftover from a previous implementation.

Removed them.

torchmetrics/image/map.py
tkupek: Resolved as of https://github.com/PyTorchLightning/metrics/pull/467#discussion_r714596707

torchmetrics/image/map.py
tkupek: Resolved as of https://github.com/PyTorchLightning/metrics/pull/467#discussion_r714596707

torchmetrics/image/map.py
tkupek: With the changes from https://github.com/PyTorchLightning/metrics/pull/467#discussion_r714596707 this gets easier now. Done

senarvi: For me the different AP values would be relevant. What makes them a little less usable is that the small/medium/large categories are hardcoded, but I would still be interested in all the values. I guess there's not so much code needed now that we don't have a separate state for every output?

torchmetrics/image/map.py
senarvi: Yes, you could do that. Are you thinking that the "unique classes" value is not always correct, or in which case do you think the user would like to specify the number of classes? It may be that the user doesn't want the class-specific statistics, but is it enough to have a boolean switch?

tests/helpers/testers.py
```diff
@@ -70,6 +71,17 @@ def _assert_allclose(pl_result: Any, sk_result: Any, atol: float = 1e-8, key: Op
         if key is None:
             raise KeyError("Provide Key for Dict based metric results.")
         assert np.allclose(pl_result[key].detach().cpu().numpy(), sk_result, atol=atol, equal_nan=True)
+    elif isinstance(pl_result, MAPMetricResults):
+        for val_id in [a for a in dir(sk_result) if not a.startswith("__")]:
```
tchaton, line 75:
```suggestion
        for val_index in [a for a in dir(sk_result) if not a.startswith("__")]:
```

tkupek:
@SkafteNicki done, please check.

tchaton (approved):
Awesome work !

torchmetrics/image/map.py
tkupek: I cannot think of a case where classes should be used that are not included in `targets`, so I agree, a boolean switch should be enough.

tests/helpers/testers.py
tkupek: Refactored this.

tkupek:
@SkafteNicki the `docs-test` run fails because of:

```
/home/runner/work/metrics/metrics/docs/source/references/modules.rst:356:Title overline too short.

*************
Detection Metrics
*************
```

This makes no sense to me, as `Image Metrics` is similarly short. Please help.

SkafteNicki:
> @SkafteNicki the `docs-test` run fails because of:
> 
> ```
> /home/runner/work/metrics/metrics/docs/source/references/modules.rst:356:Title overline too short.
> 
> *************
> Detection Metrics
> *************
> ```
> 
> This makes no sense to me, as `Image Metrics` is similarly short. Please help.

You need equal amount of "*" as there is text:
******************
Detection Metrics
******************

tkupek:
> 
> You need equal amount of "*" as there is text:
> 

Got it. Thanks 🙂

SkafteNicki (approved):
Still looks good to me :]

torchmetrics/image/map.py
senarvi: Fine.

twsl:
Since the compute method isn't returning a tensor but a custom result wrapper, calling the metric using eg. `self.log('train_map', self.train_map, on_step=True, on_epoch=False)` (see [docs](https://torchmetrics.readthedocs.io/en/latest/pages/lightning.html)) fails with a detach not found error. Also being a custom class, the result cannot be easily logged directly via `self.log` or `self.log_dict`. Documenting the use of `.__dict__` could be a quick partial fix, yet the logger cannot log the map_per_class lists eigher.

tkupek:
@twsl any suggestions on how to deal with this? Should we replace the class with a tensor and document which value is where? Is there a better option?

twsl:
Not really. The bert and rogue metrics return dictionaries, so that could be a solution, but I'm not sure if there are any checks in place to validate whether that works. I'll try to take a closer look at the log methods later.

tkupek:
@twsl thanks alot. Dicts would work easily.

twsl:
As I said in my first reply, maybe the list objects should be a big tensor as it caused problems for me. Adding a test using the logger probably wouldn't hurt

tkupek:
@twsl I changed the outputs from `list` to `Tensor` and the return type to `dict`. Can you please check if this resolves your issue?

tkupek:
@twsl thanks for the catch! Fixed the default and changed the empty return from `None` to an empty tensor.
Can you verify if this works?

twsl:
```The metric `tensor([], device='cuda:0')` does not contain a single element, thus it cannot be converted to a scalar.```
Empty tensor doesnt work, setting -1 or 0 instead of an empty one ir probably the easiest fix.

senarvi:
I'm not so familiar with TorchMetrics, so this might not be related to this pull request, but I wonder why I have to call `reset()` manually after each epoch. This is how I use the metric:

    def __init__(self):
        self.map = MAP(compute_on_step=False)
    
    def validation_step(self, batch, batch_idx):
        x, y = batch
        preds = self(x)
        self.map(preds, y)
    
    def validation_epoch_end(self, outputs):
        self.log_dict(self.map.compute())
        self.map.reset()

If I don't call `reset()` in the end, the predictions will accumulate from all validation epochs. Makes sense, but the [TorchMetrics documentation](https://torchmetrics.readthedocs.io/en/stable/pages/lightning.html) says that "using TorchMetrics with Lightning offers additional benefits: The .reset() method of the metric will automatically be called at the end of an epoch", and gives an example that does more or less what I'm doing.

elstonaug:
Hello, One question - do we need to exclude class 0 which is the background class when updating / handling the metrics?

PR #478, opened 2021-08-24
fixes metric hashing

justusschock:
# Before submitting

## What does this PR do?
We need the hashing to be unique. PyTorch relies on that for children discoverability.

before `hash(AUROC()) == hash(AUROC())` evaluated to true when metrics had an empty non-tensor state (like an empty list) causing some attributes to be not appearing in the children.
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/tests/bases/test_hashing.py
+++ b/tests/bases/test_hashing.py
@@ -1,0 +1,20 @@
+from tests.helpers.testers import DummyMetric, DummyNonTensorMetric
+import pytest
+
+pytest.mark.parametrize(
+    "metric_cls",
+    [
+        (DummyMetric,),
+        (DummyNonTensorMetric,),
+    ],
+)
+def test_metric_hashing(metric_cls):
+    """
+    Tests that hases are different. 
+    See the Metric's hash function for details on why this is required.
+    """
+    instance_1 = metric_cls()
+    instance_2 = metric_cls()
+
+    assert hash(instance_1) != hash(instance_2)
+    assert id(instance_1) != id(instance_2)
--- a/tests/bases/test_hashing.py
+++ b/tests/bases/test_hashing.py
@@ -2,6 +2,7 @@ import pytest
 
 from tests.helpers.testers import DummyListMetric, DummyMetric
 
+
 @pytest.mark.parametrize(
     "metric_cls",
     [
--- a/tests/bases/test_hashing.py
+++ b/tests/bases/test_hashing.py
@@ -1,5 +1,6 @@
-from tests.helpers.testers import DummyMetric, DummyNonTensorMetric
 import pytest
+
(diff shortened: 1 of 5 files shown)


torchmetrics/metric.py
```diff
@@ -502,7 +502,12 @@ def _filter_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
         return filtered_kwargs
 
     def __hash__(self) -> int:
-        hash_vals = [self.__class__.__name__]
+        # we need to add the id here, since PyTorch requires the hash of an module to be unique.
+        # they rely on that for children discovery
+        # (see https://github.com/pytorch/pytorch/blob/v1.9.0/torch/nn/modules/module.py#L1544)
+        # For metrics that include tensors it is not a problem,
+        # since their hash is unique based on the memory location but we cannot rely on that for every metric.
+        hash_vals = [self.__class__.__name__, id(self)]
```
tchaton, line 510:
```suggestion
        # we need to add the id here, since PyTorch requires a module hash to be unique.
        # Internally, PyTorch nn.Module rely on that for children discovery
        # (see https://github.com/pytorch/pytorch/blob/v1.9.0/torch/nn/modules/module.py#L1544)
        # For metrics that include tensors it is not a problem,
        # since their hash is unique based on the memory location but we cannot rely on that for every metric.
        hash_vals = [self.__class__.__name__, id(self)]
```

tchaton (approved):
LGMT !

tests/helpers/testers.py
```diff
@@ -511,6 +511,20 @@ def compute(self):
...
+
+    def __init__(self):
+        super().__init__()
+        self.add_state("x", [], dist_reduce_fx=None)
+
+    def update(self):
+        pass
+
+    def compute(self):
+        pass
```
SkafteNicki, line 525: this is the exact same as the `DummyListMetric` below
```suggestion
```

tests/bases/test_hashing.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from tests.helpers.testers import DummyMetric, DummyNonTensorMetric
```
SkafteNicki, line 3:
```suggestion
from tests.helpers.testers import DummyMetric, DummyListMetric
```

tests/bases/test_hashing.py
```diff
@@ -0,0 +1,23 @@
...
+
+pytest.mark.parametrize(
+    "metric_cls",
+    [
+        (DummyMetric,),
+        (DummyNonTensorMetric,),
```
SkafteNicki, line 9:
```suggestion
        (DummyListMetric,),
```

SkafteNicki (approved):

Borda (approved):

Issue #508, opened 2021-09-08
Implement MultioutputWrapper


an1lam:
### Discussed in https://github.com/PyTorchLightning/metrics/discussions/498

We've been using torchmetrics internally and enjoying it, but two things we've had to work around are:

1. Some metrics like `SpearmanCorrcoef` don't support multioutput mode (unlike `R2Score` for example). 
2. Even some (all?) metrics that do support multioutput mode, don't have support for missing data (`NaN` labels for certain features is our use-case).

As a simple solution to this, we've implemented a `MultioutputWrapper` class (similar to `BootstrapAggregator`) that takes a `base_metric` and copies it `num_outputs` times, updating each underlying on each metric call. While we've tried to make our internal version as generic as possible, we've admittedly only tested it with the metrics we currently use. Examples include `R2Score`, `SpearmanCorrcoef`, and `Accuracy` with (N, C, O) (where O is number of outputs/tasks) dimensional predictions and (N, O) dimensional labels.

Since we've found this wrapper quite helpful, we'd like to contribute it back to the base package so others can use it.

PR #506, opened 2021-09-08
Simple aggregation metrics

SkafteNicki:
# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [ ] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/505
Introduces a couple of simple aggregation metrics for aggregating simple streams of values. 
As `self.log` in lightning do not accept user defined `reduce_fx`, it would be great if torchmetrics can supply some simple metrics for aggregating data. This PR implements `MeanMetric` (updated version of `AverageMeter`), `SumMetric`, `MinMetric`, `MaxMetric` and `CatMetric` that importantly also can deal with `nan` values (what started the issue).

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #510, opened 2021-09-09
Implement MultioutputWrapper

an1lam:
# Before submitting

## What does this PR do?

Implements the `MultioutputWrapper` feature described in #508. Summarizing, `MultioutputWrapper` takes a base metric and a `num_outputs` param and returns a metric that will compute `shape[output_dim]` many scalars and return them in a list. The idea is to support `multioutput` mode for metrics like `SpearmanCorrcoef` that don't support it natively.

### DIFF
--- a/tests/wrappers/test_multioutput.py
+++ b/tests/wrappers/test_multioutput.py
@@ -1,0 +1,27 @@
+import torch
+
+from torchmetrics.wrappers.multioutput import MultioutputWrapper
+from torchmetrics.classification import Accuracy
+from torchmetrics.regression import R2Score
+
+
+def test_multioutput_wrapper():
+    # Multiple outputs, same shapes
+    preds1 = torch.tensor([[1, 2], [3, 4], [5, 6]], dtype=torch.float)
+    target1 = torch.tensor([[1, 4], [3, 2], [5, 6]], dtype=torch.float)
+    preds2 = torch.tensor([[7, 8], [9, 10], [11, 12]], dtype=torch.float)
+    target2 = torch.tensor([[7, 8], [9, 10], [11, 12]], dtype=torch.float)
+
+    r2 = MultioutputWrapper(R2Score(), num_outputs=2)
+    r2.update(preds1, target1)
+    r2.update(preds2, target2)
+
+    # R2 score computed using sklearn's r2_score
+    torch.testing.assert_allclose(r2.compute(), [1, 0.8857])
+
+    # Multiple outputs, different shapes
+    acc = MultioutputWrapper(Accuracy(num_classes=3), num_outputs=2)
+    preds = torch.tensor([[[0.1, 0.3], [0.8, 0.3], [0.1, 0.4]], [[0.8, 0.3], [0.1, 0.4], [0.1, 0.3]]])
+    target = torch.tensor([[1, 2], [1, 1]])
+    acc.update(preds, target)
+    torch.testing.assert_allclose(acc.compute(), [0.5, 1.0])
--- a/tests/wrappers/test_multioutput.py
+++ b/tests/wrappers/test_multioutput.py
@@ -8,7 +8,7 @@ from sklearn.metrics import accuracy_score
 from sklearn.metrics import r2_score as sk_r2score
 
 from tests.helpers import seed_all
-from tests.helpers.testers import BATCH_SIZE, MetricTester, NUM_BATCHES, NUM_CLASSES
+from tests.helpers.testers import BATCH_SIZE, NUM_BATCHES, NUM_CLASSES, MetricTester
 from torchmetrics import Metric
 from torchmetrics.classification import Accuracy
 from torchmetrics.regression import R2Score
@@ -73,7 +73,8 @@ num_targets = 2
 Input = namedtuple("Input", ["preds", "target"])
 
 _multi_target_regression_inputs = Input(
-    preds=torch.rand(NUM_BATCHES, BATCH_SIZE, num_targets), target=torch.rand(NUM_BATCHES, BATCH_SIZE, num_targets),
+    preds=torch.rand(NUM_BATCHES, BATCH_SIZE, num_targets),
+    target=torch.rand(NUM_BATCHES, BATCH_SIZE, num_targets),
 )
 _multi_target_classification_inputs = Input(
     preds=torch.rand(NUM_BATCHES, BATCH_SIZE, NUM_CLASSES, num_targets),
--- a/tests/wrappers/test_multioutput.py
+++ b/tests/wrappers/test_multioutput.py
@@ -54,6 +54,7 @@ class _MultioutputMetric(Metric):
 
     @torch.jit.unused
     def forward(self, *args, **kwargs):
+        """Run forward on the underlying metric."""
         return self.metric(*args, **kwargs)
 
     def reset(self) -> None:
--- a/tests/wrappers/test_multioutput.py
+++ b/tests/wrappers/test_multioutput.py
@@ -1,48 +1,132 @@
+from collections import namedtuple
 from functools import partial
+from typing import Any, Callable, Optional
 
 import pytest
 import torch
 from sklearn.metrics import accuracy_score
+from sklearn.metrics import r2_score as sk_r2score
 
+from tests.helpers import seed_all
+from tests.helpers.testers import BATCH_SIZE, NUM_BATCHES, MetricTester
+from torchmetrics import Metric
 from torchmetrics.classification import Accuracy
 from torchmetrics.regression import R2Score
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_6
 from torchmetrics.wrappers.multioutput import MultioutputWrapper
+
+seed_all(42)
 
 
 def _multioutput_sk_accuracy(preds, target, num_outputs):
     accs = []
+    print("sk:", preds[:4], target[:4])
     for i in range(num_outputs):
         accs.append(accuracy_score(torch.argmax(preds[:, :, i], dim=1), target[:, i]))
     return accs
 
 
+class _MultioutputMetric(Metric):
+    """Multi-output version of the R2 score class for testing."""
+
+    def __init__(
+        self,
+        base_metric_class,
+        num_outputs: int = 1,
+        compute_on_step: bool = True,
+        dist_sync_on_step: bool = False,
+        process_group: Any = None,
+        dist_sync_fn: Optional[Callable] = None,
+        **base_metric_kwargs,
+    ) -> None:
+        super().__init__(
+            compute_on_step=compute_on_step,
+            dist_sync_on_step=dist_sync_on_step,
+            process_group=process_group,
+            dist_sync_fn=dist_sync_fn,
+        )
+        self.metric = MultioutputWrapper(
+            base_metric_class(**base_metric_kwargs),
+            num_outputs=num_outputs,
+            compute_on_step=compute_on_step,
+            dist_sync_on_step=dist_sync_on_step,
+            dist_sync_fn=dist_sync_fn,
+        )
+
+    def update(self, preds: torch.Tensor, target: torch.Tensor) -> None:
+        """Update the each pair of outputs and predictions."""
+        return self.metric.update(preds, target)
+
+    def compute(self) -> torch.Tensor:
+        """Compute the R2 score between each pair of outputs and predictions."""
+        return self.metric.compute()
+
+    @torch.jit.unused
+    def forward(self, *args, **kwargs):
+        return self.metric(*args, **kwargs)
+
+    def reset(self) -> None:
+        """Reset the underlying metric state."""
+        self.metric.reset()
+
+
+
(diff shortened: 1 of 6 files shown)


an1lam:
I'm not sure why the build is failing but it seems to be totally independent of my change & instead has to do with Docker. It looks like it's also failing on #506, so I'm assuming this is not caused by my change?

torchmetrics/wrappers/multioutput.py
```diff
@@ -0,0 +1,112 @@
...
+            represents the length of the batch or dataset being passed in.
+        compute_on_step: bool = True,
+            Whether to recompute the metric value on each update step.
+        dist_sync_on_step: bool = False,
+            Required for distributed training support. See torchmetrics docs for additional details.
+        process_group:
+            Specify the process group on which synchronization is called. default: None (which selects the entire world)
+        dist_sync_fn: Callable = None,
+            Required for distributed training support. See torchmetrics docs for additional details.
+    """
```
SkafteNicki, line 62: some examples in the end would be nice, similar to:
https://github.com/PyTorchLightning/metrics/blob/3a4ff87bdf1f0fdf394b414041e5858adc819b12/torchmetrics/regression/r2.py#L75-L87

torchmetrics/wrappers/multioutput.py
```diff
@@ -0,0 +1,112 @@
...
+        self.remove_nans = remove_nans
+
+    def update(self, *args: Any, **kwargs: Any) -> None:
+        """Update each underlying metric with the corresponding output."""
+        for i, metric in enumerate(self.metrics):
+            selected_args = apply_to_collection(
+                args, torch.Tensor, torch.index_select, dim=self.output_dim, index=torch.tensor(i)
+            )
+            selected_kwargs = apply_to_collection(
+                kwargs, torch.Tensor, torch.index_select, dim=self.output_dim, index=torch.tensor(i)
```
SkafteNicki, line 92: I am not sure if this will give device errors since index should can be on different device than the input
```suggestion
                args, torch.Tensor, torch.index_select, dim=self.output_dim, index=torch.tensor(i, device=self.device)
            )
            selected_kwargs = apply_to_collection(
                kwargs, torch.Tensor, torch.index_select, dim=self.output_dim, index=torch.tensor(i, device=self.device)
```

tests/wrappers/test_multioutput.py
```diff
@@ -0,0 +1,29 @@
+import torch
+
+from torchmetrics.classification import Accuracy
+from torchmetrics.regression import R2Score
+from torchmetrics.wrappers.multioutput import MultioutputWrapper
+
+
+def test_multioutput_wrapper():
```
SkafteNicki, line 8: For testing it would be great if we tested on random input instead of specific

torchmetrics/wrappers/multioutput.py
```diff
@@ -0,0 +1,112 @@
...
+        process_group:
+            Specify the process group on which synchronization is called. default: None (which selects the entire world)
+        dist_sync_fn: Callable = None,
+            Required for distributed training support. See torchmetrics docs for additional details.
+    """
+
+    def __init__(
+        self,
+        base_metric: Metric,
+        num_outputs: int = 1,
```
SkafteNicki, line 67: I would make this a required parameter since it is so essential for what the class is doing
```suggestion
        num_outputs: int,
```

SkafteNicki (commented):
Few comments, PR is looking good already :]

torchmetrics/wrappers/multioutput.py
an1lam: Done!

torchmetrics/wrappers/multioutput.py
an1lam: Done!

torchmetrics/wrappers/multioutput.py
an1lam: Added two, let me know if these seem good/are enough.

tests/wrappers/test_multioutput.py
an1lam: Done!

an1lam:
@SkafteNicki not exactly sure at what point I move this out of draft mode/whether I am the one supposed to do it. Let me know!

SkafteNicki:
@an1lam Changed it to review status, to get all test running.

tests/wrappers/test_multioutput.py
```diff
@@ -0,0 +1,48 @@
...
+        (
+            MultioutputWrapper(Accuracy(num_classes=3), num_outputs=2),
+            partial(_multioutput_sk_accuracy, num_outputs=2),
+            partial(torch.rand, 10, 3, 2),
+            partial(torch.randint, 3, (10, 2)),
+            2,
+        ),
+    ],
+)
+def test_multioutput_wrapper(metric, compare_metric, pred_generator, target_generator, num_rounds):
```
SkafteNicki, line 38: Based on how things are parameterized, it looks like it can fit into our standard testing framework like this:
https://github.com/PyTorchLightning/metrics/blob/e6d998e11fc2b93286d62beaa70f6a47647d77eb/tests/regression/test_r2.py#L71-L84
It would be great if you could do this, because the wrapper should then also be tested automatically for cuda and multi device (ddp) support

SkafteNicki (commented):
I am basically fine with the PR as it is, but it would be great to further improve the testing (I doubt it will show any problems as this is just a wrapper around other metrics) :]

tests/wrappers/test_multioutput.py
an1lam: Awesome, thanks for the tip! Will do.

an1lam: So in trying to get this to work, I realized it actually won't work because `forward` calls `reset` but in my current setup (and the bootstrap wrapper's by the way), setting `self.metrics = nn.ModuleList` doesn't register the underlying metrics as part of the state. As a result, `forward` will reset the state meaning `compute()` will return the result from the last batch rather than over all batches.

One idea I had was trying to make `metrics` part of state, but that won't work if I continue to use an `nn.ModuleList`. Alternatively, I was wondering if you think it makes sense to try and make the wrapper's state the union of the underlying states (with indexes to differentiate). Other ideas also welcome!

an1lam: Ok going to override `forward` actually, this isn't too messy.

tests/wrappers/test_multioutput.py
```diff
@@ -0,0 +1,143 @@
...
+from torchmetrics import Metric
+from torchmetrics.classification import Accuracy
+from torchmetrics.regression import R2Score
+from torchmetrics.wrappers.multioutput import MultioutputWrapper
+
+seed_all(42)
+
+
+class _MultioutputMetric(Metric):
+    """Multi-output version of the R2 score class for testing."""
```
an1lam, line 21: This should say "Test class that allows passing base metric as a class rather than its instantiation to the wrapper."

tests/wrappers/test_multioutput.py
```diff
@@ -0,0 +1,143 @@
...
+    def forward(self, *args, **kwargs):
+        """Run forward on the underlying metric."""
+        return self.metric(*args, **kwargs)
+
+    def reset(self) -> None:
+        """Reset the underlying metric state."""
+        self.metric.reset()
+
+
+num_classes = 3
```
an1lam, line 71: Use `NUM_CLASSES` from helper.

torchmetrics/wrappers/__init__.py
```diff
@@ -12,4 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from torchmetrics.wrappers.bootstrapping import BootStrapper  # noqa: F401
+from torchmetrics.wrappers.multioutput import MultioutputWrapper  # noqa: F401
```
an1lam, line 15: This is the wrong order. Should go after `MetricTracker`.

an1lam:
I'm not totally sure what to do about the failing tests on older versions. It looks like the issue is I'm using `movedim`, which isn't supported in older versions of Pytorch. Is there a decorator I can add that flags this new class as incompatible with these old versions? Is this even an acceptable solution?

SkafteNicki:
Hi @an1lam, please check your examples, they do not seem to render correctly:
image

SkafteNicki (commented):
LGTM :]

SkafteNicki (approved):

an1lam:
@SkafteNicki do you think I should wait until you merge #542 then (and merge `master` afterwards) to merge?

SkafteNicki:
@an1lam yes that PR should fix the last test that fails here :]

torchmetrics/wrappers/multioutput.py
```diff
@@ -0,0 +1,167 @@
...
+            This is sometimes unnecessary but harmless for metrics such as `R2Score` but useful
+            for certain classification metrics that can't handle additional 1-item dimensions.
+        compute_on_step: bool = True
+            Whether to recompute the metric value on each update step.
+        dist_sync_on_step: bool = False
+            Required for distributed training support. See torchmetrics docs for additional details.
+        process_group: Optional[Any]
+            Specify the process group on which synchronization is called. default: None (which selects the entire world)
+        dist_sync_fn: Callable = None,
+            Required for distributed training support. See torchmetrics docs for additional details.
```
Borda, line 65:
```suggestion
        base_metric:
            Metric being wrapped.
        num_outputs:
            Expected dimensionality of the output dimension. This parameter is
            used to determine the number of distinct metrics we need to track.
        output_dim:
            Dimension on which output is expected. Note that while this provides some flexibility, the output dimension
            must be the same for all inputs to update. This applies even for metrics such as `Accuracy` where the labels
            can have a different number of dimensions than the predictions. This can be worked around if the output
            dimension can be set to -1 for both, even if -1 corresponds to different dimensions in different inputs.
        remove_nans:
            Whether to remove the intersection of rows containing NaNs from the values passed through to each underlying
            metric. Proper operation requires all tensors passed to update to have dimension `(N, ...)` where N
            represents the length of the batch or dataset being passed in.
        squeeze_outputs:
            If true, will squeeze the 1-item dimensions left after `index_select` is applied.
            This is sometimes unnecessary but harmless for metrics such as `R2Score` but useful
            for certain classification metrics that can't handle additional 1-item dimensions.
        compute_on_step:
            Whether to recompute the metric value on each update step.
        dist_sync_on_step:
            Required for distributed training support. See torchmetrics docs for additional details.
        process_group:
            Specify the process group on which synchronization is called. default: None (which selects the entire world)
        dist_sync_fn:
            Required for distributed training support. See torchmetrics docs for additional details.
```
typing and defaults are taken from sinature

torchmetrics/wrappers/multioutput.py
```diff
@@ -0,0 +1,167 @@
...
+    return nan_idxs
+
+
+class MultioutputWrapper(Metric):
+    """Wrap a base metric to enable it to support multiple outputs.
+
+    Several torchmetrics metrics, such as :class:`torchmetrics.regression.spearman.SpearmanCorrcoef` lack support for
+    multioutput mode. This class wraps such metrics to support computing one metric per output.
+    Unlike specific torchmetric metrics, it doesn't support any
+    aggregation across outputs. This means if you set `num_outputs` to 2, `compute()` will return a Tensor of dimension
```
Borda, line 29:
```suggestion
    Unlike specific torchmetric metrics, it doesn't support any aggregation across outputs. This means if you set `num_outputs` to 2, `compute()` will return a Tensor of dimension
```

torchmetrics/wrappers/multioutput.py
```diff
@@ -0,0 +1,167 @@
...
+    (2, ...) where ... represents the dimensions the metric returns when not wrapped.
+
+    In addition to enabling multioutput support for metrics that lack it, this class also supports, albeit in a crude
+    fashion, dealing with missing labels (or other data). When `remove_nans` is passed, the class will remove the
+    intersection of NaN containing "rows" upon each update for each output. For example, suppose a user uses
+    `MultioutputWrapper` to wrap :class:`torchmetrics.regression.r2.R2Score` with 2 outputs, one of which occasionally
+    has missing labels for classes like `R2Score` is that this class supports removing NaN values
+    (parameter `remove_nans`) on a per-output basis. When `remove_nans` is passed the wrapper will remove all rows
```
Borda, line 36:
```suggestion
    In addition to enabling multioutput support for metrics that lack it, this class also supports, albeit in a crude
    fashion, dealing with missing labels (or other data). When ``remove_nans`` is passed, the class will remove the
    intersection of NaN containing "rows" upon each update for each output. For example, suppose a user uses
    `MultioutputWrapper` to wrap :class:`torchmetrics.regression.r2.R2Score` with 2 outputs, one of which occasionally
    has missing labels for classes like ``R2Score`` is that this class supports removing NaN values
    (parameter ``remove_nans``) on a per-output basis. When ``remove_nans`` is passed the wrapper will remove all rows
```

Borda (approved):

Issue #531, opened 2021-09-18
Metric not moved to device


cowwoc:
## 🐛 Bug

Version 1.4.7

Per https://torchmetrics.readthedocs.io/en/latest/pages/overview.html#metrics-and-devices if a metric is properly defined (identified as a child of a module) then it is supposed to be automatically moved to the same device as the module. Unfortunately, in my own project this does not occur.

When I run this code:

```
class MyModel(LightningModule):
  def __init__(self):
    self.accuracy= Accuracy()

  def forward(self, input):
    print(f"self.device: {self.device}")
    print(f"self.accuracy.device: {self.accuracy.device}")
```

I get:

```
self.device: cuda:0
self.accuracy.device: cpu
```

### Expected behavior

I expect the metric to be on `cuda:0`.

### Environment

- PyTorch Version (e.g., 1.0): 1.9.0+cu111
- OS (e.g., Linux): Windows 10.0.19043.1237
- How you installed PyTorch (`conda`, `pip`, source): pip
- Python version: 3.9.7
- CUDA/cuDNN version: 11.1
- GPU models and configuration: GeForce RTX 3080

cowwoc:
Interestingly, any states registered inside the metric *are* moved to the right device but `self.device` inside the metric evaluates to the wrong value. The reason this is relevant is I've got methods in my metric that create new Tensors on `self.device`. I suspect other users will also expect `self.device` to map to the same device used by the metric's state variables.

Borda:
see #340

cowwoc:
@Borda The linked issue does not resolve this problem. I am already doing what it recommends.

The documentation claims that the metric's device will be updated, but it is not. I consider this a bug report not a question.

Either the documentation or implementation are wrong. Please reopen this issue.

Borda:
I see, then the issue is in docs, no metrics is automatically moved unless you use it with PL logging... Mind send PR fix for it?

cowwoc:
The example code found at https://torchmetrics.readthedocs.io/en/latest/pages/overview.html#metrics-and-devices does say anything about having to use PL logging. Quoting the relevant parts:

> when properly defined inside a Module or LightningModule the metric will be be automatically move to the same device as the the module when using .to(device)

```
from torchmetrics import Accuracy, MetricCollection

class MyModule(torch.nn.Module):
    def __init__(self):
        ...
        # valid ways metrics will be identified as child modules
        self.metric1 = Accuracy()
[...]

PR #542, opened 2021-09-21
Fix child device

SkafteNicki:
# Before submitting

## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/531
Currently if a metric is a child of a module, the `device` property (only the property, the metric states are still correctly updated) of the metric will not get properly update (only if `.cuda()`, `.cpu()`, `to()` is directly called on the metric) when calling `.cuda()` on the module. This PR fixes it.

## PR review

## Did you have fun?

### DIFF
--- a/torchmetrics/metric.py
+++ b/torchmetrics/metric.py
@@ -506,14 +506,22 @@ class Metric(Module, ABC):
                 this._defaults[key] = [fn(v) for v in value]
 
             current_val = getattr(this, key)
+            device = None
             if isinstance(current_val, Tensor):
-                setattr(this, key, fn(current_val))
+                new_val = fn(current_val)
+                setattr(this, key, new_val)
+                device = new_val.device
             elif isinstance(current_val, Sequence):
-                setattr(this, key, [fn(cur_v) for cur_v in current_val])
+                new_val = [fn(cur_v) for cur_v in current_val]
+                setattr(this, key, new_val)
+                device = new_val[0].device if new_val else None
             else:
                 raise TypeError(
                     "Expected metric state to be either a Tensor" f"or a list of Tensor, but encountered {current_val}"
                 )
+
+            # make sure to update the device attribute
+            if device is not None: self._device = device
 
         # Additional apply to forward cache and computed attributes (may be nested)
         if this._computed is not None:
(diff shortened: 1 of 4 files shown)


cowwoc:
@SkafteNicki Thank you for working on this fix!

SkafteNicki:
@cowwoc if you have the time, please confirm that this fixes your original issue :]

torchmetrics/metric.py
```diff
@@ -515,6 +515,10 @@ def _apply(self, fn: Callable) -> Module:
                     "Expected metric state to be either a Tensor" f"or a list of Tensor, but encountered {current_val}"
                 )
 
+        # make sure to update the device attribute
+        # if the dummy tensor moves device by fn function we should also update the attribute
+        self._device = fn(torch.zeros(1, device=self.device)).device
```
justusschock, line 520: Why is this needed at all. Shouldn't this be handled by our wrapped `.to` function?

SkafteNicki: Because if you call `.to` on a parent module, it does actually not call the `.to` method of the child modules but it instead call `_apply` method. This change actually makes the wrapped `cuda`, `cpu`, `to` we have right now obsolete.

justusschock: I see, thanks :) But then do we also want to get the dtype there or is that something we only want with an explicit call of `.to` etc?

SkafteNicki: the `dtype` casting was actually remove in a earlier PR, because users were worried that if they run mixed-precision training that their metrics would be automatically casted to half precision given them wrong results. It can still be done but you need to explicitly call the special method `set_type`.

justusschock: Should we then extend this PR to remove the legacy wrappings for the device changes? Since this won't change behaviour or API it should be fine, right?

justusschock (approved):

an1lam:
Thanks @SkafteNicki!

Borda (approved):

torchmetrics/metric.py
SkafteNicki: @justusschock you are right, it is removed now :]

Issue #148, opened 2021-03-29
Metrics support for sweeping


maximsch2:
## 🚀 Feature

We would like to have tighter integration of metrics and sweeping. This requires a few features:
1. Knowing if `higher_is_better` (e.g. are we trying to minimize or maximize the metric in a sweep)
2. Knowing what value to optimize for. E.g. if a recall@precision metric returns both recall value and corresponding threshold, we want to optimize by maximizing recall and ignoring the threshold.

### Alternatives

An alternative implementation will be for each metric to have `is_better(left: TMetricResult, right: TMetricResult)` where TMetricResult is whatever `compute` returns. 

If we don't have it, people will have to have wrappers around the metrics to support this functionality in sweepers.

SkafteNicki:
I think it is great addition.
@maximsch2 is there a specific framework that you had in mind where this would give better integration?

maximsch2:
Yeah, maybe something like this: https://ax.dev/tutorials/tune_cnn.html

But I do imagine that any sort of sweeping requires us to be able to a) select the target metric b) compare two runs to see if the metric

SkafteNicki:
@maximsch2 what do you think about something like this:
```python
_REGISTER = {}

def register(metric, minimize, index=None):
    if minimize:
        compare_fn = torch.less
        init_val = torch.tensor(float("inf"))
    else:
        compare_fn = torch.greater
        init_val = -torch.tensor(float("inf"))
    _REGISTER[metric] = (minimize, compare_fn, init_val, index)

register(MeanSquaredError, True)


class MetricCompare:
    def __init__(self, metric):
        self.base_metric = metric
        minimize, compare_fn, init_val, index = _REGISTER[type(metric)]
        self._minimize = minimize
        self._compare_fn = compare_fn
        self._index = index
        self._init_val = init_val
        self._new_val = deepcopy(init_val)
        self._old_val = deepcopy(init_val)
        
    def update(self, *args, **kwargs):
        self.base_metric.update(*args, **kwargs)

    def compute(self):
        self._old_val = self._new_val
        val = self.base_metric.compute()
        self._new_val = val.detach()
        return val
    
    def reset(self):
        self.base_metric.reset()
        self._new_val = deepcopy(self._init_val)
        self._old_val = deepcopy(self._init_val)

    @property
    def has_improved(self):
        if self._index is None:
            return self._compare_fn(self._new_val, self._old_val)
        else:
[...]

PR #544, opened 2021-09-24
Metric sweeping

SkafteNicki:
# Before submitting

## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/148
Introduce the `higher_is_better` static property that determines if a metric is optimal when it is maximized (`True`), minimized (`False`) or this is undefined (`None`).
Still missing adding some metrics were I am not sure what the property should be.
cc: @maximsch2 

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `MultioutputWrapper` ([#510](https://github.com/PyTorchLightning/metrics/pull/510))
 
 
+- Added metric sweeping `higher_is_better` as constant attribute ([#544](https://github.com/PyTorchLightning/metrics/pull/544))
+
+
 ### Changed
 
 - `AveragePrecision` will now as default output the `macro` average for multilabel and multiclass problems ([#477](https://github.com/PyTorchLightning/metrics/pull/477))
--- a/tests/helpers/testers.py
+++ b/tests/helpers/testers.py
@@ -151,6 +151,8 @@ def _class_test(
     )
     with pytest.raises(RuntimeError):
         metric.is_differentiable = not metric.is_differentiable
+    with pytest.raises(RuntimeError):
+        metric.higher_is_better = not metric.higher_is_better
 
     # check that the metric is scriptable
     if check_scriptable:
--- a/torchmetrics/audio/si_sdr.py
+++ b/torchmetrics/audio/si_sdr.py
@@ -65,6 +65,7 @@ class SI_SDR(Metric):
     """
 
     is_differentiable = True
+    higher_is_better = True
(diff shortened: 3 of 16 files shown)


Borda:
I think it shall be property so it is not possible to change it, the same as is differentiable...
also, do we have another name candidate, this sounds a bit hacky to me... 
@PyTorchLightning/core-metrics @justusschock @maximsch2

SkafteNicki:
Fine with this being a property instead.
Name was suggested by @maximsch2

SkafteNicki:
@lucadiliello what is the correct value for all our retrieval metrics (you are the expert)?
Is a model optimal when they are maximized or minimized or is it not well defined?

lucadiliello:
> @lucadiliello what is the correct value for all our retrieval metrics (you are the expert)?
> Is a model optimal when they are maximized or minimized or is it not well defined?

Hi @SkafteNicki , all retrieval metrics should have `higher_is_better=True` apart from FallOut, which should be minimised.

Borda:
> Only reason I wanted this to be a class attribute was to give the user the possibility of actually getting this value even before the class was initialized.

that is a good point but in the same light why the `is_differentiable` would be needed first to be initialized?

we can ban certain attributes from being overwritten as for example:
```py
    def _ _setattr_ _(self, name, value):
        if name in ("higher_is_better", "is_diffrenetiable"):
            raise RuntimeError(f"Can't chnage const `{name}`."
        self._ _dict_ _[name] = value
```

SkafteNicki:
> 
> 
> > Only reason I wanted this to be a class attribute was to give the user the possibility of actually getting this value even before the class was initialized.
> 
> that is a good point but in the same light why the `is_differentiable` would be needed first to be initialized?
> 
> we can ban certain attributes from being overwritten as for example:
> 
> ```python
>     def _ _setattr_ _(self, name, value):
>         if name in ("higher_is_better", "is_diffrenetiable"):
>             raise RuntimeError(f"Can't chnage const `{name}`."
>         self._ _dict_ _[name] = value
> ```

You are completely right. I think we should also change `is_differentiable` to be this way and it looks like a good solution to me :]

SkafteNicki:
Anyone got a better name? On top of my head

* `higher_is_better`: options `True`, `False`, `None`
* `maximize` (or `minimize`): options `True`, `False`, `None`
* `direction` or `optim_direction`: options `maximize`, `minimize`, `None`
* ...

tchaton (approved):

Borda (approved):

SkafteNicki:
@Borda please wait with merge, going to add the property to some more metrics

tchaton (approved):
LGTM !

Borda:
@SkafteNicki mind check the failing tests on GPU?

PR #583, opened 2021-10-25
Remove error in ROC and AUROC

SkafteNicki:
# Before submitting

## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/discussions/581
We got so many complains that the following lines give many trouble for users when calculating roc and auroc values:
```python
    if fps[-1] <= 0:
        raise ValueError("No negative samples in targets, false positive value should be meaningless")
    fpr = fps / fps[-1]


    if tps[-1] <= 0:
        raise ValueError("No positive samples in targets, true positive value should be meaningless")
    tpr = tps / tps[-1]
```
while the errors are technically correct they break users workflow (for example when running sanity checking in lightning, you are likely to get this error because you sample a batch where a class is missing). Often they need to do something like:
```python
try:
    val = auroc(preds, target)
except ValueError:
    pass
```
This PR replaces the errors with a warning and overwrite `tps` and `fps` with zeros (this will result in a auroc value of 0 which corresponds to the worst case scenario for your classifier). 

## PR review

## Did you have fun?

### DIFF
--- a/docs/source/pages/lightning.rst
+++ b/docs/source/pages/lightning.rst
@@ -39,6 +39,12 @@ The example below shows how to use a metric in your `LightningModule <https://py
             # log epoch metric
             self.log('train_acc_epoch', self.accuracy.compute())
 
+.. note::
+    ``self.log`` in Lightning only supports logging of *scalar-tensors*. While the vast majority of metrics in torchmetrics returns a scalar tensor, some metrics such as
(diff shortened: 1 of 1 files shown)


jmerkow:
As I understand it, this will replace AUROC values when the label isn't present with 0s.  Is this correct?  I think a better behavior would be to ignore these values in the average calculation (under macro).  Perhaps this can be a parameter? i.e. skip_missing_labels=True/False or similiar.

PR #627, opened 2021-11-17
[precision.py] Wall-clock time speedup by performing top-k slicing at first and replacing `argsort` with `topk`

zuoxingdong:
## What does this PR do?

This PR proposes two speedup improvements:
1. By performing top-k slicing before indexing the target
2. Replace `argsort` with `topk`: it seems that the latter is much more efficient than former

## Pitch

A toy example to illustrate **[1]**:

```python
a = torch.randn(1024)
b = torch.randint(0, 2, size=(1024,))

%timeit -n 1000 b[torch.argsort(a, dim=-1, descending=True)][:256]
# 139 µs ± 13 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
%timeit -n 1000 b[torch.argsort(a, dim=-1, descending=True)[:256]]
# 124 µs ± 13.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

On a single CPU, it's about 10% faster.
Since retrieval metrics are computed per example over the dataset, the overall speedup could be significant in certain cases.

A toy example to illustrate **[2]**:

```python
a = torch.randn(1024)
b = torch.randint(0, 2, size=(1024,))

%timeit -n 1000 b[torch.argsort(a, dim=-1, descending=True)[:256]]
# 114 µs ± 17.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
%timeit -n 1000 b[a.topk(256, dim=-1)[1]]
# 46.7 µs ± 4.33 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

This leads to ~**2.4X** speedup in CPU mode which can be quite significant.


zuoxingdong:
@SkafteNicki It seems that those CI failures are irrelevant to this PR, if I'm not mistaken?

SkafteNicki:
@zuoxingdong the failing tests seems to be related to the changes in the PR
Skærmbillede 2021-11-17 104103

zuoxingdong:
> @zuoxingdong the failing tests seems to be related to the changes in the PR Skærmbillede 2021-11-17 104103

@SkafteNicki Thanks a lot for pointing this out!
It seems that the failures are due to an out-of-bound value of `k`. 
The existing implementation calculates the metrics simply using `k` to divide `relevant`. But I think we shouldn't do this when `k` is larger than the second dim. Instead, an upper bound cutoff is supposed to take place. 
What do you think?

SkafteNicki:
Hi @zuoxingdong, that makes sense to me and it also seems that the change have fixed the failing tests :]

SkafteNicki (approved):

zuoxingdong:
@SkafteNicki Thanks a lot for your feedback :)
One tiny detail, in the changelog, I guess the order between `torch.argsort` and `torch.topk` should be flipped around:

"Use `torch.argsort` instead of `torch.topk` in retrieval precision for speedup"
-> 
"Use `torch.topk` instead of `torch.argsort` in retrieval precision for speedup"

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Use `torch.argsort` instead of `torch.topk` in retrieval precision for speedup ([#627](https://github.com/PyTorchLightning/metrics/pull/627))
```
SkafteNicki, line 22:
```suggestion
- Use `torch.topk` instead of `torch.argsort` in retrieval precision for speedup ([#627](https://github.com/PyTorchLightning/metrics/pull/627))
```

Borda (approved):

zuoxingdong:
@SkafteNicki Since it goes well with `retrieval_precision`, I'd like to work on migrating similar changes to other retrieval metrics under functional folder (e.g. `average_precision`, `fall_out`, `hit_rate`, `ndcg`, `r_precision`, `recall`, `reciprocal_rank`) as they share the same logic of speeding up via `topk`

Do you think it makes sense? If so, I'll start working on that :)

SkafteNicki:
Hi @zuoxingdong,
Performance improvements are always welcome :]
Just remember to do a bit of testing (using `timeit` for example) to make sure that it actually improves the code.

lucadiliello:
Thank you @zuoxingdong for your work. If you need help just contact me, I will be happy to help. There have always been a performance issue with retrieval metrics due to the many sort needed. If you can speed up things it will be very nice.

PR #632, opened 2021-11-22
Torch-based mAP

### LINKED ISSUE #587: re-implement mAP instead of using COCOtools
## 🚀 Feature

As Coco is not maintained for a few years can we in the next step have our own implementation?
I have also opened a few fixes to the original repo but no reply for weeks/months

### Alternatives

https://github.com/PyTorchLightning/metrics/pull/467#discussion_r737354204

twsl:
## What does this PR do?

Fixes #587

Simply converted the pycocotools source code to pytorch and torchvision instead of numpy.
The rest in based on @tkupek work.
I plan to rename the file as soon as I know that it's working, right now I still need both for comparison.

## Before submitting

## PR review

## Did you have fun?


tkupek:
Very cool! Thanks for porting that.

torchmetrics/detection/map_new.py
```diff
@@ -0,0 +1,600 @@
...
+                for m, maxDet in enumerate(m_list):
+                    E = [evalImgs[Nk + Na + i] for i in i_list]
+                    E = [e for e in E if e is not None]
+                    if len(E) == 0:
+                        continue
+                    dtScores = torch.cat([e["dtScores"][:maxDet] for e in E])
+
+                    # different sorting method generates slightly different results.
+                    # mergesort is used to be consistent as Matlab implementation.
+                    inds = torch.argsort(dtScores, descending=True)
```
twsl, line 536: This seems to be a problem.

twsl:
Is that correct, that there are two scores while only one prediction @tkupek?
https://github.com/PyTorchLightning/metrics/blob/57320637b9e19b11f5600fb061ff7b0c841fc515/tests/detection/test_map.py#L61-L63

tkupek:
> Is that correct, that there are two scores while only one prediction @tkupek?
> 
> https://github.com/PyTorchLightning/metrics/blob/57320637b9e19b11f5600fb061ff7b0c841fc515/tests/detection/test_map.py#L61-L63

obviously a bug, you can remove the second score

torchmetrics/detection/map_new.py
tkupek: Weird that different sorting has an impact... Do you know how big it is?

torchmetrics/detection/map_new.py
```diff
@@ -0,0 +1,600 @@
...
+
+    .. note::
+        This metric requires you to have `torchvision` version 0.8.0 or newer installed (with corresponding
+        version 1.7.0 of torch or newer). Please install with ``pip install torchvision`` or
+        ``pip install torchmetrics[detection]``.
+
+    .. note::
+        As the pycocotools library cannot deal with tensors directly, all results have to be transfered
+        to the CPU, this might have an performance impact on your training.
+
```
tkupek, line 147:
```suggestion
```

torchmetrics/detection/map_new.py
```diff
@@ -0,0 +1,600 @@
...
+    For an example on how to use this metric check the `torchmetrics examples\
+    <https://github.com/PyTorchLightning/metrics/blob/master/tm_examples/detection_map.py>`_
+
+    .. note::
+        This metric is a wrapper for the
+        `pycocotools <https://github.com/cocodataset/cocoapi/tree/master/PythonAPI/pycocotools>`_,
+        which is a standard implementation for the mAP metric for object detection. Using this metric
+        therefore requires you to have `pycocotools` installed. Please install with ``pip install pycocotools`` or
+        ``pip install torchmetrics[detection]``.
```
tkupek, line 137:
```suggestion
    .. note::
        This metric is following the mAP implementation of
        `pycocotools <https://github.com/cocodataset/cocoapi/tree/master/PythonAPI/pycocotools>`_,
        , a standard implementation for the mAP metric for object detection.
```

torchmetrics/detection/map_new.py
```diff
@@ -0,0 +1,600 @@
...
+    Raises:
+        ImportError:
+            If ``torchvision`` is not installed or version installed is lower than 0.8.0
+        ValueError:
+            If ``class_metrics`` is not a boolean
+    """
+
+    def __init__(
+        self,
+        in_fmt: str = "xyxy",
```
tkupek, line 178: I'd prefer a self-speaking name as `box_format`

torchmetrics/detection/map_new.py
```diff
@@ -0,0 +1,600 @@
...
+        num_classes: Optional[int] = None,
```
tkupek, line 179: In the current implementation we could avoid the num_classes and determine those with the collected `preds` and `targets`. Have a look at the `get_classes` method of my latest PR.

torchmetrics/detection/map_new.py
```diff
@@ -0,0 +1,600 @@
...
+        iou_thresholds: Optional[List[float]] = None,
```
tkupek, line 180: I think it makes sense to define the thresholds, as it's mostly unnecessary to compute all (as pycocotools does it).
Let' have a meaningful default.

```
(AP) @[ IoU=0.50:0.95 | area= all | maxDets=100 ]
(AR) @[ IoU=0.50:0.95 | area= all | maxDets=100 ] 
```

torchmetrics/detection/map_new.py
twsl: Done

torchmetrics/detection/map_new.py
twsl: I prefer to calculate the thresholds because especially the `rec_thresholds` is quite huge to hardcode it.

torchmetrics/detection/map_new.py
twsl: I'm not sure, it actually has a big impact, but I cant validate yet, cause there is still an issue inside the `_evaluate_image` function that I have to fix tomorrow

twsl: Tests pass, so let's hope, this works for now

justusschock:
@twsl thanks a lot for this? Have you checked the results against the ones from pycocotools?

I'm asking since @mibaumgartner mentioned these could differ simply when switching numpy's version of sorting to torch's version.

twsl:
> @twsl thanks a lot for this? Have you checked the results against the ones from pycocotools?
> 
> I'm asking since @mibaumgartner mentioned these could differ simply when switching numpy's version of sorting to torch's version.

@justusschock I am aware of that and there is still a comment left in the code from pycocotools highlighting the potential problem.
The pytorch implementation passes all tests so far, but if @mibaumgartner can supply a critical samples that can cause issues, we could properly evaluate that

torchmetrics/detection/map.py
```diff
@@ -160,22 +136,24 @@ class MAP(Metric):
...
 
-    .. note::
-        As the pycocotools library cannot deal with tensors directly, all results have to be transfered
-        to the CPU, this might have an performance impact on your training.
-
     Args:
+        box_format:
+            Input format of given boxes. Supported formats are [‘xyxy’, ‘xywh’, ‘cxcywh’].
+        iou_thresholds:
+            IoU thresholds for evaluation.
```
SkafteNicki, line 152: please add more description
```suggestion
            IoU thresholds for evaluation. If set to `None` it corresponds to ... . Else provide a list of floats 
```

torchmetrics/detection/map.py
```diff
@@ -160,22 +136,24 @@ class MAP(Metric):
...
+        rec_thresholds:
+            Recall thresholds for evaluation.
```
SkafteNicki, line 154: please add more description
```suggestion
            Recall thresholds for evaluation. . If set to `None` it corresponds to ... . Else provide a list of floats 
```

torchmetrics/detection/map.py
```diff
@@ -160,22 +136,24 @@ class MAP(Metric):
...
+        max_detection_thresholds:
+            Thresholds on max detections per image.
```
SkafteNicki, line 156: please add more description
```suggestion
            Thresholds on max detections per image. If set to `None` will use thresholds `[1, 10, 100]`. Else please provide a list of floats.
```

torchmetrics/detection/map.py
```diff
@@ -214,17 +194,28 @@ def __init__(
...
-                "`MAP` metric requires that `pycocotools` installed."
-                " Please install with `pip install pycocotools` or `pip install torchmetrics[detection]`"
-            )
         if not (_TORCHVISION_AVAILABLE and _TORCHVISION_GREATER_EQUAL_0_8):
             raise ImportError(
                 "`MAP` metric requires that `torchvision` version 0.8.0 or newer is installed."
                 " Please install with `pip install torchvision` or `pip install torchmetrics[detection]`"
             )
 
+        self.box_format = box_format
```
SkafteNicki, line 203: please add input validation
```suggestion
        allowed_box_formats = (‘xyxy’, ‘xywh’, ‘cxcywh’)
        if box_format not in allowed_box_formats:
            raise ValueError(
              f"Expected argument `box_format` to be one of {allowed_box_formats} but got {box_format}"
            )
        self.box_format = box_format
```

torchmetrics/detection/map.py
```diff
@@ -201,6 +177,10 @@ class MAP(Metric):
 
     def __init__(
         self,
+        box_format: str = "xyxy",
+        iou_thresholds: Optional[List[float]] = None,
+        rec_thresholds: Optional[List[float]] = None,
+        max_detection_thresholds: Optional[List[float]] = None,
```
SkafteNicki, line 183: should this be int?
```suggestion
        max_detection_thresholds: Optional[List[int]] = None,
```

SkafteNicki (commented):
Please add docstring to as many methods as possible

tests/detection/test_map.py
```diff
@@ -59,7 +55,7 @@
             ),  # coco image id 74
             dict(
                 boxes=torch.Tensor([[0.00, 2.87, 601.00, 421.52]]),
-                scores=torch.Tensor([0.699, 0.423]),
+                scores=torch.Tensor([0.699]),
```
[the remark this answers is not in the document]
twsl: That score it probably a left over from an additional ground truth. It should have been caught by the `_input_validator`, I fixed the issue. For previous calculation it probably didn't matter because it was not returned by any index based loop.

mibaumgartner:
@twsl As far as I saw the current tests do not cover the sorting "problem" and thus won't fail.

As long as the predictions don't have the same score, the order (after sorting) is well defined (which is the case for the current test suite). The problem only occurs if multiple predictions have the same score (AFAIK this is not covered by the tests) which leads to an "undefined" sequence. Depending on the ordering of the predictions with the same score, the metric changes. This can happen somewhat frequently especially when the method predicts multiple objects with a score of `1.0`.

twsl:
I haven't found a suitable example yet, but `torch.argsort` seems stable. I cannot produce any sequence of numbers where argsorting behaves differently from numpy using mergesort.

justusschock (approved):

SkafteNicki:
@twsl some test are still failing with:
```
ValueError: Input boxes, labels and scores of sample 0 in preds have a different length (expected 1 labels and scores, got 0 labels and 0)
```
do the input checking also need to change?

twsl:
@SkafteNicki i think, the target in the first update in `test_empty_preds` should not be accepted and the input validator correctly rejects it because the dimensions mismatch. I think, we could remove it.

torchmetrics/detection/map.py
```diff
@@ -12,87 +12,74 @@
...
-    _TORCHVISION_AVAILABLE,
-    _TORCHVISION_GREATER_EQUAL_0_8,
-)
+from torchmetrics.utilities.imports import _TORCHVISION_AVAILABLE, _TORCHVISION_GREATER_EQUAL_0_8
 
 if _TORCHVISION_AVAILABLE and _TORCHVISION_GREATER_EQUAL_0_8:
-    from torchvision.ops import box_convert
+    from torchvision.ops import box_area, box_convert, box_iou
 else:
     box_convert = None
```
Borda, line 26:
```suggestion
    box_convert = box_iou = box_area = None
```

torchmetrics/detection/map.py
```diff
@@ -12,87 +12,74 @@
...
-    mar_small: Tensor
-    mar_medium: Tensor
-    mar_large: Tensor
-    map_per_class: Tensor
-    mar_100_per_class: Tensor
+    def __getattr__(self, key: str) -> Tensor:
+        if key in self:
+            return self[key]
+        else:
+            raise AttributeError(f"No such attribute: {key}")
```
Borda, line 40: isn't this default behavior?

torchmetrics/detection/map.py
```diff
@@ -12,87 +12,74 @@
...
 
-# noinspection PyMethodMayBeStatic
-class WriteToLog:
-    """Logging class to move logs to log.debug()."""
 
-    def write(self, buf: str) -> None:  # skipcq: PY-D0003, PYL-R0201
-        for line in buf.rstrip().splitlines():
-            log.debug(line.rstrip())
+class MAPMetricResults(BaseMetricResults):
```
Borda, line 52: would Data class be a better solution here?
https://docs.python.org/3/library/dataclasses.html

torchmetrics/detection/map.py
```diff
@@ -12,87 +12,74 @@
...
-    def flush(self) -> None:  # skipcq: PY-D0003, PYL-R0201
-        for handler in log.handlers:
-            handler.flush()
+    __slots__ = ("map", "map_50", "map_75", "map_small", "map_medium", "map_large")
 
-    def close(self) -> None:  # skipcq: PY-D0003, PYL-R0201
-        for handler in log.handlers:
-            handler.close()
 
+class MARMetricResults(BaseMetricResults):
```
Borda, line 58: with data class, we could also define types...

torchmetrics/detection/map.py
```diff
@@ -160,22 +147,31 @@ class MAP(Metric):
...
+            with step `0.05`. Else provide a list of floats.
+            default: ``None``
+        rec_thresholds:
+            Recall thresholds for evaluation. If set to `None` it corresponds to the stepped range `[0,...,1]`
+            with step `0.01`. Else provide a list of floats.
+            default: ``None``
+        max_detection_thresholds:
+            Thresholds on max detections per image. If set to `None` will use thresholds `[1, 10, 100]`.
+            Else please provide a list of ints.
+            default: ``None``
```
Borda, line 174:
```suggestion
        box_format:
            Input format of given boxes. Supported formats are [‘xyxy’, ‘xywh’, ‘cxcywh’].
        iou_thresholds:
            IoU thresholds for evaluation. If set to `None` it corresponds to the stepped range `[0.5,...,0.95]`
            with step `0.05`. Else provide a list of floats.
        rec_thresholds:
            Recall thresholds for evaluation. If set to `None` it corresponds to the stepped range `[0,...,1]`
            with step `0.01`. Else provide a list of floats.
        max_detection_thresholds:
            Thresholds on max detections per image. If set to `None` will use thresholds `[1, 10, 100]`.
            Else please provide a list of ints.
```
all defaults are automatically taken from the function/method signature

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+                else _fix_empty_tensors(item["boxes"])
+            )
             self.groundtruth_labels.append(item["labels"])
 
+    def _get_classes(self) -> List:
+        """Returns a list of unique classes found in groundtruth and detection data."""
+        if len(self.detection_labels) > 0 or len(self.groundtruth_labels) > 0:
+            return torch.cat(self.detection_labels + self.groundtruth_labels).unique().tolist()
+        else:
+            return []
```
Borda, line 315:
```suggestion
        return []
```

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+                        # if match successful and best so far, store appropriately
+                        iou = ious[d, g]
+                        m = g
+                    # if match made store id of match for both dt and gt
+                    if m == -1:
+                        continue
+
+                    dt_ignore[tind, d] = gt_ignore[m]
+                    dt_matches[tind, d] = True
+                    gt_matches[tind, m] = True
```
Borda, line 425: it may make code easier to read if we move this body to a separate private function/method...

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+                Calculate average precision. Else calculate average recall.
+                default: `True`
+            iou_threshold:
+                IoU threshold. If set to `None` it all values are used. Else results are filtered.
+            area_range:
+                Bounding box area range key.
+                default: ``all``
+            max_dets:
+                Maximum detections.
+                default: ``100``
```
Borda, line 462:
```suggestion
            ap:
                Calculate average precision. Else calculate average recall.
            iou_threshold:
                IoU threshold. If set to `None` it all values are used. Else results are filtered.
            area_range:
                Bounding box area range key.
            max_dets:
                Maximum detections.
```

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+            for area in area_ranges
+            for id in img_ids
+        ]
+
+        T = len(self.iou_thresholds)
+        R = len(self.rec_thresholds)
+        K = len(class_ids)
+        A = len(self.bbox_area_ranges)
+        M = len(self.max_detection_thresholds)
+        I = len(img_ids)  # noqa: E741
```
Borda, line 516: lets avoid using capital letters fro variables, capitals are reserved for class names :]

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+        T = len(self.iou_thresholds)
+        G = len(gt)
+        D = len(dt)
+        gt_matches = torch.zeros((T, G), dtype=torch.bool)
+        dt_matches = torch.zeros((T, D), dtype=torch.bool)
+        gt_ignore = ignore_area_sorted
+        dt_ignore = torch.zeros((T, D), dtype=torch.bool)
+        if len(ious) > 0:
+            for tind, t in enumerate(self.iou_thresholds):
+                for d in range(D):
```
Borda, line 402: for readability pls name variables with at least three letters... otherwise, it is impossible to use any replace tool in future

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+
+                        inds = torch.searchsorted(rc, self.rec_thresholds, right=False)
+                        # TODO: optimize
+                        try:
+                            for ri, pi in enumerate(inds):  # range(min(len(inds), len(pr))):
+                                # pi = inds[ri]
+                                q[ri] = pr[pi]
+                                ss[ri] = dt_scores_sorted[pi]
+                        except Exception:
+                            pass
```
Borda, line 575: let's make this as one protected method

torchmetrics/detection/map.py
```diff
@@ -279,20 +291,339 @@ def update(self, preds: List[Dict[str, Tensor]], target: List[Dict[str, Tensor]]
...
+                        precision[t, :, k, a, m] = q
+                        scores[t, :, k, a, m] = ss
```
Borda, line 577: let's make this another private method...
the reason is that it is too complex and so harder to follow and eventually with using/calling internal functions we can increase correctness with intermediate typing check :rabbit:

torchmetrics/detection/map.py
twsl: Using this you get the correct error message, an `AttributeError` instead of a `KeyError`

torchmetrics/detection/map.py
twsl: The problem with dataclasses is, that depending on the max threshold property the variable names could change. As soon as you dont use the default COCO parameters, a dictionary is more flexibel

torchmetrics/detection/map.py
twsl: refactor method exists ;)
But I get the point

torchmetrics/detection/map.py
Borda: I see so maybe add a comment to the code, just the same as you told me here...

Borda:
@twsl over all it looks Great! mind checking the last failing test for new PT versions: `tests/detection/test_map.py::test_empty_preds - ValueError: Input boxe...`

CHANGELOG.md
```diff
@@ -28,8 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Scalar metrics will now consistently have additional dimensions squeezed ([#622](https://github.com/PyTorchLightning/metrics/pull/622))
-
-
+- Migrate MAP metrics from pycocotools to PyTorch ([#632](https://github.com/PyTorchLightning/metrics/pull/632))
```
Borda, line 31:
```suggestion


- Migrate MAP metrics from pycocotools to PyTorch ([#632](https://github.com/PyTorchLightning/metrics/pull/632))
```

Borda (approved):

torchmetrics/detection/map.py
```diff
@@ -12,90 +12,77 @@
...
-    mar_1: Tensor
-    mar_10: Tensor
-    mar_100: Tensor
-    mar_small: Tensor
-    mar_medium: Tensor
-    mar_large: Tensor
-    map_per_class: Tensor
-    mar_100_per_class: Tensor
+    def __getattr__(self, key: str) -> Tensor:
+        if key in self:
```
Borda, line 37:
```suggestion
        # Using this you get the correct error message, an AttributeError instead of a KeyError
        if key in self:
```

PR #688, opened 2021-12-19
Add Test strategy

### LINKED ISSUE #675: Test all metrics on GPU
## 🚀 Feature

Add an option to run tests on a single GPU.

### Motivation

Currently tests for metrics are usually run via  [`run_class_metric_test`](https://github.com/PyTorchLightning/metrics/blob/master/tests/helpers/testers.py#L391), which defaults to CPU if only a sinlge GPU is present. This can cause issues with CUDA to go into a release unnoticed.

### Pitch

Fix quietly switching to cpu even if ddp=False and GPUs are available.
Updating `run_class_metric_test` to test on single gpu, cpu and both using ddp if available/selected.

### Alternatives

Accept incomplete test coverage.

twsl:
## What does this PR do?

Fixes #675

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -2,6 +2,7 @@ coverage>5.2
 codecov>=2.1
 pytest>=6.0
 pytest-cov>2.10
+pytest-cases>=3.6.5
 # pytest-xdist
 # pytest-flake8
 flake8
--- a/tests/detection/test_map.py
+++ b/tests/detection/test_map.py
@@ -16,8 +16,9 @@ from collections import namedtuple
 
 import pytest
 import torch
+from pytest_cases import parametrize_with_cases
 
-from tests.helpers.testers import MetricTester
+from tests.helpers.testers import MetricTester, MetricTesterDDPCases
 from torchmetrics.detection.map import MAP
 from torchmetrics.utilities.imports import _TORCHVISION_AVAILABLE, _TORCHVISION_GREATER_EQUAL_0_8
 
@@ -174,8 +175,8 @@ class TestMAP(MetricTester):
 
     atol = 1e-1
 
-    @pytest.mark.parametrize("ddp", [False, True])
-    def test_map(self, ddp):
+    @parametrize_with_cases("ddp,device", cases=MetricTesterDDPCases, has_tag="strategy")
+    def test_map(self, ddp, device):
         """Test modular implementation for correctness."""
         self.run_class_metric_test(
             ddp=ddp,
@@ -185,6 +186,7 @@ class TestMAP(MetricTester):
             sk_metric=_compare_fn,
             dist_sync_on_step=False,
             check_batch=False,
+            device=device,
             metric_args={"class_metrics": True},
         )
 
--- a/tests/helpers/testers.py
+++ b/tests/helpers/testers.py
@@ -349,8 +349,7 @@ class MetricTesterDDPCases:
     def case_ddp_false_device_cpu(self):
         return False, "cpu"
 
-    # @pytest.mark.skipif(os.cpu_count() < 2, reason="More than one CPU Core required")
-    @case(tags="strategy")  # order is important, @case after @pytest.mark or markers as argument for @case
+    @case(tags="strategy")
     def case_ddp_true_device_cpu(self):
         return True, "cpu"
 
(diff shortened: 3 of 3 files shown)


Borda:
well seems that some tests at the beginning is hanging, have you tried to run it locally?

tests/classification/test_auc.py
```diff
@@ -55,22 +55,29 @@ def sk_auc(x, y, reorder=False):
 
 @pytest.mark.parametrize("x, y", _examples)
 class TestAUC(MetricTester):
-    @pytest.mark.parametrize("ddp", [False])
+    @pytest.mark.parametrize(MetricTesterDDPCases.name_strategy(), MetricTesterDDPCases.cases_strategy())
```
twsl, line 58: DDP apparently fails and therefore was probably excluded in the tests, yet there is no mention in the docs, that DDP is not supported.

twsl: @Borda @SkafteNicki any idea how to handle this? removing the full ddp/device strategy and opening an issue to fix auc would be my recommendation

tests/audio/test_pesq.py
```diff
@@ -75,25 +75,28 @@ def average_metric(preds, target, metric_func):
 class TestPESQ(MetricTester):
     atol = 1e-2
 
-    @pytest.mark.parametrize("ddp", [True, False])
+    @pytest.mark.parametrize(MetricTesterDDPCases.name_strategy(), MetricTesterDDPCases.cases_strategy())
```
Borda, line 78: I would say it is cleaner if it will be a constant or function returning tuple:
```py
def ddp_name_options():
    return "ddp", (True, False)
```
and later:
```py
@pytest.mark.parametrize(*ddp_name_options())
```

Borda: but anyway why is this needed?

Borda (commented):
hi, thx for this PR, just very sure if it is really needed... what you try to set here is running CPU and GPU in a single test call, but even now we are trying to optimize the test to run as efficiently as possible, so on CPU machine run just CPU version and on GPU run CUDA version (and skip CPU version as it was run already elsewhere)
Said so I think that is also by user expectation, test on the best available resources, so on GPU machine run CUDA tests and on CPU imagine the rest... so if you have a case that you have GPU and you want to run CPU test you can simply disable GPU visibility for the particular run, for example: `CUDA_VISIBLE_DEVICES=-1 pytest ...`

tests/audio/test_pesq.py
twsl: The reason I added this stategy was that now all available test cases are generated and you can explicitly test them locally. This way you can guarantee full coverage, don't depend on the ci machines or some launch flags and each test states on which device it runs. This way you dont have to check the machine stats to e.g. find out, if a test failed on single or multi gpu.
In order to optimize the tests, having an additional skip condition with an env var could achieve the same. Especially for test optimization you could then run ddp=false on a machine with a single gpu and ddp=True on one with multiple.
I'm fine putting it into a single function, my idea was to stick with two parameter style of `parametrize`

justusschock: I agree with @Borda that it would be better to  return a tuple there.

For the general purpose: I think we should split this between arguments for every test (like devices and maybe also ddp strategies) for which we should use pytest fixtures (that's what they are there for) and then you could call `pytest --devices=XX tests`. For test-specific arguments this approach is fine IMO

tests/classification/test_auc.py
SkafteNicki: Its hard for me to remember, but basically AUC needs ordered input to be working and while we are testing on ordered input when we move to `ddp` it somehow gets unordered. I wonder if we should just set the `reorder` flag to `True` whenever `ddp=True`.

tests/audio/test_pesq.py
twsl: I'm unsure how to continue with this. Any suggestions?

maximsch2:
@SkafteNicki , I think it makes sense to merge this first before merging #867 to make sure things continue working correctly on GPUs.

Borda:
Thank you for your suggestion, but atm we are in the process of additional test optimization :innocent:

Issue #143, opened 2021-03-29
Smart update of Collection of CompositionalMetrics


victorjoos:
## 🚀 Feature

When updating metrics that are composed of other metrics there are two ways of dealing with updating too many times : 

 * Always update the underlying metric (the pytorch lightning way)
 * Never update the underlying metric (the pytorch ignite way: https://pytorch.org/ignite/metrics.html#ignite.metrics.MetricsLambda)

I don't think there is a clean way of only updating the necessary metrics in the general case (when you're just updating all the metrics yourself), but I think that when you combine your metrics in a collection, it could be useful to only update the "base" metric, instead of all metrics.

### Motivation

I often want to use a base metric multiple times, and then I have to be careful not to update too many of them. A somewhat convoluted example (because the f1 score is already implemented) : 

```python
prec = Precision()
recall = Recall()
f1 = 2 * (prec * recall) / (prec + recall)
prec.update(pred, gt)
recall.update(pred, gt)
f1.update(pred, gt) # Shouldn't do this, because it updates prec and recall twice. 
```

### Pitch
Continuing last example : 

```python
collection = MetricCollection([prec, recall, f1])
collection.update(pred, gt)
```
This should only update prec and recall once.

### Alternatives

The alternative is to always define metrics from scratch, but this causes duplication of computation during the update phase.

Borda:
@justusschock what do you think about it? ;]

SkafteNicki:
I had a similar proposal here: https://github.com/PyTorchLightning/metrics/issues/76

justusschock:
@Borda Yes, this is necessary at some point. The only question here is: Should we define "compatible" metrics, that share their internal state or should this be defined by the user?

IF we define it, we have to maintain this list, but we know about the implementation details of the metrics. If the user defines it, we don't need to care about that list but expect the user to know about our internals.

PR #709, opened 2022-01-04
Smart update of metric collection

SkafteNicki:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/143

Update `MetricCollection` such that metrics that have similar metric states (`Precision` and `Recall` for example) we only calculate the metric state for one of them and then broadcast it to all other metrics with similar metric states.
After the first call to update we checks if the metric states matches between any metrics. If they do we merge them into a compute group of metrics where we will only update the first and broadcast the state to the remaining metrics in the group

## Before submitting

## PR review

## Did you have fun?


torchmetrics/__init__.py
```diff
@@ -2,7 +2,7 @@
 import logging as __logging
 import os
 
-from torchmetrics.__about__ import *  # noqa: F401, F403
+from torchmetrics.__about__ import *  # noqa: F403
```
SkafteNicki, line 5:
```suggestion
from torchmetrics.__about__ import * # noqa: F401, F403
```

torchmetrics/__init__.py
```diff
@@ -25,7 +25,7 @@
     SignalDistortionRatio,
     SignalNoiseRatio,
 )
-from torchmetrics.classification import (  # noqa: E402, F401
+from torchmetrics.classification import (  # noqa: E402
```
SkafteNicki, line 28:
```suggestion
from torchmetrics.classification import ( # noqa: E402, F401
```

torchmetrics/__init__.py
```diff
@@ -185,6 +196,7 @@
     "StatScores",
     "SumMetric",
     "SymmetricMeanAbsolutePercentageError",
+    "WER",
```
SkafteNicki, line 199:
```suggestion
```

torchmetrics/collections.py
```diff
@@ -117,11 +125,83 @@ def update(self, *args: Any, **kwargs: Any) -> None:
...
+            state2 = getattr(metric2, key)
+
+            if type(state1) != type(state2):
+                return False
+
+            if isinstance(state1, Tensor) and isinstance(state2, Tensor):
+                if state1.shape != state2.shape:
+                    return False
+                if not torch.allclose(state1, state2):
+                    return False
```
justusschock, line 193:
```suggestion
                return state1.shape == state2.shape and torch.allclose(state1, state2):
```

torchmetrics/collections.py
```diff
@@ -117,11 +125,83 @@ def update(self, *args: Any, **kwargs: Any) -> None:
...
+
+            if isinstance(state1, list) and isinstance(state2, list):
+                if any(s1.shape != s2.shape for s1, s2 in zip(state1, state2)):
+                    return False
+                if any(torch.allclose(s1, s2) for s1, s2 in zip(state1, state2)):
```
justusschock, line 198:
```suggestion
                if not all(torch.allclose(s1, s2) for s1, s2 in zip(state1, state2)):
```

When moving it to one line as suggested below, this will be outdated anyways, but currently this behaviour is wrong I think :) It either should be `any(not torch.allclose(...))` or `not all(torch.allclose(...))`

torchmetrics/collections.py
```diff
@@ -117,11 +125,83 @@ def update(self, *args: Any, **kwargs: Any) -> None:
...
+                    return False
```
justusschock, line 199:
```suggestion
                return all(s1.shape == s2.shape and torch.allclose(s1, s2) for s1, s2 in zip(state1, state2))
```

SkafteNicki:
Based on the following script:
```python
from torchmetrics import Accuracy, Precision, Recall, MetricCollection
import torch
import time
import numpy as np

def experiment(enable, steps):

    m = MetricCollection([Accuracy(num_classes=10), Precision(num_classes=10), Recall(num_classes=10)], enable_compute_groups=enable)

    for _ in range(steps):
        preds = torch.randn(256, 10).softmax(dim=-1)
        target = torch.randint(10, (256,))
        m.update(preds, target)

    m.compute()

times = [ ]
for _ in range(5):
    start = time.time()
    experiment(True)
    times.append(time.time() - start)

times2 = [ ]
for _ in range(5):
    start = time.time()
    experiment(False)
    times2.append(time.time() - start)

print(f"{np.mean(times)}+-{np.std(times)}")
print(f"{np.mean(times2)}+-{np.std(times)}")
```
I get 
```
0.42464628219604494+-0.6570260998311238
0.2263005256652832+-0.6570260998311238
```
for `steps=100` and
```
8.417768621444703+-1.1484917218472177
18.352816343307495+-1.1484917218472177
```
for `steps=10000`. The dynamical creating of the compute groups are an expensive search, and therefore not worth it if the user is only doing a few steps. However, as the number of steps increase it will be mush faster. In the example we have three metrics that share a metric state and potentially the new implementation can therefore be close to 3x times faster when the number of steps are large enough.

justusschock:
@SkafteNicki: What about also adding an option for the user to provide compute_groups manually? Something like `compute_groups = False | "auto" | [[0, 1, 2], [3,4]] | [["accuracy", "f1"], ["precision", "recall]]`?

torchmetrics/collections.py
```diff
@@ -45,6 +44,13 @@ class name as key for the output dict.
 
         postfix: a string to append after the keys of the output dict
 
+        enable_compute_groups:
+            By defualt the MetricCollection will try to reduce the computations needed for the metrics in the collection
```
justusschock, line 48:
```suggestion
            By default the MetricCollection will try to reduce the computations needed for the metrics in the collection
```

SkafteNicki:
> @SkafteNicki: What about also adding an option for the user to provide compute_groups manually? Something like `compute_groups = False | "auto" | [[0, 1, 2], [3,4]] | [["accuracy", "f1"], ["precision", "recall]]` for disabling, dynamic search, when using with a Sequence and with a dict?
> 
> That way we would only do some sanity checks (should be a lot cheaper than exhaustive search) and avoid the search at all.
> 

Sounds like a good idea. I implement that :]

> Also you check every metric against every other. But you only need to check against one metric in each group, right?

Actually not every metric is run against every other. I already only compare the first in each group to the first in any other group:
https://github.com/PyTorchLightning/metrics/blob/3ada9afb7d4948ceabc9c06aa862dfed58b24e12/torchmetrics/collections.py#L160-L161

torchmetrics/collections.py
```diff
@@ -45,6 +44,14 @@ class name as key for the output dict.
 
         postfix: a string to append after the keys of the output dict
 
+        compute_groups:
+            By defualt the MetricCollection will try to reduce the computations needed for the metrics in the collection
```
justusschock, line 48:
```suggestion
            By default the MetricCollection will try to reduce the computations needed for the metrics in the collection
```

again :D

justusschock (commented):
Just a nit :)

@SkafteNicki did this yield the expected performance boost?

torchmetrics/collections.py
Borda: I see how you enjoy it in your office - just find anything :D

Borda:
> @SkafteNicki did this yield the expected performance boost?

would be cool if we could mention any benchmark on docs to give user reason to use it :rabbit:

torchmetrics/collections.py
justusschock: I just looked through it again since it popped up in my notifications, but yes, I quite enjoyed it when I saw this :P

docs/source/pages/overview.rst
```diff
@@ -292,6 +292,14 @@ inside your LightningModule
     have the same call signature. If this is not the case, input that should be
     given to different metrics can given as keyword arguments to the collection.
 
+An additional advantage of using the ``MetricCollection`` object is that it will
+automatically try to reduce the computations needed by finding groups of metrics
+that share the same underlying metric state. If such metrics are found only one
+of them are actually updated and the updated state will be broadcastet to the rest
```
justusschock, line 298:
```suggestion
of them is actually updated and the updated state will be broadcasted to the rest
```

docs/source/pages/overview.rst
```diff
@@ -292,6 +292,14 @@ inside your LightningModule
...
+of the metrics. In the example above, this will lead to a 2x-3x lower computational
+cost compared to disabling this feature. See the `compute_groups` argument
+for more info.
```
justusschock, line 301:
```suggestion
cost compared to disabling this feature. 
However, this speedup comes with a high cost upfront, where the state-groups have to be determined after the first update. This cost can be significant for very low steps (approx. up to 100) but still leads to an overall speedup for everything beyond that. In case the groups are known beforehand, these can also be set manually to avoid the cost of the dynamic search.
See the :param:`torchmetrics.MetricCollection.compute_groups` argument for more information on this topic.
```

justusschock (approved):
LGTM :) added some more text on potential slowdowns

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added new image metric `UniversalImageQualityIndex` ([#824](https://github.com/PyTorchLightning/metrics/pull/824))
 
 
+- Added smart update of `MetricCollection` ([#718](https://github.com/PyTorchLightning/metrics/pull/709))
+
```
Borda, line 21:
```suggestion


```

docs/source/pages/overview.rst
```diff
@@ -292,6 +292,14 @@ inside your LightningModule
     have the same call signature. If this is not the case, input that should be
     given to different metrics can given as keyword arguments to the collection.
 
+An additional advantage of using the ``MetricCollection`` object is that it will
+automatically try to reduce the computations needed by finding groups of metrics
+that share the same underlying metric state. If such metrics are found only one
```
Borda, line 297:
```suggestion
that share the same underlying metric state. If such a group of metrics is found only one
```
can we also say that only the first os updated to reduce number one :)

docs/source/pages/overview.rst
```diff
@@ -292,6 +292,14 @@ inside your LightningModule
...
+of them are actually updated and the updated state will be broadcastet to the rest
+of the metrics. In the example above, this will lead to a 2x-3x lower computational
```
Borda, line 299:
```suggestion
of the metrics within the group. In the example above, this will lead to a 2x-3x lower computational
```

docs/source/pages/overview.rst
```diff
@@ -292,6 +292,14 @@ inside your LightningModule
...
+cost compared to disabling this feature. See the `compute_groups` argument
```
Borda:
```suggestion
cost compared to disabling this feature. See the ``compute_groups`` argument
```

docs/source/pages/overview.rst
Borda: > high cost upfront

how much high, this sounds like rather skip it...

Borda: I see so the overhead is fixed, so it may not pay off for very small scale training of about 100 batches but widely pays of for mid and large scale trainings

Borda: I think that this is the best place we want to show a graph which we can under as matplotlib figure
see: https://github.com/PyTorchLightning/pytorch-lightning/issues/5630

docs/source/pages/overview.rst
```diff
@@ -292,6 +292,14 @@ inside your LightningModule
...
+for more info.
+
```
Borda, line 303: maybe add some examples of such groups, so for classification, it is quite natural, but it may be hard for audio...

tests/bases/test_collections.py
```diff
@@ -278,3 +277,87 @@ def compute(self):
...
+    target = torch.randint(3, (10,))
+    # compute groups should kick in here
+    m.update(preds, target)
+    m2.update(preds, target)
+
+    # compare results for correctness
+    res_cg = m.compute()
+    res_without_cg = m2.compute()
+    for key in res_cg.keys():
+        assert torch.allclose(res_cg[key], res_without_cg[key])
```
Borda, line 339: shall we include some time measure to be sure that the update does not explode...
so for example one test with 3x the same classic metric and reasonable nb batches to be sure that without > with

torchmetrics/collections.py
```diff
@@ -45,6 +44,14 @@ class name as key for the output dict.
 
         postfix: a string to append after the keys of the output dict
 
+        compute_groups:
+            By default the MetricCollection will try to reduce the computations needed for the metrics in the collection
+            by checking if they belong to the same *compute group*. All metrics in a compute group share the same metric
```
Borda, line 49:
```suggestion
            by checking if they belong to the same **compute group**. All metrics in a compute group share the same metric
```

torchmetrics/collections.py
```diff
@@ -45,6 +44,14 @@ class name as key for the output dict.
...
+            state and are therefore only different in their compute step e.g. accuracy, precision and recall can all be
+            computed from the true positives/negatives and false positives/nagatives. By default this argument is `True`
```
Borda, line 51:
```suggestion
            computed from the true positives/negatives and false positives/negatives. By default, this argument is ``True``
```

justusschock:
Seems sphinx doesn't know `:param:`, should we try `:attr:` instead? Param may only work when being within this function...

SkafteNicki:
@justusschock I changed the text to not include a hyperlink but instead just reference the class docs that are just below the sentence

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added new image metric `UniversalImageQualityIndex` ([#824](https://github.com/PyTorchLightning/metrics/pull/824))
 
 
+- Added smart update of `MetricCollection` ([#718](https://github.com/PyTorchLightning/metrics/pull/709))
```
Borda, line 20:
```suggestion
- Added smart update of `MetricCollection` ([#709](https://github.com/PyTorchLightning/metrics/pull/709))
```

docs/source/pages/overview.rst
```diff
@@ -300,6 +300,17 @@ inside your LightningModule
     have the same call signature. If this is not the case, input that should be
     given to different metrics can given as keyword arguments to the collection.
 
+An additional advantage of using the ``MetricCollection`` object is that it will
+automatically try to reduce the computations needed by finding groups of metrics
+that share the same underlying metric state. If such a group of metrics is found only one
+of them is actually updated and the updated state will be broadcasted to the rest
+of the metrics within the group. In the example above, this will lead to a 2x-3x lower computational
+cost compared to disabling this feature. However, this speedup comes with a high cost upfront, where
```
Borda, line 308:
```suggestion
cost compared to disabling this feature. However, this speedup comes with a fixed cost upfront, where
```
the high sounds scary...

docs/source/pages/overview.rst
```diff
@@ -300,6 +300,17 @@ inside your LightningModule
...
+the state-groups have to be determined after the first update. This cost can be significant for very
```
Borda, line 309:
```suggestion
the state-groups have to be determined after the first update. This overhead can be significantly higher then gains speed-up for very
```

docs/source/pages/overview.rst
```diff
@@ -300,6 +300,17 @@ inside your LightningModule
...
+low steps (approx. up to 100) but still leads to an overall speedup for everything beyond that.
```
Borda, line 310:
```suggestion
a low number of steps (approx. up to 100) but still leads to an overall speedup for everything beyond that.
```

docs/source/pages/overview.rst
```diff
@@ -300,6 +300,17 @@ inside your LightningModule
...
+In case the groups are known beforehand, these can also be set manually to avoid the cost of the
```
Borda, line 311:
```suggestion
In case the groups are known beforehand, these can also be set manually to avoid this extra cost of the
```

Borda (approved):

tchaton:
@SkafteNicki Beautiful addition !

PR #717, opened 2022-01-05
Removed input_order from text unit tests

jscottcronin:
#### Historical Work:
The following issue was created to align input order for `preds` and `targets` in the text (NLG) module: https://github.com/PyTorchLightning/metrics/issues/686
The following PR was merged with the fix: https://github.com/PyTorchLightning/metrics/pull/696

Fixes #704 
The work in this PR removes references to `input_order` in the `torchmetrics.tests.text.helpers._class_test` and `torchmetrics.tests.text.helpers._functional_test`. All unit tests also remove this parameter from calls to these methods. All unit tests are passing locally.

## Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure to update the docs?
- [x] Did you write any new necessary tests?

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #725, opened 2022-01-06
Solve pesq test failures

quancs:
## What does this PR do?

see disscussions [here](https://github.com/PyTorchLightning/metrics/pull/717#issuecomment-1006069751)

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/.github/workflows/ci_test-conda.yml
+++ b/.github/workflows/ci_test-conda.yml
@@ -85,8 +85,8 @@ jobs:
         pip --version
         python ./requirements/adjust-versions.py requirements.txt
         python ./.github/prune-packages.py requirements/image.txt torchvision
-        pip install --requirement requirements.txt 
-        pip install --requirement requirements/test.txt 
+        pip install --requirement requirements.txt
+        pip install --requirement requirements/test.txt
         echo "pip list"
         pip list
         echo "conda list"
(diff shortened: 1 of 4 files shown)


Borda:
What I found they mostly recommend reinstall NumPy and also it mostly connected with legacy `pycocotools`

quancs:
> What I found they mostly recommend reinstall NumPy and also it mostly connected with legacy `pycocotools`

I think both is OK. Numpy installed twice during the environment setup. While during the twice installation, pesq is compiled on the first numpy version (newer), so it cannot be loaded after installing the second numpy. Reinstalling pesq can make pesq get compiled on the later installed numpy

quancs:
@Borda Can this be merged? ^^

Borda:
> > What I found they mostly recommend reinstall NumPy and also it mostly connected with legacy `pycocotools`
> 
> I think both is OK. Numpy installed twice during the environment setup. While during the twice installation, pesq is compiled on the first numpy version (newer), so it cannot be loaded after installing the second numpy. Reinstalling pesq can make pesq get compiled on the later installed numpy

would it be the better way to find the minimal NumPy version feasible for `pesq` and `pycocotools` so there is no need to reinstall it and both then will be using the same NumPy package?

Borda:
Ultimately the best would be implementing PESQ in-house so then `pesq` and `pycocotools` are installed only for tests and so in the same time...

Borda (approved):
switched to the fixed NumPy version
(turned out we had the same problem in past with PT 1.4 so setting the freeze to all PT versions)

PR #761, opened 2022-01-16
adding detection.iou

### LINKED ISSUE #659: Add IoU/GIoU for boxes
### Motivation

IoU/GIoU are ametrics essential for the detection domain.

### Pitch

Add metrics, for fast implementation these could be based on torchvisions functions.

jscottcronin:
## What does this PR do?
Adds in IOU and GIOU metrics for object detection.

Fixes #659 

I took a first attempt at this PR and there are a few issues I'm not sure how to address:
1) Naming: there is already a function `iou` in functional/classification/iou.py causing name collisions. We could go with `box_iou` to match torchvision, but its not my favorite name.
2) Currently, we do not check that the box label for each prediction matches the ground truth label of the overlapping box. If the labels do not match, then IoU should be 0, even if IoU > threshold. Not sure how to best handle this within our framework
3) We should not have a pred box match two different ground truth boxes. With N pred boxes and M ground truth boxes, the IoUs calc results in a NxM matrix and each row should have, AT MOST, 1 value > threshold. Do we want to check for this case?
4) The result is a matrix NxM of IoU values. But this is not great as a metric. Should we sum this value, or take the mean? Currently I'm flattening all IOU values > threshold and concatenating before compute.
5) This calculation cannot be run on a batch of data -> it is run independently for each image in the batch. Is this an issue?
6) During testing, the `len(preds) != len(target)` because the number of predictions may not necessarily match the number of ground truth boxes. However, this causes an error in the first line of our `_class_test` function: https://github.com/PyTorchLightning/metrics/blob/master/tests/helpers/testers.py#L147

## Before submitting

## PR review

## Did you have fun?


Borda:
> * Naming: there is already a function `iou` in functional/classification/iou.py causing name collisions. We could go with `box_iou` to match torchvision, but its not my favorite name.

the module name is deprecated and will be removed
https://github.com/PyTorchLightning/metrics/blob/b1d0a7ac3fad7131e37027ae074173461b143c2f/torchmetrics/classification/iou.py#L27-L28
as a module name, I would stick with `box_iou.py` and then the functional `intersection_over_union` or `box_intersection_over_union` and analogically for the Class name

> * Currently, we do not check that the box label for each prediction matches the ground truth label of the overlapping box. If the labels do not match, then IoU should be 0, even if IoU > threshold. Not sure how to best handle this within our framework

yes (if the labels do not match, then IoU should be 0, even if IoU > threshold)

> * We should not have a pred box match two different ground truth boxes. With N pred boxes and M ground truth boxes, the IoUs calc results in a NxM matrix and each row should have, AT MOST, 1 value > threshold. Do we want to check for this case?

if I get your question correctly, then we shall have some max overlap matching first (if there are 2 predictions and 2 true objects and all have some overlap it shall be recognized as two pairs)

> * The result is a matrix NxM of IoU values. But this is not great as a metric. Should we sum this value, or take the mean? Currently I'm flattening all IoU values > threshold and concatenating before compute.

not sure if I understand here...

> * This calculation cannot be run on a batch of data -> it is run independently for each image in the batch. Is this an issue?

ok

> * During testing, the `len(preds) != len(target)` because the number of predictions may not necessarily match the number of ground truth boxes. However, this causes an error in the first line of our `_class_test` function: https://github.com/PyTorchLightning/metrics/blob/master/tests/helpers/testers.py#L147

this van be overcome with your questions above that `preds` and `target` will be list per image, for example:
```py
preds: list[list] = [<list-bboxes-for-the-first-image>, <list-bboxes-for-the-second-image>, ...]
target: list[list] = [<list-bboxes-for-the-first-image>, <list-bboxes-for-the-second-image>, ...]
```

cc: @SkafteNicki @twsl @tkupek

tkupek:
> If the labels do not match, then IoU should be 0

So far, I only know IOU as a helper function to compare a set of boxes. Labels are usually handled separately. But for a metric, it makes sense to take care of the label correctness. Any source that this is the standard approach to check for labels?

In mAP we treat these cases as false-negative.

> len(preds) != len(target)

This checks the number of samples, not the number of boxes.

Borda:
@jscottcronin @tkupek how is it going here. still just a draft? :rabbit:

jscottcronin:
> @jscottcronin @tkupek how is it going here. still just a draft? 🐰

@Borda - It is and I haven't made much progress. I do have a good understanding of the ticket, but I currently do not have good internet, and unlikely to finish it in the next week. If it needs to be brought in sooner, please go ahead and take it. If it can wait, I will be able to get to it, just not sure when.

Borda:
@jscottcronin @tkupek @SkafteNicki how is it going here? :rabbit: 
we are considering a new feature release next week, do you think we can finish this addition by then so it will be included? :pray:

jscottcronin:
> @jscottcronin @tkupek @SkafteNicki how is it going here? 🐰
> we are considering a new feature release next week, do you think we can finish this addition by then so it will be included? 🙏

Hi @Borda - I am unable to continue this work for the foreseeable future. If there are other folks who'd like to take a shot, please give them the opportunity! Thank you!

twsl:
I'd like to finish this implementation

SkafteNicki:
> I'd like to finish this implementation

@twsl feel free to do so :]

SkafteNicki:
Closing as the work is being continued in #1284.

PR #785, opened 2022-01-20
Clarify docs on reset method 

### LINKED ISSUE #766: Metric Cache seems to get inconsistent when state variable is a list
## 🐛 Bug

I debugged into this for a longer time now. I am not 100% sure if it is a bug, but i have no idea what i am doing wrong.

I have attached my custom metric to the `Trainer` and call `trainer.validate(model, datamodule)` to compute the metric on the whole Validation set. In each step i print my metric in the `.compute()` function. 

What confused me a lot is, that the `val_mymetric` value is always the same as the metric value of the last batch. Heres my output (only two batches):

```python
mymetric after 1. batch: 0.109375
mymetric after 2. batch: 0.20454546809196472
# end output of .validate():
{'val_cross_entropy': 0.2601596713066101,
 'val_mymetric': 0.20454546809196472}
```

I also found that in the last computation the metric only holds the values from the last batch in `preds`

### My Metric (only important functions)

```python
class MyMetric(Metric, metaclass=abc.ABCMeta):
    def __init__(self):
        super(MyMetric, self).__init__()
        self.softmax = nn.Softmax(dim=1)
        self.add_state("preds", list())
        self.add_state("labels", list())

    def reset(self):
        self.preds.clear()
        self.labels.clear()

    def update(self, preds: torch.Tensor, target: torch.Tensor):
        self.preds.append(self.softmax(preds))
        self.labels.append(target)
```

IMHO the problem is the following:

```python
# metric.py

# save context before switch
```

The cache is taking the exact same object that get's cleared by the `.reset()` function. I think the cache should be built with a copy of each object? The `id()` of the cache object and metric object was the same.

### To Reproduce

as the metric is confidential i am not able to add a bigger code snippet here.

### Expected behavior

At the end of the epoch, the Cache-Content should hold all elements that come in by `update()` and not be cleared by the `.reset()`

### Environment

PyTorch Version (e.g., 1.0): 1.10.0
Flash Version: 0.6.0
OS (e.g., Linux): Ubuntu on Docker
How you installed PyTorch (conda, pip, source): pip with virtualenv
Build command you used (if compiling from source):
Python version: 3.8.12
CUDA/cuDNN version: 11.4
torchmetrics: 0.6.2

SkafteNicki:
## What does this PR do?

Fixes #766
Make documentation more clear that ``reset`` should not be implemented by user.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/docs/source/pages/implement.rst
+++ b/docs/source/pages/implement.rst
@@ -13,7 +13,7 @@ To implement your own custom metric, subclass the base :class:`~torchmetrics.Met
 We provide the remaining interface, such as ``reset()`` that will make sure to correctly reset all metric
 states that have been added using ``add_state``. You should therefore not implement ``reset()`` yourself.
 Additionally, adding metric states with ``add_state`` will make sure that states are correctly synchronized
-in distributed settings (DDP). To see how metric states are synchronized across distributed processes, 
+in distributed settings (DDP). To see how metric states are synchronized across distributed processes,
 refer to ``add_state()`` docs from the base ``Metric`` class.
 
 Example implementation:
--- a/docs/source/pages/implement.rst
+++ b/docs/source/pages/implement.rst
@@ -10,11 +10,11 @@ To implement your own custom metric, subclass the base :class:`~torchmetrics.Met
 - ``update()``: Any code needed to update the state given any inputs to the metric.
 - ``compute()``: Computes a final value from the state of the metric.
 
-All you need to do is call ``add_state`` correctly to implement a custom metric with DDP.
-``reset()`` is called on metric state variables added using ``add_state()``.
-
-To see how metric states are synchronized across distributed processes, refer to ``add_state()`` docs
-from the base ``Metric`` class.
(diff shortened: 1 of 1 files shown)


docs/source/pages/implement.rst
```diff
@@ -10,11 +10,11 @@ To implement your own custom metric, subclass the base :class:`~torchmetrics.Met
 - ``update()``: Any code needed to update the state given any inputs to the metric.
 - ``compute()``: Computes a final value from the state of the metric.
 
-All you need to do is call ``add_state`` correctly to implement a custom metric with DDP.
-``reset()`` is called on metric state variables added using ``add_state()``.
-
-To see how metric states are synchronized across distributed processes, refer to ``add_state()`` docs
-from the base ``Metric`` class.
+We provide the remaining interface, such as ``reset()`` that will make sure to correctly reset all metric
```
Borda, line 13: I would rather move reset to the list above as it is public API and with the updated text you say it is quite an important method...

Borda (approved):

justusschock (approved):

PR #792, opened 2022-01-23
Deprecate/compute on step

### LINKED ISSUE #789: Deprecate `compute_on_step`
## 🚀 Feature

Deprecate `compute_on_step` argument in all modular metrics.

### Motivation

Setting `compute_on_step` argument to `False` essentially turn `forward` into `update`, since this is the only code that is then being executed:
https://github.com/PyTorchLightning/metrics/blob/5b6cb1d0e735544e6c42cc39bbe7eb29561b75e1/torchmetrics/metric.py#L204-L205
It is sometimes a good thing to have multiple ways of doing certain operations, however in this case IMO it will be much more clear to the user that `update` always do not return anything and `forward` always returns something.

### Pitch

The flag have been there since the beginning, but we as we are now starting the process of refactoring it is worth considering if it should still be part of the framework when it just adds a duplicate of a feature we already have.

SkafteNicki:
## What does this PR do?

Fixes #789
Deprecates `compute_on_step` argument.

## Before submitting

## PR review

## Did you have fun?


justusschock (approved):

Borda (approved):

Borda:
@SkafteNicki mind checking the last two failing tests?

Borda:
```
FAILED tests/bases/test_composition.py::test_compositional_metrics_forward[4-False]
FAILED tests/bases/test_composition.py::test_compositional_metrics_forward[metric_b1-False]
```

Alec-Stashevsky:
@SkafteNicki why was `compute_on_step` deprecated. I use torchmetrics with base PyTorch (non Lightning) and found that  having `compute_on_step` == True (the default) drastically affected my results. I needed to keep it off. How would I go about that in version 0.8.0? 

Thanks

fierval:
Agree with @Alec-Stashevsky 's comment. If I am reading the code correctly:

```python
    def forward(self, *args: Any, **kwargs: Any) -> Any:
        """Automatically calls ``update()``.

        Returns the metric value over inputs if ``compute_on_step`` is True.
        """
        # add current step
        if self._is_synced:
            raise TorchMetricsUserError(
                "The Metric shouldn't be synced when performing ``update``. "
                "HINT: Did you forget to call ``unsync`` ?."
            )

        # global accumulation
        self.update(*args, **kwargs)

        if self.compute_on_step:
            self._to_sync = self.dist_sync_on_step
            # skip restore cache operation from compute as cache is stored below.
            self._should_unsync = False

            # save context before switch

            # call reset, update, compute, on single batch
            self._enable_grad = True  # allow grads for batch computation
            self.reset()
            self.update(*args, **kwargs)
            self._forward_cache = self.compute()

            # restore context
            for attr, val in cache.items():
                setattr(self, attr, val)
            self._is_synced = False

            self._should_unsync = True
            self._to_sync = True
            self._computed = None
            self._enable_grad = False

            return self._forward_cache
```
This used to be the `forward` function in 0.7.3. Now that the `if` statement is removed, does it mean that the metric is finalized & reset on every batch? That's not what I really want.

SkafteNicki:
Hi @Alec-Stashevsky 
If you set `compute_on_step=False` you can directly see in the code that @fierval posted that calling `forward` is the exact same as calling `update` directly. Therefore instead of using a metric like this:
```python
metric = Metric(compute_on_step=False)
metric(input)
```
you should instead just do
```python
metric = Metric()
metric.update(input)
```
@fierval the change in 0.8 just have the effect that `forward` will always accumulate to the global metrics state (the first call to `update`) *AND* also calculate the metric on the current batch. The metric is therefore only finalized on the current batch and not the global state.

Issue #799, opened 2022-01-25
More image metrics


Borda:
## 🚀 Feature

just found this package and seems we have not yet all Image metrics :] 
some inspiration came from https://github.com/andrewekhalel/sewar but we aim on own implementation with `torch`

### Alternatives

- ~Mean Squared Error (MSE)~
- ~Root Mean Sqaured Error (RMSE)~
- ~Peak Signal-to-Noise Ratio (PSNR)~
- ~Structural Similarity Index (SSIM)~
- ~Universal Quality Image Index (UQI)~
- ~Multi-scale Structural Similarity Index (MS-SSIM)~
- ~Erreur Relative Globale Adimensionnelle de Synthèse (ERGAS)~
- ~Spatial Correlation Coefficient (SCC)~
- ~Relative Average Spectral Error (RASE)~
- ~Spectral Angle Mapper (SAM)~
- ~Spectral Distortion Index (D_lambda)~
- ~Spatial Distortion Index (D_S)~
- ~Quality with No Reference (QNR)~
- ~Visual Information Fidelity (VIF)~
- ~Block Sensitive - Peak Signal-to-Noise Ratio (PSNR-B)~

### Additional context

any of these metrics would be a separate PR/addition :]

nishant42491:
i would like to work on this can you assign it to me

ankitaS11:
Hi, @Borda

Thanks for creating the issue tracker. Would love to help in this, as I also look forward to contributing to the lightning framework.

I would like to begin with MSE and RMSE, if that's fine with you and everyone else. Please let me know if I can start working on these two metrics.

Thanks!

SkafteNicki:
@ankitaS11 Please note we already have MSE and RMSE in the regression package:
https://torchmetrics.readthedocs.io/en/latest/references/modules.html#meansquarederror (this takes an `squared` argument)
So I do not think we should re-implement them in the image domain.

Borda:
> i would like to work on this can you assign it to me

cool, which one do you want to take? (we have many :zap:)

ankitaS11:
> @ankitaS11 Please note we already have MSE and RMSE in the regression package: https://torchmetrics.readthedocs.io/en/latest/references/modules.html#meansquarederror (this takes an `squared` argument) So I do not think we should re-implement them in the image domain.

[...]

PR #800, opened 2022-01-26
new metric SCC for Images

nishant42491:
## What does this PR do?

Adds a new metric Spatial Correlation Coefficient for Images

part of #799

## Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)-Yes
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?-Yes
- [ ] Did you make sure to update the docs?-No
- [ ] Did you write any new necessary tests?-No

## PR review

## Did you have fun?


SkafteNicki:
Hi @nishant42491, thanks for wanting to contribute.
However, I can only see that you have created a new file. Is that correct?

nishant42491:
> Hi @nishant42491, thanks for wanting to contribute. However, I can only see that you have created a new file. Is that correct?

yes, that's correct. sorry for the delay as I was trying to understand how to calculate the SCC coefficient as there are not many resources available online to understand the implementation of Laplacian filters. I am working with my college professors to understand more about the SCC metric how ever due to my college exams my progress has been slow. however I will try to speed things up again really sorry for the delay.

stancld:
Hi @nishant42491, do you have any updates here? :]

nishant42491:
> Hi @nishant42491, do you have any updates here? :]

yep, I have taken references from andrewkhalels sewar repo containing information about the SCC metric and have created a scc.py  file containing its implementation however  I am not 100 % sure that it's correct as I still don't understand the SCC metric completely. Ill be more than happy to change anything that needs changing. I would love any suggestions on how to proceed further. Thank you a lot for your patience with me  I really appreciate it :].

nishant42491:
@SkafteNicki @stancld I have finished creating a functional and class-based interface for the scc metric and they seem to be working properly. I have tested it against the sewar repos SCC metric and my metric is giving me the right outputs. However, I am stuck on implementing the test_scc.py file as I don't exactly know what to write in the tests file. I have tried using the sewar repos scc metric in the tests file but it does not work with batched inputs so i am not sure what to test my metric against. any suggestions on how to move forward will really help me out. Thank you in advance for your help :]

SkafteNicki:
Hi @nishant42491,
I tried to take a stab at finishing this PR. I written tests and added some docs, but the tests fails with assertion when I try to compare against the implementation from `sewar`. Could you maybe help with getting this PR finished, by providing at least an example how to get this implementation to match the one from `sewar`.

nishant42491:
> Hi @nishant42491, I tried to take a stab at finishing this PR. I written tests and added some docs, but the tests fails with assertion when I try to compare against the implementation from `sewar`. Could you maybe help with getting this PR finished, by providing at least an example how to get this implementation to match the one from `sewar`.



> Hi @nishant42491, I tried to take a stab at finishing this PR. I written tests and added some docs, but the tests fails with assertion when I try to compare against the implementation from `sewar`. Could you maybe help with getting this PR finished, by providing at least an example how to get this implementation to match the one from `sewar`.

I think the difference between the two metrics arises because the package in swear has implemented a high pass filter whilst I have not however the difference did not seem significant whilst i was testing the metric

nishant42491:
@SkafteNicki  should I try implementing a high pass filter for my metric too to try and make it pass the tests you have written?

SkafteNicki:
@nishant42491 yes, we kind of need the two metrics to produce the same output, so we make sure that this implementation is doing the right thing.
Maybe the reason the tests are failing is that the input is random? did you test with some structured input?

nishant42491:
image
image
image
image
puppy
mud_turtle

@SkafteNicki I have received outputs 0.079 and 0.080 for 2 images for swear's metric and my metric respectively

nishant42491:
will try and implement the high pass filter which should solve the problem. @SkafteNicki  Thnx for the help on the tests  i really appreciate it :]

SkafteNicki:
@nishant42491 is it correct that current differences in implementation is:
* The input to `sewar` should be `target`, `preds` whereas your implementation takes in oppesit order `preds`, `target` (which is the same as the rest of out codebase)
* The input to `sewar` is expected to be single images of shape `[H, W, C]` whereas your implementation takes in `[B, C, H, W]` (so the channel dimension should be moved when we compare)

Just trying to figure out why I cannot get the numbers to match (I know the high pass filter is missing, but it should still be fairly close as to my understanding).

nishant42491:
> @nishant42491 is it correct that current differences in implementation is:
> 
> * The input to `sewar` should be `target`, `preds` whereas your implementation takes in oppesit order `preds`, `target` (which is the same as the rest of out codebase)
> * The input to `sewar` is expected to be single images of shape `[H, W, C]` whereas your implementation takes in `[B, C, H, W]` (so the channel dimension should be moved when we compare)
> 
> Just trying to figure out why I cannot get the numbers to match (I know the high pass filter is missing, but it should still be fairly close as to my understanding).

yep, those are the differences between the sewar implementation and the metric's implementation. for the inputs, I have tested 
In both metrics  the difference is slight generally up to the 3rd decimal place.

nishant42491:
> > @nishant42491 is it correct that current differences in implementation is:
> > 
> > * The input to `sewar` should be `target`, `preds` whereas your implementation takes in oppesit order `preds`, `target` (which is the same as the rest of out codebase)
> > * The input to `sewar` is expected to be single images of shape `[H, W, C]` whereas your implementation takes in `[B, C, H, W]` (so the channel dimension should be moved when we compare)
> > 
> > Just trying to figure out why I cannot get the numbers to match (I know the high pass filter is missing, but it should still be fairly close as to my understanding).
> 
> yep, those are the differences between the sewar implementation and the metric's implementation. for the inputs, I have tested In both metrics the difference is slight generally up to the 3rd decimal place.

one more thing is that my default kernel size is 9, whilst sewras kernel size is 8 so you have to set the sewar metrics kernel size to 9 explicitly   to get similar results

nishant42491:
> @nishant42491 tried changing the implementation based on what you have told me but cannot get it to match. Below is shown the wrapped reference implementation from `sewar` that should do everything correctly:
> 
> * summarizes a batch of input
> * flips the input
> * permute the dimensions
> * changes the window size to 9
> 
> ```python
> from sewar.full_ref import scc
> 
> def _reference_scc(preds, target, reduction):
>     val = 0.0
>     for p, t in zip(preds, target):
>         val += scc(t.permute(1, 2, 0).numpy(), p.permute(1, 2, 0).numpy(), ws=9)
>     val = val if reduction == "sum" else val / preds.shape[0]
>     return val
> 
> from torchmetrics.functional import spatial_correlation_coefficient
> 
> import torch
> _ = torch.manual_seed(42)
> 
> BATCH_SIZE = 10
> CHANNELS = 3
> SIZE = 100
> preds = torch.randint(0, 255, (BATCH_SIZE, CHANNELS, SIZE, SIZE)).float()
> target = torch.randint(0, 255, (BATCH_SIZE, CHANNELS, SIZE, SIZE)).float()
> 
> print(spatial_correlation_coefficient(preds, target, reduction='sum').item())
> print(_reference_scc(preds, target, reduction='sum'))
> ```
> 
> can you find what I do wrong?

Your Implementation seems correct. I'm not quite sure why the outputs differ.

PR #818, opened 2022-01-31
3D extension for SSIM

### LINKED ISSUE #812: 3D SSIM
The metrics `StructuralSimilarityIndexMeasure` and `MultiScaleStructuralSimilarityIndexMeasure` current only support 2D image data. A 3D version would be nice to have, and an extension to allow 3D data should be straightforward to implement. If you are interested in such an extension, I could take over the implementation.

weningerleon:
## What does this PR do?

Fixes #812
Changes StructuralSimilarityIndexMeasure to be able to handle 3D images. Deciding whether the 2D or 3D version is used is determined automatically based on kernel_size. Basic sanity checks are carried out. Produces sensible results on my dataset.

## Before submitting

## PR review

## Did you have fun?


torchmetrics/functional/image/ssim.py
```diff
@@ -85,9 +106,14 @@ def _ssim_update(preds: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
...
-    if len(preds.shape) != 4:
+    if kernel_dimension == 2 and len(preds.shape) != 4:
+        raise ValueError(
+            "2D kernel is used. Expected `preds` and `target` to have BxCxHxW shape."
+            f" Got preds: {preds.shape} and target: {target.shape}."
+        )
+    if kernel_dimension == 3 and len(preds.shape) != 5:
         raise ValueError(
-            "Expected `preds` and `target` to have BxCxHxW shape."
+            "3D kernel is used. Expected `preds` and `target` to have BxCxHxWxD shape."
```
justusschock, line 116:
```suggestion
            "3D kernel is used. Expected `preds` and `target` to have BxCxDxHxW shape."
```

AFAIK this should be standard for 3d. Maybe reflect this in the variable names for the 3d kernel generation as well?

torchmetrics/functional/image/ssim.py
```diff
@@ -149,15 +180,27 @@ def _ssim_compute(
 
     channel = preds.size(1)
     dtype = preds.dtype
-    kernel = _gaussian_kernel(channel, kernel_size, sigma, dtype, device)
     pad_h = (kernel_size[0] - 1) // 2
     pad_w = (kernel_size[1] - 1) // 2
 
-    preds = F.pad(preds, (pad_h, pad_h, pad_w, pad_w), mode="reflect")
-    target = F.pad(target, (pad_h, pad_h, pad_w, pad_w), mode="reflect")
+    if is3D:
```
justusschock, line 186:
```suggestion
    if is_3d:
```

torchmetrics/functional/image/ssim.py
```diff
@@ -149,15 +180,27 @@ def _ssim_compute(
...
+
+    else:
+        preds = F.pad(preds, (pad_h, pad_h, pad_w, pad_w), mode="reflect")
+        target = F.pad(target, (pad_h, pad_h, pad_w, pad_w), mode="reflect")
+        kernel = _2d_gaussian_kernel(channel, kernel_size, sigma, dtype, device)
 
     input_list = torch.cat((preds, target, preds * preds, target * target, preds * target))  # (5 * B, C, H, W)
-    outputs = F.conv2d(input_list, kernel, groups=channel)
+
+    if is3D:
```
justusschock, line 199:
```suggestion
    if is_3d:
```

justusschock (commented):
Some first comments.

Usually there are 2 ways for testing:

1.) Use random data and compare the results with a gold-standard implementation (like sklearn)

2.) Use predefined-data, calculate result manually once and hardcode those results as targets.

Ideally we would prefer Option 1 since that way you can always test it with additional data if required. I think the sklearn implementation [here](https://scikit-image.org/docs/stable/api/skimage.metrics.html?highlight=ssim#structural-similarity) should also work with 3D data :)

Edit: have a look at the 2d tests [here](https://github.com/PyTorchLightning/metrics/blob/master/tests/image/test_ssim.py)

torchmetrics/functional/image/ssim.py
```diff
@@ -274,7 +274,7 @@ def structural_similarity_index_measure(
         >>> structural_similarity_index_measure(preds, target)
         tensor(0.9219)
     """
-    preds, target = _ssim_update(preds, target)
```
justusschock, line 234: I wonder if we shouldn't default to a single value (11) here and expand that to the necessary dimensions based on `preds.ndim` or `target.ndim`. Would be cleaner for 2d and 3d then :) Especially if this is a default variable

torchmetrics/functional/image/ssim.py
```diff
@@ -70,7 +70,28 @@ def _gaussian_kernel(
...
+        sigma: Standard deviation of the gaussian kernel
+        dtype: data type of the output tensor
+        device: device of the output tensor
+    """
+
+    gaussian_kernel_x = _gaussian(kernel_size[0], sigma[0], dtype, device)
+    gaussian_kernel_y = _gaussian(kernel_size[1], sigma[1], dtype, device)
+    gaussian_kernel_z = _gaussian(kernel_size[2], sigma[2], dtype, device)
+    kernel_xy = torch.matmul(gaussian_kernel_x.t(), gaussian_kernel_z)  # (kernel_size, 1) * (1, kernel_size)
+    kernel = torch.mul(kernel_xy, gaussian_kernel_z.expand(11, 11, 11))
```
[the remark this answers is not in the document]
weningerleon: That was a mistake, thanks for the catch

torchmetrics/functional/image/ssim.py
weningerleon: That would be possible. A single StructuralSimilarityIndexMeasure object could then even handle 2d as well as 3 data.

justusschock: However, we would still need to allow the user to provide the sequence manually for anisotropic kernels I think :)

weningerleon:
I think your current tests that compare the results from torchmetrics with the results from skimage is faulty.
Have a look at the 2d tests mentioned by @justusschock, function `_sk_ssim`: 
You permute the 4D tensor (BxCxHxW) to BxHxWxC, with B=batch_size, C=image channels. Skimage treats the BxHxWxC tensor as a 3D image with C channels. However, it needs to be treated as _B_ 2D images (skimage does not support batching). This also explains why you need to set atol to 6e-3, which is really high.

torchmetrics/functional/image/ssim.py
weningerleon: Sure, that's still possible

justusschock:
@weningerleon If you could fix those while working on it, this would also be highly appreciated :D

weningerleon:
> @weningerleon If you could fix those while working on it, this would also be highly appreciated :D

Yes, have a look at the updates files ;)

torchmetrics/functional/image/ssim.py
```diff
@@ -77,14 +77,30 @@ def _ssim_compute(
     Example:
         >>> preds = torch.rand([16, 1, 16, 16])
         >>> target = preds * 0.75
-        >>> preds, target = _ssim_update(preds, target)
+        >>> preds, target = _ssim_update(preds, target, 2)
```
SkafteNicki, line 80:
```suggestion
        >>> preds, target = _ssim_update(preds, target)
```

torchmetrics/functional/image/ssim.py
```diff
@@ -77,14 +77,30 @@ def _ssim_compute(
...
         >>> _ssim_compute(preds, target)
         tensor(0.9219)
     """
-    if len(kernel_size) != 2 or len(sigma) != 2:
+    is_3d = len(preds.shape) == 5
+
+    if not isinstance(kernel_size, Sequence):
+        kernel_size = [kernel_size, kernel_size, kernel_size] if is_3d else [kernel_size, kernel_size]
```
SkafteNicki, line 87:
```suggestion
        kernel_size = 3*[kernel_size] if is_3d else 2*[kernel_size]
```

torchmetrics/functional/image/ssim.py
```diff
@@ -77,14 +77,30 @@ def _ssim_compute(
...
+    if not isinstance(sigma, Sequence):
+        sigma = [sigma, sigma, sigma] if is_3d else [sigma, sigma]
```
SkafteNicki, line 89:
```suggestion
        sigma = 3*[sigma] if is_3d else 2*[sigma]
```

justusschock (approved):

Borda (approved):

Borda:
@weningerleon mind chek, all tests are failing with:
```
ValueError: Expected `kernel_size` and `sigma` to have the same length. Kernel_size dimensionality: 3, sigma dimensionality: 2.
```

stancld:
> @weningerleon mind chek, all tests are failing with:
> 
> ```
> ValueError: Expected `kernel_size` and `sigma` to have the same length. Kernel_size dimensionality: 3, sigma dimensionality: 2.
> ```
> 
> or @stancld mind helping here?

I'd suggest the following solution:

```diff
def structural_similarity_index_measure(
    preds: Tensor,
    target: Tensor,
-   kernel_size: Sequence[int] = (11, 11),
+   kernel_size: Union[int, Sequence[int]] = 11,
-   sigma: Sequence[float] = (1.5, 1.5),
+   sigma: Union[float, Sequence[float]] = 1.5,
    reduction: str = "elementwise_mean",
    data_range: Optional[float] = None,
    k1: float = 0.01,
    k2: float = 0.03,
) -> Tensor:
```

and

```diff
def multiscale_structural_similarity_index_measure(
    preds: Tensor,
    target: Tensor,
-   kernel_size: Sequence[int] = (11, 11),
+   kernel_size: Union[int, Sequence[int]] = 11,
-   sigma: Sequence[float] = (1.5, 1.5),
+   sigma: Union[float, Sequence[float]] = 1.5,
    reduction: str = "elementwise_mean",
    data_range: Optional[float] = None,
    k1: float = 0.01,
    k2: float = 0.03,
    betas: Tuple[float, ...] = (0.0448, 0.2856, 0.3001, 0.2363, 0.1333),
    normalize: Optional[Literal["relu", "simple"]] = None,
) -> Tensor:
```

to match the logic used in module metrics. This also should fix the issue.

torchmetrics/image/ssim.py
```diff
@@ -57,8 +57,8 @@ class StructuralSimilarityIndexMeasure(Metric):
 
     def __init__(
         self,
-        kernel_size: Sequence[int] = (11, 11),
-        sigma: Sequence[float] = (1.5, 1.5),
+        kernel_size: Union[int, Sequence[int]] = 11,
+        sigma: Union[float, Sequence[float]] = 1.5,
```
stancld, line 61: I'd update types and default values for `kernel_size` and `sigma` accordingly for functional metrics as well.

weningerleon: Thanks for the update! Sorry I missed this here...

weningerleon: I updated the same thing in the functional file

stancld:
> Please note: 3D reflection padding was only introduced in torch 1.10. As the current implementation uses this function, SSIM will raise a `NotImplementedError` if used in the 3D version for any torch version older than 1.10. How should we deal with this?

I'll try to think about some possible solutions. 🤔

stancld:
> Please note: 3D reflection padding was only introduced in torch 1.10. As the current implementation uses this function, SSIM will raise a `NotImplementedError` if used in the 3D version for any torch version older than 1.10. How should we deal with this?

It should be doable to implement 3D reflection padding for `torch<1.10` on our own. I'll send a proposal for that :]

stancld:
> Please note: 3D reflection padding was only introduced in torch 1.10. As the current implementation uses this function, SSIM will raise a `NotImplementedError` if used in the 3D version for any torch version older than 1.10. How should we deal with this?

@weningerleon Regarding the limited availability of the 3D reflection padding (only for `torch>=1.10`), we can define our own 3D reflection padding using the following logic:

```python
def _single_dimension_pad(inputs: Tensor, dim: int, pad: int) -> Tensor:
    _max = inputs.shape[dim] - 2
    x = torch.index_select(inputs, dim, torch.tensor(range(pad, 0, -1)))
    y = torch.index_select(inputs, dim, torch.tensor(range(_max, _max - pad, -1)))
    return torch.cat((x, inputs, y), dim)

def _reflection_pad_3d(inputs: Tensor, pad_h: int, pad_w: int, pad_d: int) -> Tensor:
    for dim, pad in enumerate([pad_h, pad_w, pad_d]):
        inputs = _single_dimension_pad(inputs, dim + 2, pad)
    return inputs
```

This function seems to provide us with the equivalent results as
```python
inputs = F.pad(inputs, (pad_h, pad_h, pad_w, pad_w, pad_d, pad_d), mode="reflect")
```
Also, it seems to be even faster than `F.pad`.

cc: @Borda @SkafteNicki (What do you think? :] I'm not so familiar with image-related metrics)

tests/image/test_ssim.py
```diff
@@ -52,16 +60,19 @@ def _sk_ssim(preds, target, data_range, multichannel, kernel_size):
...
+            sk_target[i],
+            sk_preds[i],
+            data_range=data_range,
+            multichannel=multichannel,
+            gaussian_weights=True,
+            win_size=kernel_size,
+            sigma=1.5,
+            use_sample_covariance=False,
+        )
+    return results
```
stancld, line 75: I'm not sure if this implementation doesn't cause some problems. Please, refer to the failing tests @weningerleon

torchmetrics/functional/image/helper.py
```diff
@@ -83,6 +85,13 @@ def _single_dimension_pad(inputs: Tensor, dim: int, pad: int) -> Tensor:
 
 
 def _reflection_pad_3d(inputs: Tensor, pad_h: int, pad_w: int, pad_d: int) -> Tensor:
-    for dim, pad in enumerate([pad_h, pad_w, pad_d]):
-        inputs = _single_dimension_pad(inputs, dim + 2, pad)
+    (torch_v0, torch_v1) = torch.__version__.split(".")[:2]
+    if int(torch_v0) > 1 or int(torch_v1) >= 10:
```
stancld, line 90: Please define rather helper variable
```python
_TORCH_GREATER_EQUAL_1_10: Optional[bool] = _compare_version("torch", operator.ge, "1.10.0")
```
in `torchmetrics/utilities/imports.py`

and then here check
```
if _TORCH_GREATER_EQUAL_1_10:
    ....
else:
    ....
```
to stay consistent with the rest of the repo :]

Otherwise good point to use torch implementation if possible, thanks for the fix :]

weningerleon:
> 

@stancld Thanks for the implementation!
I suggest defaulting to the pyTorch implementation if the current (or newer) version is used, and include a deprecation warning:

```python
def _reflection_pad_3d(inputs: Tensor, pad_h: int, pad_w: int, pad_d: int) -> Tensor:
    (torch_v0, torch_v1) = torch.__version__.split(".")[:2]
    if int(torch_v0) > 1 or int(torch_v1) >= 10:
        inputs = F.pad(inputs, (pad_h, pad_h, pad_w, pad_w, pad_d, pad_d), mode="reflect")
    else:
        rank_zero_warn(
            "An older version of pyTorch is used. For optimal speed, please upgrade to at least pyTorch 1.10."
        )
        for dim, pad in enumerate([pad_h, pad_w, pad_d]):
            inputs = _single_dimension_pad(inputs, dim + 2, pad)
    return inputs
```

torchmetrics/functional/image/helper.py
weningerleon: Ok I'll do

tests/image/test_ssim.py
stancld, line 75:
```suggestion
    return structural_similarity(
        sk_target,
        sk_preds,
        data_range=data_range,
        multichannel=multichannel,
        gaussian_weights=True,
        win_size=kernel_size,
        sigma=1.5,
        use_sample_covariance=False,
    )
```

tests/image/test_ssim.py
```diff
@@ -70,8 +81,6 @@ def _sk_ssim(preds, target, data_range, multichannel, kernel_size):
 )
 @pytest.mark.parametrize("kernel_size", [5, 11])
 class TestSSIM(MetricTester):
```
stancld, line 83:
```suggestion
class TestSSIM(MetricTester):
    atol = 6e-3
    
```

stancld (changes requested):
It looks to me tests work fine locally with the old settings

tests/image/test_ssim.py
weningerleon: The problem with the old settings is, they compute the wrong thing, as skimage does not support batching. For example, if a 2D image (BxHxW) is tested with the old setting, skimage interprets it as a 3D image, and does 3D SSIM. The results are thus garbage, this is why atol needed to be set higher (if you average over a high number of random distributed variables, it is no wonder that something similar comes out)

stancld: Yes, you're right. I've had a deeper look at `skimage` package and they don't consider batch as the first dimension. It seems to me that the following should be working out:

```python
def _sk_ssim(preds, target, data_range, multichannel, kernel_size):
    is_3d = len(preds.shape) == 5

    if is_3d:
        sk_preds = preds.permute(0, 2, 3, 4, 1).squeeze(-1).numpy()
        sk_target = target.permute(0, 2, 3, 4, 1).squeeze(-1).numpy()
    else:
        sk_preds = preds.permute(0, 2, 3, 1).squeeze(-1).numpy()
        sk_target = target.permute(0, 2, 3, 1).squeeze(-1).numpy()

    ssim = []
    for i in range(len(sk_preds)):
        ssim.append(
            structural_similarity(
                sk_target[i],
                sk_preds[i],
                data_range=data_range,
                multichannel=multichannel,
                gaussian_weights=True,
                win_size=kernel_size,
                sigma=1.5,
                use_sample_covariance=False,
            )
        )
    return torch.tensor(ssim).mean()
```

weningerleon: Looks better!
There are a few further issues:
1) Our current SSIM implementation reduces over all dimension -including the batch dimension- at the end.
2) The multichannel setting for the skimage test will produce erroneous results.
I'm currently working on these, and will update you with a new version later today

stancld: Great, thank you a lot for all the endavour! I'll just leave it for now as it is. Please, ping me once you have any update! :]

stancld: However, if these points are general both for 2D and 3D case, it might be better to open a separate issue and PR for these :]

weningerleon:
Updates:
- Make ssim behavior closer to the behavior of skimage
- Add uniform kernel to ssim
- Update tests
- Bug fixes in implementation

weningerleon:
@stancld please have a look at the new changes. There are quite a few, if something needs more explanation please tell me

stancld:
> @stancld please have a look at the new changes. There are quite a few, if something needs more explanation please tell me. The tests run without error locally for me, please also have a look at whats happening there...

Hi @weningerleon - I'm not fully sure it's desirable to incorporate all these supposed changes into this single PR that should rather focus on the 3D SSIM. Just maybe let me ping @justusschock / @Borda - any opinion on this? :]

justusschock:
IMO it is fine to have them all in this PR, since they are all focused on SSIM in general (2D and 3D) and also fix the SSIM tests :)

The only thing we should do is to have multiple changelog entries for this: 1 with the 3D thing and one for the fixes :)

weningerleon:
Updated the changelog :)
@stancld, what you perhaps could look at is the averaging over batches. Before, the SSIM was automatically averaged over all images in the batch. I have deactivated that (it's necessary to have access to the per-image results), ssim now outputs a tensor of size n given a batchsize n. However, I'm unsure if such batch-averaging is standard for this package, perhaps you want to add an option etc. for this...

Borda:
@stancld still requesting changes? :rabbit:

stancld:
> @stancld still requesting changes? 🐰

Gonna have a look in the afternoon :] However, a lot of changes have been made since your approvals were given. Maybe the reviews should be re-requested just to be sure everything is ok? :]

torchmetrics/image/ssim.py
```diff
@@ -185,18 +201,24 @@ def __init__(
...
+                "Argument `kernel_size` expected to be an sequence or an int." f"or a single int. Got {kernel_size}"
             )
-        self.kernel_size = kernel_size
+        if isinstance(kernel_size, Sequence) and (
+            len(kernel_size) not in (2, 3) or not all(isinstance(ks, int) for ks in kernel_size)
+        ):
+            raise ValueError(
+                "Argument `kernel_size` expected to be an sequence of size 2 or 3 where each element is an int, "
+                f"or a single int. Got {kernel_size}"
+            )
```
SkafteNicki, line 214: please add a `Raises` section to the docstring similar to:
https://github.com/PyTorchLightning/metrics/blob/5cbc7add8877e6524b140eb35f57a26b53fe5670/torchmetrics/image/fid.py#L182-L188

torchmetrics/image/ssim.py
```diff
@@ -164,9 +180,9 @@ class MultiScaleStructuralSimilarityIndexMeasure(Metric):
 
     def __init__(
         self,
-        kernel_size: Sequence[int] = (11, 11),
-        sigma: Sequence[float] = (1.5, 1.5),
-        reduction: str = "elementwise_mean",
+        gaussian_kernel: bool = True,
+        kernel_size: Union[int, Sequence[int]] = 11,
+        sigma: Union[float, Sequence[float]] = 1.5,
```
SkafteNicki, line 185: can we reorder to `kernel_size`, `sigma`, `gaussian_kernel`?
what happened to the reduction parameter? that need to be deprecated if it is being removed

torchmetrics/image/ssim.py
```diff
@@ -66,13 +71,17 @@ class StructuralSimilarityIndexMeasure(Metric):
...
-        reduction: str = "elementwise_mean",
+        gaussian_kernel: bool = True,
+        sigma: Union[float, Sequence[float]] = 1.5,
+        kernel_size: Union[int, Sequence[int]] = 11,
         data_range: Optional[float] = None,
         k1: float = 0.01,
         k2: float = 0.03,
         compute_on_step: Optional[bool] = None,
+        dist_sync_on_step: bool = False,
+        process_group: Optional[Any] = None,
```
SkafteNicki, line 82: these arguments should not be there anymore. probably a merge that went wrong
```suggestion
```

torchmetrics/image/ssim.py
```diff
@@ -23,20 +23,24 @@
 
 
 class StructuralSimilarityIndexMeasure(Metric):
-    """Computes Structual Similarity Index Measure (SSIM_).
+    """Computes Structual Similarity Index Measure.
```
SkafteNicki, line 26: what happened to the link?
```suggestion
    """Computes Structual Similarity Index Measure (SSIM_).
```

torchmetrics/functional/image/ssim.py
```diff
@@ -124,43 +165,56 @@ def _ssim_compute(
...
-    reduction: str = "elementwise_mean",
+    gaussian_kernel: bool = True,
+    sigma: Union[float, Sequence[float]] = 1.5,
+    kernel_size: Union[int, Sequence[int]] = 11,
     data_range: Optional[float] = None,
     k1: float = 0.01,
     k2: float = 0.03,
+    return_full_image: bool = False,
+    return_contrast_sensitivity: bool = False,
 ) -> Tensor:
```
SkafteNicki, line 199:
```suggestion
) -> Union[Tensor, Tuple[Tensor, Tensor]]:
```

SkafteNicki (changes requested):
I am fine with the removal of the `reduction` argument but we need to deprecate it :]

torchmetrics/functional/image/helper.py
```diff
@@ -35,16 +43,59 @@ def _gaussian_kernel(
...
+    gaussian_kernel_z = _gaussian(kernel_size[2], sigma[2], dtype, device)
+    kernel_xy = torch.matmul(gaussian_kernel_x.t(), gaussian_kernel_y)  # (kernel_size, 1) * (1, kernel_size)
+    kernel = torch.mul(
+        kernel_xy.unsqueeze(-1).repeat(1, 1, kernel_size[2]),
+        gaussian_kernel_z.expand(kernel_size[0], kernel_size[1], kernel_size[2]),
+    )
+    return kernel.expand(channel, 1, kernel_size[0], kernel_size[1], kernel_size[2])
+
+
+def _single_dimension_pad(inputs: Tensor, dim: int, pad: int) -> Tensor:
```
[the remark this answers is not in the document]
weningerleon: ping @stancld

torchmetrics/image/ssim.py
weningerleon: That ordering works for me.

Concerning the reduction parameter: Before, there was no option of no reduction. However, it definitely is necessary to have access to the metric for every image in the batch. 
I wasn't sure what is "standard" in this package for the reduction parameter. If you wish to have this parameter, we can add it again. I also opened this up to discussion somewhere else in this PR.

torchmetrics/functional/image/helper.py
```diff
@@ -35,16 +43,59 @@ def _gaussian_kernel(
...
+    _max = inputs.shape[dim] - 2
+    x = torch.index_select(inputs, dim, torch.arange(pad, 0, -1))
+    y = torch.index_select(inputs, dim, torch.arange(_max, _max - pad, -1))
+    return torch.cat((x, inputs, y), dim)
+
+
+def _reflection_pad_3d(inputs: Tensor, pad_h: int, pad_w: int, pad_d: int) -> Tensor:
```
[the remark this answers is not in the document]
weningerleon: ping @stancld

CHANGELOG.md
```diff
@@ -17,6 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added new image metric `UniversalImageQualityIndex` ([#824](https://github.com/PyTorchLightning/metrics/pull/824))
 
 
+- Added support for 3D image in `StructuralSimilarityIndexMeasure` ([#818](https://github.com/PyTorchLightning/metrics/pull/818))
+
+
+- Added support for uniform kernel in `StructuralSimilarityIndexMeasure` ([#818](https://github.com/PyTorchLightning/metrics/pull/818))
```
Borda, line 23:
```suggestion
- Added support for 3D image and uniform kernel in `StructuralSimilarityIndexMeasure` ([#818](https://github.com/PyTorchLightning/metrics/pull/818))
```

CHANGELOG.md
```diff
@@ -31,6 +37,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- kernel_size is derived from sigma in case of gaussian weights in `StructuralSimilarityIndexMeasure` ([#818](https://github.com/PyTorchLightning/metrics/pull/818))
+
+- Fixed tests for `StructuralSimilarityIndexMeasure` ([#818](https://github.com/PyTorchLightning/metrics/pull/818))
+
```
Borda, line 43:
```suggestion
```

torchmetrics/functional/image/ssim.py
```diff
@@ -49,42 +48,66 @@ def _ssim_update(preds: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
...
-
-            - ``'elementwise_mean'``: takes the mean (default)
-            - ``'sum'``: takes the sum
-            - ``'none'``: no reduction will be applied
-
+        gaussian_kernel: If true (default), a gaussian kernel is used, if false a uniform kernel is used
+        sigma: Standard deviation of the gaussian kernel (default: 1.5), anisotropic kernels are possible.
+            Ignored if a uniform kernel is used
+        kernel_size: size of the uniform kernel (default: 11), anisotropic kernels are possible.
+            Ignored if a gaussian kernel is used
```
Borda, line 69:
```suggestion
        sigma: Standard deviation of the gaussian kernel, anisotropic kernels are possible.
            Ignored if a uniform kernel is used
        kernel_size: the size of the uniform kernel, anisotropic kernels are possible.
            Ignored if a Gaussian kernel is used
```

torchmetrics/functional/image/ssim.py
```diff
@@ -49,42 +48,66 @@ def _ssim_update(preds: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
...
         data_range: Range of the image. If ``None``, it is determined from the image (max - min)
         k1: Parameter of SSIM.
         k2: Parameter of SSIM.
+        return_full_image: If true, the full ssim image is returned as a second arguments.
+            Mutually exlusive with return_contrast_sensitivity
+        return_contrast_sensitivity: If true, the contrast term is returned as a second argument.
+            The luminance term can be obtained with luminance=ssim/contrast
+            Mutually exlusive with return_full_image
```
Borda, line 77:
```suggestion
        return_full_image: If true, the full ``ssim`` image is returned as a second argument.
            Mutually exlusive with ``return_contrast_sensitivity``
        return_contrast_sensitivity: If true, the contrast term is returned as a second argument.
            The luminance term can be obtained with luminance=ssim/contrast
            Mutually exclusive with ``return_full_image``
```

torchmetrics/functional/image/ssim.py
```diff
@@ -49,42 +48,66 @@ def _ssim_update(preds: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
...
+
+    if not isinstance(kernel_size, Sequence):
+        kernel_size = 3 * [kernel_size] if is_3d else 2 * [kernel_size]
+    if not isinstance(sigma, Sequence):
+        sigma = 3 * [sigma] if is_3d else 2 * [sigma]
+
+    if len(kernel_size) != len(target.shape) - 2:
+        raise ValueError(
+            f"`kernel_size` has dimension {len(kernel_size)}, but expected to be two less that target dimensionality, "
+            f"which is: {len(target.shape)}"
```
Borda, line 97:
```suggestion
            f"`kernel_size` has dimension {len(kernel_size)}, but expected to be two less that target dimensionality,"
            f" which is: {len(target.shape)}"
```

torchmetrics/functional/image/ssim.py
```diff
@@ -49,42 +48,66 @@ def _ssim_update(preds: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
...
+    if len(kernel_size) not in (2, 3):
         raise ValueError(
-            "Expected `kernel_size` and `sigma` to have the length of two."
-            f" Got kernel_size: {len(kernel_size)} and sigma: {len(sigma)}."
+            f"Expected `kernel_size` dimension to be 2 or 3. `kernel_size` dimensionality: {len(kernel_size)}"
+        )
+    if len(sigma) != len(target.shape) - 2:
+        raise ValueError(
+            f"`kernel_size` has dimension {len(kernel_size)}, but expected to be two less that target dimensionality, "
+            f"which is: {len(target.shape)}"
```
Borda, line 106:
```suggestion
            f"`kernel_size` has dimension {len(kernel_size)}, but expected to be two less that target dimensionality,"
            f" which is: {len(target.shape)}"
```

torchmetrics/functional/image/ssim.py
```diff
@@ -124,43 +165,56 @@ def _ssim_compute(
...
-
-            - ``'elementwise_mean'``: takes the mean (default)
-            - ``'sum'``: takes the sum
-            - ``'none'``: no reduction will be applied
-
+        gaussian_kernel: If true (default), a gaussian kernel is used, if false a uniform kernel is used
+        sigma: Standard deviation of the gaussian kernel (default: 1.5), anisotropic kernels are possible.
+            Ignored if a uniform kernel is used
+        kernel_size: size of the uniform kernel (default: 11), anisotropic kernels are possible.
+            Ignored if a gaussian kernel is used
```
Borda, line 209:
```suggestion
        sigma: Standard deviation of the gaussian kernel, anisotropic kernels are possible.
            Ignored if a uniform kernel is used
        kernel_size: the size of the uniform kernel, anisotropic kernels are possible.
            Ignored if a Gaussian kernel is used
```

torchmetrics/functional/image/ssim.py
```diff
@@ -124,43 +165,56 @@ def _ssim_compute(
...
         data_range: Range of the image. If ``None``, it is determined from the image (max - min)
         k1: Parameter of SSIM.
         k2: Parameter of SSIM.
+        return_full_image: If true, the full ssim image is returned as a second arguments.
+            Mutually exlusive with return_contrast_sensitivity
+        return_contrast_sensitivity: If true, the contrast term is returned as a second argument.
+            The luminance term can be obtained with luminance=ssim/contrast
+            Mutually exlusive with return_full_image
```
Borda, line 217:
```suggestion
        return_full_image: If true, the full ``ssim`` image is returned as a second argument.
            Mutually exclusive with ``return_contrast_sensitivity``
        return_contrast_sensitivity: If true, the constant term is returned as a second argument.
            The luminance term can be obtained with luminance=ssim/contrast
            Mutually exclusive with ``return_full_image``
```

torchmetrics/image/ssim.py
```diff
@@ -26,17 +26,21 @@ class StructuralSimilarityIndexMeasure(Metric):
...
-            - ``'sum'``: takes the sum
-            - ``'none'``: no reduction will be applied
-
+        preds: estimated image
+        target: ground truth image
+        gaussian_kernel: If true (default), a gaussian kernel is used, if false a uniform kernel is used
+        sigma: Standard deviation of the gaussian kernel (default: 1.5), anisotropic kernels are possible.
+            Ignored if a uniform kernel is used
+        kernel_size: size of the uniform kernel (default: 11), anisotropic kernels are possible.
+            Ignored if a gaussian kernel is used
```
Borda, line 35:
```suggestion
        sigma: Standard deviation of the gaussian kernel, anisotropic kernels are possible.
            Ignored if a uniform kernel is used
        kernel_size: the size of the uniform kernel, anisotropic kernels are possible.
            Ignored if a Gaussian kernel is used
```

torchmetrics/image/ssim.py
```diff
@@ -26,17 +26,21 @@ class StructuralSimilarityIndexMeasure(Metric):
...
         data_range: Range of the image. If ``None``, it is determined from the image (max - min)
         k1: Parameter of SSIM.
         k2: Parameter of SSIM.
+        return_full_image: If true, the full ssim image is returned as a second arguments.
+            Mutually exlusive with return_contrast_sensitivity
+        return_contrast_sensitivity: If true, the contrast term is returned as a second argument.
+            The luminance term can be obtained with luminance=ssim/contrast
+            Mutually exlusive with return_full_image
```
Borda, line 43:
```suggestion
        return_full_image: If true, the full ``ssim`` image is returned as a second argument.
            Mutually exclusive with ``return_contrast_sensitivity``
        return_contrast_sensitivity: If true, the constant term is returned as a second argument.
            The luminance term can be obtained with luminance=ssim/contrast
            Mutually exclusive with ``return_full_image``
```

justusschock (approved):

Borda:
@SkafteNicki @stancld ^^

torchmetrics/image/ssim.py
weningerleon: Yep, exactly, with `reduction=None` that function returned the full SSIM images. However, the SSIM metric is a scalar metric, i.e., `reduction=None` should return one scalar per image. This was not available.

weningerleon: You can have a look at the skimage implementation: https://github.com/scikit-image/scikit-image/blob/main/skimage/metrics/_structural_similarity.py#L15-L254
A scalar (cropped and mean SSIM image) is returned. In my opinion, this should also be the default here (of course in the batched version).

weningerleon: Note that the SSIM metric should not exactly the same as the mean of the SSIM image. The SSIM image has the same size as the input image. However, the SSIM metric needs to be calculated on a cropped version in order to avoid border effects

SkafteNicki: I agree that we should more or less match the skimage implementation. I just want to make sure that if I call:
```python
structural_similarity_index_measure(preds, target)
```
on a batch of images, then it will return a scalar and not vector equal to the batch size (also to keep things backwards compatible). The final reduction is what I want the `reduction` parameter to take care of.

justusschock: Adding to @SkafteNicki : the `reduction` is the reduction across the batch dimension. Meaning that if that one is `none` it would be equal to having one value per sample in the batch. I think it should default to some averaging to get a single scalar for a whole batch

Borda:
@SkafteNicki @stancld pls ^^ :rabbit:

justusschock:
@Borda why are the tests not running?

Borda:
> @Borda why are the tests not running?

I guess it is due to GH outage...

Borda:
@justusschock mind check the failing tests, pls? :rabbit:

SkafteNicki (approved):

weningerleon:
Glad I could help to improve this awesome project!
Thanks a lot for your support!

PR #833, opened 2022-02-10
Refactor/move args to kwargs

SkafteNicki:
## What does this PR do?

Move the three common arguments ``dist_sync_on_step``, ``process_group`` and ``dist_sync_fn`` to a ``kwargs`` argument.
The arguments are only meant for real power-users and therefore it does not make sense that they need to be directly available as arguments. This solution is complete backwards compatible and it is therefore more a question about getting more clean documentation as we do not need to repeat theses arguments everywhere.

Currently only updated one metric so we can agree on the refactor before applying it to all other metrics.

## Before submitting

## PR review

## Did you have fun?


docs/source/pages/overview.rst
```diff
@@ -348,3 +348,24 @@ In practise this means that:
     val = metric.compute() # this value cannot be back-propagated
 
 A functional metric is differentiable if its corresponding modular metric is differentiable.
+
+.. _Metric kwargs:
+
+****************************
+Advance distributed settings
```
justusschock, line 355:
```suggestion
Advanced distributed settings
```

docs/source/pages/overview.rst
```diff
@@ -348,3 +348,24 @@ In practise this means that:
...
+****************************
+
+If you are running in a distributed environment, TorchMetrics will automatically take care of the distributed
```
justusschock, line 358:
```suggestion
If you are running in a distributed environment, ``TorchMetrics`` will automatically take care of the distributed
```

docs/source/pages/overview.rst
```diff
@@ -348,3 +348,24 @@ In practise this means that:
...
+synchronization for you. However, the following three keyword arguments can be given to any modular metric for
```
justusschock, line 359:
```suggestion
synchronization for you. However, the following three keyword arguments can be given to any metric class for
```

I think modular means something different :D

docs/source/pages/overview.rst
```diff
@@ -348,3 +348,24 @@ In practise this means that:
...
+further control over the distributed synchronization:
```
justusschock, line 360:
```suggestion
further control over the distributed aggregation:
```

docs/source/pages/overview.rst
```diff
@@ -348,3 +348,24 @@ In practise this means that:
...
+
+- ``dist_sync_on_step``: This argument is ``bool`` that indicates if the metric should syncronize between
+  different devices every time ``forward`` is called. Setting this to ``True`` is in general not recommended
+  as syncronization is an expensive operation to do after each batch.
+
+- ``process_group``: By default we syncronize across the *world* i.e. all proceses being computed on. You
+  can provide an ``torch._C._distributed_c10d.ProcessGroup`` in this argument to specify exactly what
```
justusschock, line 367: Is this really the only way to specify this? Also this doesn't do cross-links and if we don't do cross-links we probably don't need the whole path including the private C part of torch, right?

torchmetrics/metric.py
```diff
@@ -70,25 +71,46 @@ class Metric(Module, ABC):
...
+            - dist_sync_fn: function that performs the allgather option on the metric state
     """
 
     __jit_ignored_attributes__ = ["device"]
     __jit_unused_properties__ = ["is_differentiable"]
     is_differentiable: Optional[bool] = None
     higher_is_better: Optional[bool] = None
+    dist_sync_on_step: bool = False
+    process_group: Optional[ProcessGroup] = None
+    dist_sync_fn: Optional[Callable] = None
```
justusschock, line 108: I think we cannot put them here, as these are static. Meaning that This way, I cannot have two instances of the same metric for different process goups or something like that. They need to be moved inside the `__init__`

SkafteNicki: True. Will change.

docs/source/pages/overview.rst
SkafteNicki: I am really not sure, but from the docs it seems so:
https://pytorch.org/docs/stable/distributed.html#torch.distributed.all_gather

docs/source/pages/overview.rst
```diff
@@ -348,3 +348,24 @@ In practise this means that:
     val = metric.compute() # this value cannot be back-propagated
 
 A functional metric is differentiable if its corresponding modular metric is differentiable.
+
+.. _Metric kwargs:
+
+****************************
+Advanced distributed settings
+****************************
```
SkafteNicki, line 356:
```suggestion
*****************************
Advanced distributed settings
*****************************
```

Borda (commented):
why do some have `metric_kwargs` and others `metric_kwargs`?

SkafteNicki:
> why do some have `metric_kwargs` and others `metric_kwargs`?

Sorry, my intension was that all should just be `kwargs`. Should be fixed now :]

Borda (approved):

justusschock (approved):

PR #851, opened 2022-02-21
Add argument `store_on_cpu` to image metrics

### LINKED ISSUE #848: Allow caching on cpu when compute some metrics.
## 🚀 Feature

Currently, some metrics like FID and KID only use GPU computing.
Since all datasets are cached in gpu memory, there is a big limitation on hardware.
It may be helpful to add `compute_on_cpu` option to compute these metrics like below.

```python
def update(self, imgs: Tensor, real: bool) -> None:  # type: ignore
    """Update the state with extracted features.

    Args:
        imgs: tensor with images feed to the feature extractor
        real: bool indicating if imgs belong to the real or the fake distribution
    """
    features = self.inception(imgs)

    if self.compute_on_cpu:
        features = features.detach().cpu()

    if real:
        self.real_features.append(features)
    else:
        self.fake_features.append(features)
```

### Motivation

I can compute FID on cloud system but not my local, because of the memory issue.
Currently, warn large memory foorprint. #468

SkafteNicki:
## What does this PR do?

Fixes #848 

Adds argument `store_on_cpu` to 3 image metrics, such that extracted features can be stored in CPU memory instead of GPU memory.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `**kwargs` argument for passing additional arguments to base class ([#833](https://github.com/PyTorchLightning/metrics/pull/833))
 
 
-- Added argument `store_on_cpu` to `FrechetInceptionDistance`, `KernelInceptionDistance`, `InceptionScore` to store features on cpu ([]())
+- Added argument `store_on_cpu` to `FrechetInceptionDistance`, `KernelInceptionDistance`, `InceptionScore` to store features on cpu ([#851](https://github.com/PyTorchLightning/metrics/pull/851))
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `ClasswiseWrapper` for better logging of classification metrics with multiple output values ([#832](https://github.com/PyTorchLightning/metrics/pull/832))
 
 
+- Added `**kwargs` argument for passing additional arguments to base class ([#833](https://github.com/PyTorchLightning/metrics/pull/833))
+
+
 ### Changed
 
 
 ### Deprecated
 
-- Deprecated method `compute_on_step` ([#792](https://github.com/PyTorchLightning/metrics/pull/792))
+- Deprecated argument `compute_on_step` ([#792](https://github.com/PyTorchLightning/metrics/pull/792))
+
+
+- Deprecated passing in `dist_sync_on_step`, `process_group`, `dist_sync_fn` direct argument ([#833](https://github.com/PyTorchLightning/metrics/pull/833))
 
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Made `num_classes` in `jaccard_index` a required argument ([#853](https://github.com/PyTorchLightning/metrics/pull/853))
+
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `**kwargs` argument for passing additional arguments to base class ([#833](https://github.com/PyTorchLightning/metrics/pull/833))
 
 
+- Added argument `store_on_cpu` to `FrechetInceptionDistance`, `KernelInceptionDistance`, `InceptionScore` to store features on cpu ([]())
+
 ### Changed
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added argument `store_on_cpu` to `FrechetInceptionDistance`, `KernelInceptionDistance`, `InceptionScore` to store features on cpu ([#851](https://github.com/PyTorchLightning/metrics/pull/851))
 
+
+- Added negative `ignore_index` for the Accuracy metric ([#362](https://github.com/PyTorchLightning/metrics/pull/362))
+
+
 ### Changed
 
 - Made `num_classes` in `jaccard_index` a required argument ([#853](https://github.com/PyTorchLightning/metrics/pull/853))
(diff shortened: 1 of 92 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `**kwargs` argument for passing additional arguments to base class ([#833](https://github.com/PyTorchLightning/metrics/pull/833))
 
 
+- Added argument `store_on_cpu` to `FrechetInceptionDistance`, `KernelInceptionDistance`, `InceptionScore` to store features on cpu ([#851](https://github.com/PyTorchLightning/metrics/pull/851))
```
Borda, line 29: I think that this could be a very similar case for all metrics using a DNN under the hood some for example other Text metrics...

Borda (approved):

CHANGELOG.md
SkafteNicki: Not necessarily all such metrics. For example the `lpips` also uses a DNN underneat but only need to store a scalar value:
https://github.com/PyTorchLightning/metrics/blob/b0bcccb12ee80cbf72e596d3bf43dfb533e30e22/torchmetrics/image/lpip.py#L140-L142
Instead it would be appropriate for all metrics where the state can be large

justusschock:
Shouldn't we add this to all metrics to specify where the states are saved?

Borda:
> Shouldn't we add this to all metrics to specify where the states are saved?

if so lets implement it in a separate pr :)

torchmetrics/image/kid.py
```diff
@@ -112,6 +112,9 @@ class KernelInceptionDistance(Metric):
             Scale-length of polynomial kernel. If set to ``None`` will be automatically set to the feature size
         coef:
             Bias term in the polynomial kernel.
+        store_on_cpu: By default features will be calculated and stored on the device the metric is located on.
+            By setting this argument to `True`, features will be forced to be stored on CPU even if they are
```
Borda, line 116:
```suggestion
            By setting this argument to ``True``, features will be forced to be stored on the CPU even if they are
```

torchmetrics/image/inception.py
```diff
@@ -61,6 +61,10 @@ class InceptionScore(Metric):
 
         splits: integer determining how many splits the inception score calculation should be split among
 
+        store_on_cpu: By default features will be calculated and stored on the device the metric is located on.
+            By setting this argument to `True`, features will be forced to be stored on CPU even if they are
```
Borda, line 65:
```suggestion
            By setting this argument to ``True``, features will be forced to be stored on the CPU even if they are
```

torchmetrics/image/fid.py
```diff
@@ -161,6 +161,10 @@ class FrechetInceptionDistance(Metric):
             - an ``nn.Module`` for using a custom feature extractor. Expects that its forward method returns
               an ``[N,d]`` matrix where ``N`` is the batch size and ``d`` is the feature size.
 
+        store_on_cpu: By default features will be calculated and stored on the device the metric is located on.
+            By setting this argument to `True`, features will be forced to be stored on CPU even if they are
```
Borda, line 165:
```suggestion
            By setting this argument to ``True``, features will be forced to be stored on the CPU even if they are
```

Borda (approved):

torchmetrics/image/kid.py
```diff
@@ -112,6 +112,9 @@ class KernelInceptionDistance(Metric):
             Scale-length of polynomial kernel. If set to ``None`` will be automatically set to the feature size
         coef:
             Bias term in the polynomial kernel.
+        store_on_cpu: By default features will be calculated and stored on the device the metric is located on.
```
rohitgr7, line 115: quick ques: there are not just stored but also computed on CPU ?

if that's the case, do you think a better name might be `compute_on_cpu`?

torchmetrics/image/fid.py
```diff
@@ -237,6 +244,10 @@ def __init__(
         else:
             raise TypeError("Got unknown input to argument `feature`")
 
+        if not isinstance(store_on_cpu, bool):
+            raise TypeError(f"Expected argument `store_on_cpu` to be a bool but got {store_on_cpu}")
```
rohitgr7, line 248:
```suggestion
            raise TypeError(f"Expected argument `store_on_cpu` to be a bool but got {store_on_cpu!r}.")
```

torchmetrics/image/inception.py
```diff
@@ -135,6 +142,10 @@ def __init__(
         else:
             raise TypeError("Got unknown input to argument `feature`")
 
+        if not isinstance(store_on_cpu, bool):
+            raise TypeError(f"Expected argument `store_on_cpu` to be a bool but got {store_on_cpu}")
```
rohitgr7, line 146:
```suggestion
            raise TypeError(f"Expected argument `store_on_cpu` to be a bool but got {store_on_cpu!r}.")
```

torchmetrics/image/kid.py
```diff
@@ -222,6 +228,10 @@ def __init__(
             raise ValueError("Argument `coef` expected to be float larger than 0")
         self.coef = coef
 
+        if not isinstance(store_on_cpu, bool):
+            raise TypeError(f"Expected argument `store_on_cpu` to be a bool but got {store_on_cpu}")
```
rohitgr7, line 232:
```suggestion
            raise TypeError(f"Expected argument `store_on_cpu` to be a bool but got {store_on_cpu!r}.")
```

rohitgr7 (approved):

torchmetrics/image/fid.py
Borda: just curious, what does it do?

torchmetrics/image/kid.py
Borda: it is just stored, you always want to use GPU if possible, otherwise the metrics could be bottle-neck :)

rohitgr7: but I see, if they are stored on CPU, I think the computation will also happen CPU only.
if that's the case, then `store_on_cpu`, might not reflect the point that computation will also happen on CPU only.

or are they moved back to the device during computation?

Borda: good point @SkafteNicki

SkafteNicki:
> Shouldn't we add this to all metrics to specify where the states are saved?

As pointed out by @justusschock we could probably benefit from a more general solution that works for all metrics where the state can be a list. Also as @rohitgr7 points out, we need to secure that calculations still happens on the right device, so we need to do some more advance moving of data. Closing this PR, trying to come up with a general solution.

PR #853, opened 2022-02-21
Update num_classes in jaccard score to be a required argument

### LINKED ISSUE #829: Jaccard Index behaviour when num_classes is absent
The [jaccard index function](https://github.com/PyTorchLightning/metrics/blob/8fade87062a7b87c1e6429bbe1c4e0112b3713a5/torchmetrics/functional/classification/jaccard.py#L69) determines the number of classes automatically.
This can result in strange results if the user forgets to specify `num_classes`. For example, in IoU determination in semantic segmentation, not all images have all classes. So determining them automatically will almost always lead to wrong number.
Additionally, the [warning](https://github.com/PyTorchLightning/metrics/blob/8fade87062a7b87c1e6429bbe1c4e0112b3713a5/torchmetrics/utilities/data.py#L156) that shows up makes it seem like something is wrong, when the user specifies the number of classes explicitly (but this is not the case at all):

```python
/run/determined/pythonuserbase/lib/python3.7/site-packages/torchmetrics/utilities/prints.py:36: RuntimeWarning: You have set 11 number of classes which is different from predicted (1000) and target (10) number of classes
```

Would it be crazy if `num_classes` requirement is default and the automatic determination of classes is a special option that must be enabled?

SkafteNicki:
## What does this PR do?

Fixes #829

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/torchmetrics/functional/classification/jaccard.py
+++ b/torchmetrics/functional/classification/jaccard.py
@@ -69,10 +69,10 @@ def _jaccard_from_confmat(
 def jaccard_index(
     preds: Tensor,
     target: Tensor,
+    num_classes: int,
     ignore_index: Optional[int] = None,
     absent_score: float = 0.0,
     threshold: float = 0.5,
-    num_classes: Optional[int] = None,
     reduction: str = "elementwise_mean",
 ) -> Tensor:
     r"""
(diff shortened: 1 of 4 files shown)


torchmetrics/functional/classification/jaccard.py
```diff
@@ -69,10 +69,10 @@ def _jaccard_from_confmat(
 def jaccard_index(
     preds: Tensor,
     target: Tensor,
+    num_classes: int,
```
Borda, line 72: this may be breaking some use-cases, as it reorders the arguments in case a user is using positional args

Borda (approved):

justusschock (approved):

PR #867, opened 2022-02-28
New argument `compute_on_cpu`

### LINKED ISSUE #848: Allow caching on cpu when compute some metrics.

### LINKED ISSUE #866: `MeanAveragePrecision` CPU overload when using multi-device ddp
## 🐛 Bug

I was using `pytorchlightning.Trainer` in `ddp` mode with 8 GPUs on a HPC. There are 5k images in my validation set, and I find all the CPU cores are fully loaded (100%) for extremely long time (~2 hours) at `validation_epoch_end()` stage.

### To Reproduce

1. using `ddp` mode with `number of device > 2`
2. using large validation/test set.

#### Code sample

```python
class Model(pl.LightningModule):
    def __init__(self):
        self.det = FasterRCNN(...)
        self.map = MeanAveragePrecision(compute_on_step=False)
        ...

    ...

    def validation_step(self, batch, batch_idx):
        images, target = batch
        preds = self.det(images)
        self.map.update(preds=preds, target=target)

    def validation_epoch_end(self, outs):
        val_map_out = self.map.compute()  <-- CPU overload
        self.map.reset()
```

### Expected behavior

The same computation takes place in `N` separate sub-process simultaneously. If the computations are on GPU, they should finish in roughly the same time. But current implementation moves all tensors to CPU and CPU overload occurs while `N` increases.

- ddp with 1 GPU, number of fully loaded GPU cores 16/32, takes ~6 mins.
- ddp with 2 GPU, number of fully loaded GPU cores 32/32, takes ~6 mins.
- ddp with >2 GPU, number of fully loaded GPU cores 32/32, takes > 1 hours.

### Environment

- OS (e.g., Linux): Linux
- Python: 3.8
- PyTorch: 1.8.1 and 1.10.2
- pytorch-lightning: 1.5.10
- torchmetrics: 0.7.2

### Temp solution

My temp solution is re-enable the mAP computation on GPU, although the GPU version is slower than CPU version #677 

I commented the following lines and made other relevant variables on the correct device.

```python
    def compute(self) -> dict:
        # move everything to CPU, as we are faster here
        # self.detection_boxes = [box.cpu() for box in self.detection_boxes]
        # self.detection_labels = [label.cpu() for label in self.detection_labels]
        # self.detection_scores = [score.cpu() for score in self.detection_scores]
        # self.groundtruth_boxes = [box.cpu() for box in self.groundtruth_boxes]
        # self.groundtruth_labels = [label.cpu() for label in self.groundtruth_labels]
```

Should we make `compute_on_cpu` optional? and any suggestion according to TorchMetrics' API design?

I'm also keeping my eyes on the work of improving mAP performance on both CPU and GPU. #742

SkafteNicki:
## What does this PR do?

Rework of #851 
Fixes #848
General solution that allows *all* metrics where the metric state is a list to store the state on CPU instead of GPU through a new keyword argument `compute_on_cpu=True/False` (default `False`). This can beneficial in situations where the user do not have that much VRAM. Consequence of enabling this feature is that code in `compute` will be executed on CPU (code in `update` will still be executed on GPU).

Fixes #866
Related as the MAP metric currently moves all metric states to CPU for faster computations. As the issue points out this can overload the CPU. Feature in this PR essentially also fixes this by making it an option to move to CPU by setting `MeanAveragePrecision(compute_on_cpu=True)`.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/docs/source/pages/overview.rst
+++ b/docs/source/pages/overview.rst
@@ -351,9 +351,19 @@ A functional metric is differentiable if its corresponding modular metric is dif
 
 .. _Metric kwargs:
 
-*****************************
-Advanced distributed settings
-*****************************
+************************
+Advanced metric settings
+************************
+
+The following is a list of additional arguments that can be given to any metric class (in the ``**kwargs`` argument)
+that will alter how metric states are stored and synced.
+
+If you are running metrics on GPU and are encountering that you are running out of GPU VRAM then the following
+argument can help:
+
+- ``compute_on_cpu`` will automatically move the metric states to cpu after calling ``update``, making sure that
+  GPU memory is not filling up. The consequence will be that the ``compute`` method will be called on CPU instead
+  of GPU. Only applies to metric states that are lists.
 
 If you are running in a distributed environment, ``TorchMetrics`` will automatically take care of the distributed
 synchronization for you. However, the following three keyword arguments can be given to any metric class for
--- a/tests/image/test_inception.py
+++ b/tests/image/test_inception.py
@@ -104,11 +104,12 @@ class _ImgDataset(Dataset):
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="test is too slow without gpu")
 @pytest.mark.skipif(not _TORCH_FIDELITY_AVAILABLE, reason="test requires torch-fidelity")
-def test_compare_is(tmpdir):
+@pytest.mark.parametrize("compute_on_cpu", [True, False])
+def test_compare_is(tmpdir, compute_on_cpu):
     """check that the hole pipeline give the same result as torch-fidelity."""
     from torch_fidelity import calculate_metrics
 
-    metric = InceptionScore(splits=1).cuda()
+    metric = InceptionScore(splits=1, compute_on_cpu=compute_on_cpu).cuda()
 
     # Generate some synthetic data
     img1 = torch.randint(0, 255, (100, 3, 299, 299), dtype=torch.uint8)
--- a/tests/regression/test_pearson.py
+++ b/tests/regression/test_pearson.py
@@ -52,9 +52,9 @@ def _sk_pearsonr(preds, target):
 )
 class TestPearsonCorrcoef(MetricTester):
     atol = 1e-2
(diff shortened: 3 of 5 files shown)


maximsch2:
A few questions:
- If I'm reading code correctly, `self.device` will still be `gpu` with this during `compute`? I'm wondering if this will break some metrics. 
- The implicit assumption here is that a List state is only appended during `update` as otherwise we'll get device mismatches, right?
- This doesn't help with other metrics that can occupy significant memory (e.g. binned curve metrics) on GPU as those are allocated as fixed size tensors. In that case though, I guess, a user can manually place them on CPU and transfer data to CPU before `update`. Is that what we would be recommending people?

justusschock:
@maximsch2 I think the device should indicate the input and output device of a metric. E.g. even if the metic states are on cpu, it would require the devices of the inputs to be cuda and handle the device-transfer logic accordingly and the output would also be on gpu.

Borda:
@SkafteNicki mind resolve conflicts :rabbit: 
@justusschock mind review? :fish:

Borda:
@SkafteNicki @justusschock pls ^^ :rabbit:

justusschock (approved):

Borda:
seems there is a DDP issue... :/

PR #874, opened 2022-03-06
added micro average option for torch metrics

### LINKED ISSUE #826: Micro Average IoU
## 🚀 Feature

It would be nice to have IoU accept parameters similar to Recall and Precision, in a way that we can able to calculate the micro average version as well. The current implementation is calculating the macro one when `reduction=elementwise_mean`.

razmikmelikbekyan:
## What does this PR do?

Added Micro average option for IoU.

Fixes #826

## Before submitting

- [X] Was this **discussed/approved** via a Github issue? (no need for typos and docs improvements)
- [X] Did you read the [contributor guideline](https://github.com/PyTorchLightning/metrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [X] Did you write any new **necessary tests**?

## PR review

## Did you have fun?


razmikmelikbekyan:
@Borda Please have a look. After initial review, I will remove all the commented-out lines, but there might be some incompatibility related issues since I was going to remove the `reduction` parameter fully.

tests/classification/test_jaccard.py
```diff
@@ -104,60 +104,61 @@ def _sk_jaccard_multidim_multiclass(preds, target, average=None):
...
     if half_ones:
         preds[:60] = 1
     jaccard_val = jaccard_index(
         preds=preds,
         target=target,
+        average=average,
         num_classes=3,
         ignore_index=ignore_index,
-        reduction=reduction,
+        # reduction=reduction,
```
justusschock, line 161: I am afraid, that we cannot simply remove this. I think we shall rather extend it, but not all the possibilities may always be restricted to some sort of averaging. That is why I prefer to keep this as reduction. 
Also we want this to be consistent across the entire codebase.

Even if we decide to change this to `average` which again, I am very much against, we would have to follow a proper deprecation cycle here

justusschock (changes requested):
implementation-wise it looks good to me. Just not sure how to proceed with different arguments for different metrics. IMO it should all be called `reduction`, but not sure how to handle different arguments. Will we have separate helper functions or just one function that does it all and some others just for input validation?

tests/classification/test_jaccard.py
razmikmelikbekyan: @justusschock the implementation is similar to Recall and Precision implementation in the codebase, where there is no `reduction` but rather `average` which is standard for sklearn as well. The current implementation extends the functionality of using reduction - `reduction=elemntwise_mean` = `average=macro`, `reduction=None` = `average=None`, I understand depreciation-related issues, but I did not catch why we should not standardize the code like for Precision and Recall, since, let's say, they are the metrics from the same family.

justusschock: Actually, I would rather do it the other way around and have the `reduction` argument everywhere as this is more general 

cc @Borda @SkafteNicki

Borda: well yeah found the naming issue while working on #891 as some metrics uses `average` and other `reduction`

Borda: just raised and ask for feedback in @PyTorchLightning/core-metrics and @PyTorchLightning/core-lightning

Borda: I agree as the average "sum" does not make much sense...

awaelchli: Would this work? 
- use "reduction" everywhere where applicable, standardize it's meaning (as they did in torch)
- where "reduction" is not applicable, drop it and allow the metric to choose their own semantics, e.g., "average". For example, for precision a "reduction=sum" does not make sense.

Borda: I would say do the update in parallel PR to all other arguments, and then come back to this PR...
cc: @justusschock for suggestions?

Borda:
@razmikmelikbekyan how is it going here? :rabbit:

razmikmelikbekyan:
> @razmikmelikbekyan how is it going here? 🐰

Honestly said I am a bit confused about the next steps. I am not sure that I will have enough time right now to refactor all the other metrics and so on and I do not understand what exactly needs to be done PR. 

Should we merge it or we want to change name back to reduction and so one (in my opinion name average_type is more relevant and adapted by the community for this kind of metrics than reduction)

Borda:
> Honestly said I am a bit confused about the next steps.

@stancld mind help with this PR? :)

justusschock:
Hey @razmikmelikbekyan ,

Very sorry for the long return time. The decisions made offline with @Borda and @SkafteNicki were that we want to get this to merge now and that we will deal with the naming inconsistencies while doing the major refactor (#1001 ). 
That being said, do you think, there is a chance that you could look into the failing tests?

razmikmelikbekyan:
> Hey @razmikmelikbekyan ,
> 
> Very sorry for the long return time. The decisions made offline with @Borda and @SkafteNicki were that we want to get this to merge now and that we will deal with the naming inconsistencies while doing the major refactor (#1001 ). That being said, do you think, there is a chance that you could look into the failing tests?

hey @justusschock, yep, I will have a look soon.

Borda (approved):

SkafteNicki (approved):

PR #876, opened 2022-03-07
Revert "Refactor: Rename update and compute methods to _update and _compute"

SkafteNicki:
Reverts PyTorchLightning/metrics#840
Based on @awaelchli feedback we are reverting to the old `update` and `compute` format.


SkafteNicki:
For now we are probably going to keep the wrapping. While I still agree that it is not great, our end users does not really seem to care that much (at least we do not receive any issues on this specific topic)

awaelchli (approved):

Borda:
> at least we do not receive any issues on this specific topic

I ca file some :D

Borda (approved):

Issue #880, opened 2022-03-09
[Feature Request] Deterministic ConfusionMatrix on GPU


mayeroa:
## 🚀 Feature

For now, we cannot use `ConfusionMatrix` with a lightning `Trainer` configured like this: `Trainer(gpus=1, deterministic=True)`.
The following error is returned:

```bash
File "./lib/python3.8/site-packages/torchmetrics/functional/classification/confusion_matrix.py", line 48, in _confusion_matrix_update
    bins = torch.bincount(unique_mapping, minlength=minlength)
RuntimeError: _bincount_cuda does not have a deterministic implementation, but you set 'torch.use_deterministic_algorithms(True)'. You can turn off determinism just for this operation if that's acceptable for your application. You can also file an issue at https://github.com/pytorch/pytorch/issues to help us prioritize adding deterministic support for this operation.
```

### Motivation

The idea would be to replace the call to `torch.bincount` with something deterministic on GPU to properly use the GPU for metrics computation under a reproductible scenario (using a `seed` and `deterministic` flag).

### Pitch

Make sure that we can run the `ConfusionMatrix` metric on GPU with deterministic algorithms.

### Alternatives

As an alternative, we can move the metric computation on cpu (instead of gpu) but this defeats the principle of having pure pytorch metrics.

### Additional context

As additional context, I have made a small script that produces the error:

```python
import os

import torch
import torch.nn.functional as F
from torch.utils.data import DataLoader, Dataset

from pytorch_lightning import LightningModule, Trainer, seed_everything
from torchmetrics import ConfusionMatrix, Accuracy


class RandomDataset(Dataset):
    def __init__(self, size, length, num_classes=10):
        self.size = size
        self.len = length
        self.num_classes = num_classes

    def __getitem__(self, index):
        image = torch.rand(self.size, self.size, 3)
        label = torch.randint(low=0, high=self.num_classes, size=(1, ))
        return image, label

    def __len__(self):
        return self.len


class BoringModel(LightningModule):
    def __init__(self, num_classes=10):
        """Testing PL Module.

        Use as follows:
        - subclass
        - modify the behavior for what you want

        """
        super().__init__()
        self.layer_1 = torch.nn.Linear(32 * 32 * 3, 32)
        self.layer_1a = torch.nn.ReLU()
        self.layer_end = torch.nn.Linear(32, num_classes)

        self.train_acc = Accuracy(num_classes=num_classes)
[...]

PR #900, opened 2022-03-22
Fix torch.bincount when running in deterministic mode and on GPU

SkafteNicki:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/880
Implement an alternative of `torch.bincount` when the user is running in deterministic mode and on GPU.

## Before submitting

## PR review

## Did you have fun?


CHANGELOG.md
```diff
@@ -124,6 +124,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed bug in MAP metric related to either no ground truth or no predictions ([#884](https://github.com/PyTorchLightning/metrics/pull/884))
 
 
+- Fixed `ConfusionMatrix` on GPU when running in deterministic mode ([]())
```
SkafteNicki, line 127:
```suggestion
- Fixed `ConfusionMatrix` on GPU when running in deterministic mode ([#900](https://github.com/PyTorchLightning/metrics/pull/900))
```

CHANGELOG.md
```diff
@@ -124,6 +124,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed bug in MAP metric related to either no ground truth or no predictions ([#884](https://github.com/PyTorchLightning/metrics/pull/884))
 
 
+- Fixed `ConfusionMatrix` on GPU when running in deterministic mode ([#900](https://github.com/PyTorchLightning/metrics/pull/900))
```
SkafteNicki, line 127:
```suggestion
- Fixed `ConfusionMatrix`, `AUROC` and `AveragePrecision` on GPU when running in deterministic mode ([#900](https://github.com/PyTorchLightning/metrics/pull/900))
```

torchmetrics/utilities/data.py
```diff
@@ -259,3 +259,25 @@ def _squeeze_scalar_element_tensor(x: Tensor) -> Tensor:
...
+
+    Returns:
+        Number of occurences for each unique element in x
+    """
+    if x.is_cuda and torch.are_deterministic_algorithms_enabled():
+        if minlength is None:
+            minlength = len(torch.unique(x))
+        output = torch.zeros(minlength, device=x.device)
+        for i in range(minlength):
+            output[i] = (x == i).sum()
```
justusschock, line 280: is there any occasion, where this might be run with floats? I don't think so, right? Because then it could be difficult with machine precision :)
But I don't think this is the case here, right?

torchmetrics/utilities/data.py
```diff
@@ -259,3 +259,25 @@ def _squeeze_scalar_element_tensor(x: Tensor) -> Tensor:
...
+        x: tensor to count
+        minlength: minimum length to count
+
+    Returns:
+        Number of occurences for each unique element in x
+    """
+    if x.is_cuda and torch.are_deterministic_algorithms_enabled():
+        if minlength is None:
+            minlength = len(torch.unique(x))
+        output = torch.zeros(minlength, device=x.device)
```
justusschock, line 278:
```suggestion
        output = torch.zeros(minlength, device=x.device, dtype=torch.long)
```

I think default is float, right? Wouldn't make sense to implicitly convert counts to float

justusschock (approved):

tests/test_utilities.py
```diff
@@ -116,3 +116,21 @@ def test_flatten_dict():
...
+    # uses custom implementation
+    res1 = _bincount(x, minlength=10)
+
+    torch.use_deterministic_algorithms(False)
+
+    # uses torch.bincount
+    res2 = _bincount(x, minlength=10)
+
+    # check for correctness
+    assert torch.allclose(res1, res2)
```
Borda, line 136: should this be rather identical?

justusschock: pytorch recommends using torch.allclose especially when dealing with floating point tensors. But once we changed the default dtype to torch.long, this should be the same. So I recommend keeping it like this, since with integer-tensors you cannot reach the specified tolerances without them being identical :)

SkafteNicki: agree here :]

torchmetrics/utilities/data.py
SkafteNicki: this function is always called with `long` input :]

Borda (approved):

tests/test_utilities.py
Borda: well but the counts are `dtype=torch.long`, right?

Borda: weel the code below has it set as `dtype=torch.long`, right?

Borda:
@SkafteNicki seems that ERGAS is now failing on Windows with the latest configurations :/

PR #902, opened 2022-03-22
[WIP] Investigate potential memory bug

### LINKED ISSUE #890: Inefficient memory consumption by MeanMetric
https://github.com/PyTorchLightning/metrics/blob/3af729508289d3babf0e166d9e8405cb2b0758a2/torchmetrics/aggregation.py#L333-L408

Old average meters eats nothing compare to 16+ gb RAB by MeanMetric
```py
class AverageMeter(object):
    """Computes and stores the average and current value"""

    def __init__(self):
        self.reset()

    def reset(self):
        self.val = .0
        self.avg = .0
        self.sum = .0
        self.count = .0

    def update(self, val, n=1):
        self.val = val
        self.sum += val * n
        self.count += n
        self.avg = self.sum / self.count
```

SkafteNicki:
## What does this PR do?

Fixes #890 
Investigating potential memory bug. Currently added test case.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/tests/bases/test_metric.py
+++ b/tests/bases/test_metric.py
@@ -11,11 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 import pickle
 from collections import OrderedDict
 
 import cloudpickle
 import numpy as np
+import psutil
 import pytest
 import torch
 from torch import Tensor, nn, tensor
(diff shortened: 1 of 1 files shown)


Borda:
fixed the missing package in tests :)

requirements/test.txt
```diff
@@ -10,8 +10,10 @@ mypy>=0.790
 phmdoctest>=1.1.1
 pre-commit>=1.0
 
+psutil
 requests
 fire
 
 cloudpickle>=1.3
 scikit-learn>=0.24
+psutil
```
SkafteNicki, line 19:
```suggestion
```

torchmetrics/metric.py
```diff
@@ -247,6 +248,7 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
         cache = {attr: getattr(self, attr) for attr in self._defaults}
 
         # call reset, update, compute, on single batch
+        self._enable_grad = True  # allow grads for batch computation
```
justusschock, line 251: shouldn't we check whether this is enabled by `torch.is_grad_enabled`?

SkafteNicki: This flag controls if we allow gradient to be passed through not torch or the user.
The only time we allow this is when the user computes the metric value on the current batch.

Every time we accumulate we want to disable this such that memory does not increase over time. 
That is actual problem this PR is solving, as we currently was not disabling it if user called `metric.update`.

tests/bases/test_metric.py
```diff
@@ -362,3 +364,42 @@ def device(self):
...
+
+
+@pytest.mark.parametrize("device", ["cpu", "cuda"])
+@pytest.mark.parametrize("requires_grad", [True, False])
+def test_constant_memory(device, requires_grad):
+    """Checks that when updating a metric the memory does not increase."""
+    if not torch.cuda.is_available() and device == "cuda":
+        pytest.skip("Test requires GPU support")
+
+    def get_memory_usage():
```
Borda, line 376: Curious if this is safe in case you run tests in parallel?

SkafteNicki: I think it is safe because we explicit ask for the memory consumption of the current process

Borda (approved):

justusschock (approved):

tchaton (approved):
LGTM !

rohitgr7 (approved):

PR #913, opened 2022-03-24
Make `best_metric` in MetricTracker more robust

### LINKED ISSUE #911: Metric Collection Tracker Comptability
## 🐛 Bug
Hi , 
In the last version , you added the option to combine MetricTracker with MetricCollection.
When I used it , I tried to use the "best_metric" method for MetricTracker. Unfortunately , When I have metrics which doesn't have a best metric option (like confusion matrix) in the MetricCollection object , "best_metric" method return an error.
 
I think that as this function should be general , you should have support for this case as well 

#### Code sample
```py
from torchmetrics import MetricTracker, MetricCollection,Accuracy,Recall,ConfusionMatrix
_ = torch.manual_seed(42)
tracker = MetricTracker(MetricCollection([Accuracy(),Recall(),ConfusionMatrix(num_classes=2)]))
for epoch in range(5):
    tracker.increment()
    for batch_idx in range(5):
        preds, target = torch.randint(0,2,(100,)), torch.randint(0, 2, (100,))
        tracker.update(preds, target)
    print(f"current stats={tracker.compute()}")
from pprint import pprint
best_res, which_epoch = tracker.best_metric(return_step=True)
pprint(best_res)
```

That's Result in this Error:
```
Traceback (most recent call last):
  File "<input>", line 16, in <module>
  File "/Users/*****/.pyenv/versions/*****/lib/python3.7/site-packages/torchmetrics/wrappers/tracker.py", line 177, in best_metric
    idx[k], best[k] = out[0].item(), out[1].item()
ValueError: only one element tensors can be converted to Python scalars
```

### Expected behavior
The solution is to wrap the iteration over the metrics in MetricCollection with a try and except and just ignore the non-computable metrics 
from the original code of MetricTracker.best_metric (line 174)
```py
            for i, (k, v) in enumerate(res.items()):
                try:
                    fn = torch.max if maximize[i] else torch.min
                    out = fn(v, 0)
                    idx[k], best[k] = out[0].item(), out[1].item()
                except:
                    some predefined message for that case 
# after running over al metrics if there no metric that has a "score"              
            if len(idx)==0:
                   some predefined error 
                   return []
```

### Environment

- OS (Linux):
- Python & PyTorch Version (3.7.5, 1.10)
- TorchMetrics version 0.8.0.dev0

SkafteNicki:
## What does this PR do?

Fixes #911 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allowed reduction `None` ([#891](https://github.com/PyTorchLightning/metrics/pull/891))
 
 
+- `MetricTracker.best_metric` will now give a warning when computing on metric that do not have a best ([#913](https://github.com/PyTorchLightning/metrics/pull/913))
+
+
 ### Deprecated
 
 - Deprecated argument `compute_on_step` ([#792](https://github.com/PyTorchLightning/metrics/pull/792))
--- a/tests/wrappers/test_tracker.py
+++ b/tests/wrappers/test_tracker.py
@@ -16,7 +16,15 @@ import pytest
 import torch
 
 from tests.helpers import seed_all
-from torchmetrics import Accuracy, MeanAbsoluteError, MeanSquaredError, MetricCollection, Precision, Recall
+from torchmetrics import (
+    Accuracy,
+    ConfusionMatrix,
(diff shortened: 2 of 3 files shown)


Borda (approved):

stancld (approved):
Good idea! LGTM! :]

PR #947, opened 2022-04-11
Bugfix/map device placement

### LINKED ISSUE #943: MeanAveragePrecision devices mismatch
Version: TorchMetrics 0.7.3

I have noticed a bug in MeanAveragePrecision() class.
How to reproduce:
The code in pytorch-lightning module is: 
```py
__init__(...):
  self.mAP = MeanAveragePrecision()

validation_step(...):
  self.mAP(preds, targets)  # preds and targets have the same device as pl-module.

validation_epoch_end(...):
  mAP_dict = self.mAP.compute()
  self.log_dict(mAP_dict)
  
```
I get a device mismatch error at line 715 in map.py, when I run my code on a GPU.

I dug deeper and found the problem:
On line 781 is the following code:
```
       # move everything to CPU, as we are faster here
        self.detection_boxes = [box.cpu() for box in self.detection_boxes]
        self.detection_labels = [label.cpu() for label in self.detection_labels]
        self.detection_scores = [score.cpu() for score in self.detection_scores]
        self.groundtruth_boxes = [box.cpu() for box in self.groundtruth_boxes]
        self.groundtruth_labels = [label.cpu() for label in self.groundtruth_labels]
```
This means, that all gt boxes and detections are stored in the CPU, but in the function _evaluate_image(...), when we have no predictions but have GT we call function __evaluate_image_preds_no_gt(...) (line 470 in map.py).
In this function we have:
``` 
       return {
            "dtMatches": torch.zeros((nb_iou_thrs, nb_det), dtype=torch.bool, device=self.device),
            "gtMatches": torch.zeros((nb_iou_thrs, nb_gt), dtype=torch.bool, device=self.device),
            "dtScores": scores_sorted,
            "gtIgnore": gt_ignore,
            "dtIgnore": det_ignore,
        }
```
The problem is the part where we return device=self.device (we moved everything to CPU before).

My questions are:
Have I discovered a bug or do I have an error in my code?
Why has nobody else before encountered this problem, since this bug makes this metric unusable?

Thank you for your answers. 

Edit: I forked the report and edited functions __evaluate_image_gt_no_preds and __evaluate_image_preds_no_gt by replacing device=self.device by device=torch.device('cpu') and so far my code is working.
I think this is a critical bug, which leaves the Class unusable.

SkafteNicki:
## What does this PR do?

Fixes #943

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -121,18 +121,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed compatibility of `ClasswiseWrapper` with the `prefix` argument of `MetricCollection` ([#843](https://github.com/PyTorchLightning/metrics/pull/843))
 
 
+## [0.7.3] - 2022-03-23
+
+### Fixed
+
 - Fixed unsafe log operation in `TweedieDeviace` for power=1 ([#847](https://github.com/PyTorchLightning/metrics/pull/847))
-
-
 - Fixed bug in MAP metric related to either no ground truth or no predictions ([#884](https://github.com/PyTorchLightning/metrics/pull/884))
-
-
 - Fixed `ConfusionMatrix`, `AUROC` and `AveragePrecision` on GPU when running in deterministic mode ([#900](https://github.com/PyTorchLightning/metrics/pull/900))
-
-
 - Fixed NaN or Inf results returned by `signal_distortion_ratio` ([#899](https://github.com/PyTorchLightning/metrics/pull/899))
-
-
 - Fixed memory leak when using `update` method with tensor where `requires_grad=True` ([#902](https://github.com/PyTorchLightning/metrics/pull/902))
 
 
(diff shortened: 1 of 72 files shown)


CHANGELOG.md
```diff
@@ -121,6 +121,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed compatibility of `ClasswiseWrapper` with the `prefix` argument of `MetricCollection` ([#843](https://github.com/PyTorchLightning/metrics/pull/843))
 
 
+- Fixed device mismatch for `MAP` metric in specific cases ([]())
```
SkafteNicki, line 124:
```suggestion
- Fixed device mismatch for `MAP` metric in specific cases ([#947](https://github.com/PyTorchLightning/metrics/pull/947))
```

tests/detection/test_map.py
```diff
@@ -99,6 +99,39 @@
     ],
 )
 
+_inputs2 = Input(
```
SkafteNicki, line 102:
```suggestion
# example from this issue https://github.com/PyTorchLightning/metrics/issues/943
_inputs2 = Input(
```

justusschock (changes requested):
The repro seems to be wrong usage. This is basically the same as having a model on gpu and the corresponding inputs on cpu. This is designed not to work and should behave exactly the way it does.

Blocking this for now :)

PR #612, opened 2021-11-09
[WIP] Remove the requirement for double update in forward

tchaton:
# Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure to update the docs?
- [x] Did you write any new necessary tests?

## What does this PR do?

This PR optimizer the forward method of the Metric base class to prevent performing 2 updates.

Parts of https://github.com/PyTorchLightning/metrics/issues/344

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

Issue #344, opened 2021-07-01
Metric API re-design 


tchaton:
## 🚀 Feature

Re-design the internal API for Base Metric class to return batch states from the update function.

**The following proposal is a BC !**

### Limitations:
Currently, the Metric API is confusing for several reasons:
* The `update` function actually performs a `compute` if `compute_on_step=True`
* The `update` function actually perform 2 updates, which is an expensive operation for some metrics, FID for example
* Users don't have a clear API to perform computation
* The Metric internals is tailored to Lightning. IMO, TM should define its own API to better faciliate metrics computation and Lightning should adapt to it.

### Proposal:
* The update function doesn't return anything
* There is 2 functions for reduction, `compute` and `compute_step`
* The update function is responsible to return a dictionary containing the batch states

Here is the internal re-design of Metric 

```py
class Metric()

    def _wrap_update()
        batch_states = update(...)
        self.add_to_rank_accumulated_states(batch_states) # uses reduction function
        self.batch_states = batch_states

class MyMetric(Metric):
    def update(self) -> Dict[str, Tensor]:
        ...
        return {"state_1": state_1, ...}

metric = MyMetric()

metric(...) # compute batch states, add them to accumulated states using reduction functions
compute() # accumulated_states compute on all ranks
compute_on_step() # batch_states compute on all ranks
compute(sync_dist=False) # accumulated_states compute per rank
compute_on_step(sync_dist=False) #  batch_states compute per rank 
```

```py
class Accuracy

    def __init__(self):
        self.add_state("correct", torch.tensor(0.), sync_dist_fn=torch.sum)
        self.add_state("total", torch.tensor(0.), sync_dist_fn=torch.sum)

    def update(self, preds, targets):
        return {"total": preds.shape[0], "correct": (preds == targets).sum()}

    def compute(self):
        return self.correct / self.total


metric = Accuracy()

None = metric.update([0, 1], [0, 0])

0.5 = metric([0, 1], [0, 0], sync_dist=True) # compute batch states and cache it, add batch states to accumulated states
1 = metric([0, 0], [0, 0], sync_dist=True)

0.75 = metric.compute()
1. = metric.compute(accumulated=False)
```

```
# accumulated=True means computing accuracy on 3 batches
# accumulated=False means computing accuracy on latest batch
acc()
acc()
acc()
```

### Additional context

sync = store cache +  all_gather + reduction
unsync = restore cache

carmocca:
[...]

PR #984, opened 2022-04-25
Refactor/remove double forward

SkafteNicki:
## What does this PR do?

Redo of https://github.com/PyTorchLightning/metrics/pull/612
Fixes part of https://github.com/PyTorchLightning/metrics/issues/344 (needs review after if we can close the issue)

## Before submitting

## PR review

## Did you have fun?


Borda (approved):
can we also include in the PR the speed-up chat? :rabbit:

torchmetrics/metric.py
```diff
@@ -240,6 +237,33 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
 
         return self._forward_cache
 
+    def _reduce_state(self, state_to_reduce: Dict[str, Any]) -> None:
+        for attr in self._defaults.keys():
```
justusschock, line 241: Can we actually take care of the ordering here? So that it's always `state_to_reduce + current_state` and never the other way around? This doesn't matter for tensors or floats or something like that but it is essential for list-states as `[1] + [2] != [2] + [1]`. Personally I have sliding window states, where whenever the window size limit is hit, the first element is deleted/overwritten. Therefore ordering matters. Same with concatenation of tensors.

torchmetrics/metric.py
```diff
@@ -240,6 +237,33 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+            current_state = getattr(self, attr)
+            incoming_state = state_to_reduce[attr]
+            reduce_fn = self._reductions[attr]
+            if reduce_fn == dim_zero_sum:
```
justusschock, line 245: this means, that in my forward I can no longer do stuff like 

```python
self.sliding_window.append(value.detach())
if len(self.sliding_window) > self.sliding_window_size:
    self.sliding_window.pop(0)
```

As the sliding window size will only be correct after the reduction and not within the actual update, right? At least when using `forward`. This may make metrics harder to implement. In fact, I am not sure if all metrics are valid in their forward now as we mostly only test `update` and `compute`.

justusschock (changes requested):
This is a substantial change. While it does indeed prevent two update calls, I feel from the user experience it might be complicating the process of implementing custom metrics.

SkafteNicki:
@justusschock I completely agree that this is not a trival change and should only be done with care. 
However, our current base tester class actually calls `forward`:
https://github.com/PyTorchLightning/metrics/blob/c494f3575f542e7c264555a71fb97e8676f64700/tests/helpers/testers.py#L180
therefore, if I can get tests to pass (which they currently are not doing) I think the change is at least safe to do to our own codebase.
Does this mean that it will not break any user implementation? Not sure to be fair. If you have a particular metric that you think is a problem then consider comment it to this PR so we can check it.

We can also consider if this should be an opt-in feature. With the changes we did to the additional metric arguments, collapsing them all into the `**kwargs` arg, we can easily do something like:
```python
class Metric(nn.Module):
    def __init__(self, **kwargs):
        ...
        self.use_fast_forward = kwargs.pop('use_fast_forward', False)  # find better name, what should default be?
        ...

    def forward(self, *args, **kwargs):
        if self.use_fast_forward:
            return self.forward_method_that_only_calls_update_once(*args, **kwargs)
        else:
            return self.forward_as_it_already_is(*args, **kwargs)
 ```
@Borda any opinions?

**Comment by Borda** [comment]
@awaelchli for visibility ^^ :rabbit:

**Comment by SkafteNicki** [comment]
Feature is currently disabled for all metrics by default and then I have gone through Regression and enabled for the metrics that supports it (all except Pearson). Some notes/questions:
* Landed on the name `full_state_update`, as we talked about the key difference is not if `update` needs to be called ones or twice, the difference is if `update` need access to the full metric state accumulated until now
* If the user set `dist_sync_on_step=True` we will always do two `update` calls (old behavior). Example: if we run on two devices, 10 samples each, after compute is called the batch metric state will contain 20 samples per device. Thus, when we do reduction afterwards the global state will contain info from 40 instead of 20 samples (because all samples have been duplicated). I see no easy way around this, so to be safe I feel like the feature should be disabled in this case.
* Do you want me to continue updating all metrics in this PR or would it be easier for you (from a reviewing standpoint) if I followed up with one PR per domain?
* I cannot remember what we landed on, but should the feature be enabled by default (corresponding to `full_state_update=False`) or disabled by default?

**Comment by SkafteNicki** [comment]
Also need to run some speedup checks again to make sure that everything is still working as we expect

**Review by Borda — APPROVED** [review]
lgtm :rabbit:

File: torchmetrics/metric.py [hunk 3]
```diff
@@ -203,37 +209,59 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
         if self._is_synced:
             raise TorchMetricsUserError(
-                "The Metric shouldn't be synced when performing ``update``. "
+                "The Metric shouldn't be synced when performing ``forward``. "
                 "HINT: Did you forget to call ``unsync`` ?."
             )
 
-        # global accumulation
-        self.update(*args, **kwargs)
+        if self.full_state_update or self.dist_sync_on_step:
```
justusschock, line 219: Just a question: Should we encapsulate the different paths to separate functions so that they can be modified independent of each other?

torchmetrics/metric.py
```diff
@@ -242,6 +270,35 @@ def forward(self, *args: Any, **kwargs: Any) -> Any:
...
+            incoming_state: a dict containing a metric state similar metric itself
+        """
+        for attr in self._defaults.keys():
+            local_state = getattr(self, attr)
+            global_state = incoming_state[attr]
+            reduce_fn = self._reductions[attr]
+            if reduce_fn == dim_zero_sum:
+                reduced = global_state + local_state
+            elif reduce_fn == dim_zero_mean:
+                reduced = (self._update_count * global_state + local_state) / (self._update_count + 1)
```
justusschock, line 286: shouldn't this be `((self._update_count -1) * global_state + local_state) / self._update_count` as the count was already updated by the preceeding update call for local state?

justusschock (approved):
LGTM in general, some questions though :)

torchmetrics/metric.py
SkafteNicki: done :)

torchmetrics/metric.py
SkafteNicki: yes, you are right, fixing :)

Borda:
@SkafteNicki ready to go? :tada:

SkafteNicki:
> @SkafteNicki ready to go? :tada: 

Code should be ready. Let me run some speed tests before we merge to make sure that we actually get the speedup that we expect :]

Borda:
> Code should be ready. Let me run some speed tests before we merge to make sure that we actually get the speedup that we expect :]

Sweet, could you pls include it also here in this PR? :)

SkafteNicki:
> > Code should be ready. Let me run some speed tests before we merge to make sure that we actually get the speedup that we expect :]
> 
> Sweet, could you pls include it also here in this PR? :)

Yes will do :]
Then we also have something to show for next release blog post.

SkafteNicki:
@justusschock and @Borda, finally created an updated figure:

results

The TLDR is that metrics where `update` is expensive to evaluate we can more or less cut computational time in half and for metrics where `update` is cheap it does not really matter if the feature is enabled or disabled.

<details>
  <summary>Code to create figure</summary>

```python
from time import perf_counter

import matplotlib.pyplot as plt
import numpy as np
import torch
from tqdm import tqdm

import torchmetrics

NUM_REPS = 5
NUM_CALLS = [1, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500]

metrics = [
    torchmetrics.MeanSquaredError,
    torchmetrics.CosineSimilarity,
    torchmetrics.Accuracy,
    torchmetrics.ConfusionMatrix,
    torchmetrics.StructuralSimilarityIndexMeasure,
    torchmetrics.audio.sdr.SignalDistortionRatio,
    torchmetrics.image.lpip.LearnedPerceptualImagePatchSimilarity,
    torchmetrics.SQuAD,
    torchmetrics.WordErrorRate,
    torchmetrics.AUROC,
]

metric_args = [{}, {}, {"num_classes": 10}, {"num_classes": 10}, {}, {}, {"net_type": "alex"}, {}, {}, {"num_classes": 3}]

inputs = [
    (torch.randn(100,), torch.randn(100,)),
    (torch.randn(100,), torch.randn(100,)),
    (torch.randn(100, 10).softmax(dim=-1), torch.randint(10, (100,))),
    (torch.randn(100, 10).softmax(dim=-1), torch.randint(10, (100,))),
    (torch.rand(5, 3, 25, 25), torch.rand(5, 3, 25, 25)),
    (torch.randn(1, 8000), torch.randn(1, 8000)),
    (torch.rand(1, 3, 32, 32), torch.rand(1, 3, 32, 32)),
    ([{"prediction_text": "1976", "id": "56e10a3be3433e1400422b22"}], [{"answers": {"answer_start": [97], "text": ["1976"]}, "id": "56e10a3be3433e1400422b22"}]),
    (["this is the prediction", "there is an other sample"], ["this is the reference", "there is another one"]),
    (torch.tensor([[0.90, 0.05, 0.05], [0.05, 0.90, 0.05], [0.05, 0.05, 0.90], [0.85, 0.05, 0.10], [0.10, 0.10, 0.80]]), torch.tensor([0, 1, 1, 2, 2]))
]


def get_metric_classes(base_class):
    class Old(base_class):
        full_state_update = True

    class New(base_class):
        full_state_update = False

    return [Old, New]


if __name__ == "__main__":
    res = {True: {}, False: {}}
    for base_metric_class, metric_args, args in zip(metrics, metric_args, inputs):
        print(f"Testing {base_metric_class}")
        name = base_metric_class.__name__
        OldClass, NewClass = get_metric_classes(base_metric_class)
        for metric, enabled in zip([OldClass(**metric_args), NewClass(**metric_args)], [False, True]):
            res[enabled][name] = np.zeros((len(NUM_CALLS), NUM_REPS))
            for i, s in tqdm(enumerate(NUM_CALLS), total=len(NUM_CALLS)):
                for r in range(NUM_REPS):
                    start = perf_counter()
                    for _ in range(s):
                        val = metric(*args)
                    end = perf_counter()
                    metric.reset()
                    res[enabled][name][i, r] = end - start

    fig, ax = plt.subplots(nrows=2, ncols=5)
    for count, metric in enumerate(metrics):
        i = count % 2
        j = count % 5
        name = metric.__name__
        mean_old = res[False][name].mean(axis=-1)
        std_old = res[False][name].std(axis=-1)
        mean_new = res[True][name].mean(axis=-1)
        std_new = res[True][name].std(axis=-1)

        ax[i, j].plot(NUM_CALLS, mean_old, label="Old standard")
        ax[i, j].fill_between(NUM_CALLS, mean_old + std_old, mean_old - std_old, alpha=0.1)

        ax[i, j].plot(NUM_CALLS, mean_new, label="New standard")
        ax[i, j].fill_between(NUM_CALLS, mean_new + std_new, mean_new - std_new, alpha=0.1)

        if i == 1:
            ax[i, j].set_xlabel("Number of forward calls", fontsize=10)
        if j == 0:
            ax[i, j].set_ylabel("Time (sec)", fontsize=10)
        ax[i, j].set_title(name, fontsize=10, fontweight='bold')
        plt.setp(ax[i, j].get_yticklabels(), fontsize=5)
        ax[i, j].legend(loc="upper left", fontsize=10)

    plt.show()

```


</details>

Borda (approved):
Sweet!

Issue #992, opened 2022-04-27
Could you implement Dice score as metric  


alqurri77:
I notice Dice score is implemented as functional, but it could not be used as metric (like accuracy metric example). Could you please implement Dice score as metric.

Borda:
@alqurri77 good point. Are you interested in sending such PR?

alqurri77:
Do you mean creating a pull request. I don't I think have enough background to do this.

Borda:
> Do you mean creating a pull request. I don't I think have enough background to do this.

it is not hard, pls have look at [Quick Open-Source Contribution Guide](https://devblog.pytorchlightning.ai/quick-contribution-guide-86d977171b3a)

MrShevan:
Hi, @Borda @alqurri77 !
If it is still unresolved, could you assign it to me?)

PR #1021, opened 2022-05-10
Dice score as metric

MrShevan:
## What does this PR do?

Fixes https://github.com/PyTorchLightning/metrics/issues/992

## Before submitting

## PR review

## Did you have fun?


torchmetrics/functional/classification/dice.py
```diff
@@ -111,3 +114,208 @@ def dice_score(
...
+
+    Example:
+        >>> from torchmetrics.functional import recall
+        >>> preds  = torch.tensor([2, 0, 2, 1])
+        >>> target = torch.tensor([1, 1, 2, 0])
+        >>> recall(preds, target, average='micro')
+        tensor(0.3333)
+
+    """
+    allowed_average = ["micro", "macro", "weighted", "samples", "none", None]
```
justusschock, line 283:
```suggestion
    allowed_average = ("micro", "macro", "weighted", "samples", "none", None)
```

torchmetrics/functional/classification/dice.py
```diff
@@ -111,3 +114,208 @@ def dice_score(
...
+        raise ValueError(f"The `top_k` should be an integer larger than 0, got {top_k}")
+
+    if not background and ignore_index is None:
+        # not compute dice for the background
+        ignore_index = 0
+    elif ignore_index is not None:
+        raise ValueError("When you set `ignore_index`, you have to set background `bg` to True.")
+
+    preds, target = _input_squeeze(preds, target)
+    reduce = "macro" if average in ["weighted", "none", None] else average
```
justusschock, line 307:
```suggestion
    reduce = "macro" if average in ("weighted", "none", None) else average
```

torchmetrics/classification/dice.py
```diff
@@ -0,0 +1,163 @@
...
+            raise ValueError(f"The `average` has to be one of {allowed_average}, got {average}.")
+
+        if not background and ignore_index is None:
+            # not compute dice for the background
+            ignore_index = 0
+        elif ignore_index is not None:
+            raise ValueError("When you set `ignore_index`, you have to set background `bg` to True.")
+
+        super().__init__(
+            reduce="macro" if average in ["weighted", "none", None] else average,
```
justusschock, line 139:
```suggestion
            reduce="macro" if average in ("weighted", "none", None) else average,
```

tests/classification/test_dice.py
```diff
@@ -11,10 +11,60 @@
...
+from tests.helpers import seed_all
+from tests.helpers.testers import MetricTester
+from torchmetrics import Dice
+from torchmetrics.functional import dice, dice_score
+from torchmetrics.functional.classification.dice import _stat_scores
+
+seed_all(42)
+
+
+def _dice_score(
```
justusschock, line 30: shouldn't this be the same as 1 - dice-dissimilarity? That one is implemented at https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.dice.html 

Could you verify if this is the same? Ideally we don't implement the metric twice, because then the implementation is likely the same :)

justusschock (commented):
Hey, looking pretty good already! Thanks a lot for your implementation!
Left some minor comments. Most of them are just to use immutable types if possible :)

The only one that concerns me a bit is the one about testing :)

justusschock:
Also: should we update the jaccard implementation to rely on this implementation? Calculating them like this should be faster than computing a whole conf map and J = D/(2-D)

cc @SkafteNicki

MrShevan:
@justusschock, May I ask you for help?) I see that all failing checks caused by `installing of dependencies`. Could this be related to using `scipy`? Should I add something in PR to avoid it?

justusschock:
@Borda mind having a look on failing audio deps?

torchmetrics/classification/dice.py
```diff
@@ -0,0 +1,156 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, Dict, Optional
+
+from torch import Tensor
+
+from torchmetrics.functional.classification.dice import _dice_compute
+
+from torchmetrics.classification.stat_scores import StatScores  # isort:skip
```
Borda, line 20:
```suggestion
from torchmetrics.classification.stat_scores import StatScores
```
why is this needed?

torchmetrics/functional/classification/dice.py
```diff
@@ -111,3 +114,202 @@ def dice_score(
...
+        numerator=numerator,
+        denominator=denominator,
+        weights=None if average != "weighted" else tp + fn,
+        average=average,
+        mdmc_average=mdmc_average,
+        zero_division=zero_division,
+    )
+
+
+def dice(
```
Borda, line 170: can we clarify the difference to `dice_score`?

MrShevan: Hi, I can clarify. `dice_score` implementation supports only 'macro' reduction, which means that metric calculates for each class separately. And, by default, in `dice_score` implementation first class (which usually background class) skipped.

So `dice` includes `dice_score`, if `average='macro'` and `ignore_index=0`. I decided not to remove the old function so that users do not experience errors when updating the library version.

But I can describe all this in the documentation:)

torchmetrics/classification/dice.py
MrShevan: I copied this from [here](https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/classification/accuracy.py#L28).

torchmetrics/functional/classification/dice.py
Borda: I see so lets set `dice_score` as deprecated and use `dice` inside...
@SkafteNicki any suggestion on which name we shall keep - `dice_score` or `dice`?

justusschock: I would prefer `dice` as naming

SkafteNicki: also prefer `dice`

torchmetrics/functional/classification/dice.py
```diff
@@ -111,3 +114,202 @@ def dice_score(
...
+    zero_division: int = 0,
+    average: str = "micro",
+    mdmc_average: Optional[str] = "global",
+    threshold: float = 0.5,
+    top_k: Optional[int] = None,
+    num_classes: Optional[int] = None,
+    multiclass: Optional[bool] = None,
+    ignore_index: Optional[int] = None,
+) -> Tensor:
+    r"""Computes `Dice`_:
```
SkafteNicki, line 182: a link needs to be inserted here:
https://github.com/PyTorchLightning/metrics/blob/master/docs/source/links.rst

torchmetrics/functional/classification/dice.py
```diff
@@ -111,3 +114,202 @@ def dice_score(
...
+
+    .. math:: \text{Precision} = \frac{\text{2 * TP}}{\text{2 * TP} + \text{FP} + \text{FN}}
```
SkafteNicki, line 184: should be updated to reflect the correct computation

torchmetrics/classification/dice.py
```diff
@@ -0,0 +1,155 @@
...
+from torch import Tensor
+
+from torchmetrics.classification.stat_scores import StatScores
+from torchmetrics.functional.classification.dice import _dice_compute
+
+
+class Dice(StatScores):
+    r"""Computes `Dice`_:
+
+    .. math:: \text{Precision} = \frac{\text{2 * TP}}{\text{2 * TP} + \text{FP} + \text{FN}}
```
SkafteNicki, line 25: same here

torchmetrics/functional/classification/dice.py
MrShevan: But, it's updated. Taken from [here](https://en.wikipedia.org/wiki/Sørensen–Dice_coefficient):
Dice = 2TP / (2TP + FP + FN)

torchmetrics/functional/classification/dice.py
MrShevan: Good, I'll do today.

torchmetrics/functional/classification/dice.py
justusschock: But it says Precision = ...

MrShevan: ooh, sorry my fault ... I'll fix it now

torchmetrics/functional/classification/dice.py
MrShevan: @Borda Hi, I renamed all `dice_scores` files to `dice`, should I add 'will be deprecated' to docstring in `dice_score` or something else?

docs/source/classification/dice.rst
```diff
@@ -0,0 +1,33 @@
...
+
+.. autofunction:: torchmetrics.functional.dice
+    :noindex:
+
+
+##########
+Dice Score
+##########
+
+Functional Interface (will be deprecated)
```
MrShevan, line 29: @Borda, @SkafteNicki, @justusschock I could not find an error in testing documentation, it turned out that the problem is that Sphinx won't let you use `Functional Interface` twice, so I added `(will be deprecated)`.

But maybe there are better ways to rename `Functional Interface` for `Dice Score`?

justusschock:
One more thing left to do from my side and then it's ready to go :)

Can you raise a deprecationwarning in the older implementation asking users to switch to the newer one?

MrShevan:
@justusschock thank you for your comment :) I've added deprecation warning.

torchmetrics/functional/classification/dice.py
```diff
@@ -97,6 +101,9 @@ def dice_score(
         >>> dice_score(pred, target)
         tensor(0.3333)
     """
+    rank_zero_warn(
+        "The `dice_score` function was deprecated in v0.9 and will be removed in v1.0. Use `dice` function instead."
+    )
```
justusschock, line 106:
```suggestion
    rank_zero_warn(
        "The `dice_score` function was deprecated in v0.9 and will be removed in v1.0. Use `dice` function instead.", DeprecationWarning
    )
```

justusschock (approved):

torchmetrics/functional/classification/dice.py
```diff
@@ -94,6 +101,10 @@ def dice_score(
         >>> dice_score(pred, target)
         tensor(0.3333)
     """
+    rank_zero_warn(
+        "The `dice_score` function was deprecated in v0.9 and will be removed in v1.0. Use `dice` function instead.",
+        DeprecationWarning,
+    )
     num_classes = preds.shape[1]
```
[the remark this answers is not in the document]
MrShevan: You mean calling `dice` function below, but with parameters, which make output identical to old version `dice_score`?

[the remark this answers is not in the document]
MrShevan: I replaced old logic with new `dice` function with specified parameters into `dice_score`. But it has limitations: `dice` not supports changing `no_fg_score`, but by default `dice` calculates as if this parameter exists and has default value for `dice_score`. So, I have converted it to default if it is different.

The same with parameter `reduction`.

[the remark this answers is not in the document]
Borda: > You mean calling `dice` function below, but with parameters, which make output identical to old version `dice_score`?

Exactly

Borda (approved):

justusschock:
@Borda there are plenty of tests that did not run on this PR. Why did you merge it like that?

MrShevan:
@justusschock Did I forget something? I can open a new one and add the necessary tests. This will be inside v0.9

MrShevan:
Oh, i got it :) thanks for all your advice @Borda @justusschock @SkafteNicki! I get a lot of experience working on your project

Borda:
> Why did you merge it like that?

I did auto-merge so all suppose to be fine at that time... or what I missed?

justusschock:
Almost all our CI configs are not strictly required, but they didn't even run (canceled for some reason)

PR #1033, opened 2022-05-16
forward upgrade utils

SkafteNicki:
## What does this PR do?

Follow up to #984 
* Change the default of the new `full_state_update` property to `None`. It will still works as `full_state_update=True` (old behavior)   but a warning will be given to the user that they should consider setting the flag.
* Add utility tool for checking if it is save to set the flag
* Adds some tests

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -17,6 +17,7 @@ from torch import tensor
 
 from torchmetrics import MeanSquaredError, PearsonCorrCoef
 from torchmetrics.utilities import check_forward_no_full_state, rank_zero_debug, rank_zero_info, rank_zero_warn
+from torchmetrics.utilities.checks import _allclose_recursive
 from torchmetrics.utilities.data import _bincount, _flatten, _flatten_dict, to_categorical, to_onehot
 from torchmetrics.utilities.distributed import class_reduce, reduce
 
@@ -137,3 +138,18 @@ def test_check_full_state_update_fn(metric_class, expected):
         input_args=dict(preds=torch.randn(100), target=torch.randn(100)),
     )
     assert out == expected
+
+
+@pytest.mark.parametrize("input, expected",
+    [
+        ((torch.ones(2,), torch.ones(2,)), True), 
+        ((torch.rand(2,), torch.rand(2,)), False),
+        (([torch.ones(2,) for _ in range(2)], [torch.ones(2,) for _ in range(2)]), True),
+        (([torch.rand(2,) for _ in range(2)], [torch.rand(2,) for _ in range(2)]), False),
+        (({f'{i}' : torch.ones(2,) for i in range(2)}, {f'{i}' :torch.ones(2,) for i in range(2)}), True),
+        (({f'{i}' : torch.rand(2,) for i in range(2)}, {f'{i}' :torch.rand(2,) for i in range(2)}), False),
+    ]
+)
+def test_recursive_allclose(input, expected):
+    res = _allclose_recursive(*input)
+    assert res == expected
--- a/tests/test_utilities.py
+++ b/tests/test_utilities.py
@@ -143,96 +143,12 @@ def test_check_full_state_update_fn(metric_class, expected):
 @pytest.mark.parametrize(
     "input, expected",
     [
-        (
-            (
-                torch.ones(
-                    2,
-                ),
-                torch.ones(
-                    2,
-                ),
-            ),
-            True,
-        ),
-        (
-            (
-                torch.rand(
-                    2,
-                ),
-                torch.rand(
-                    2,
-                ),
-            ),
-            False,
-        ),
-        (
-            (
-                [
-                    torch.ones(
-                        2,
-                    )
-                    for _ in range(2)
-                ],
-                [
-                    torch.ones(
-                        2,
-                    )
-                    for _ in range(2)
-                ],
-            ),
-            True,
-        ),
-        (
-            (
-                [
-                    torch.rand(
-                        2,
-                    )
-                    for _ in range(2)
-                ],
-                [
-                    torch.rand(
-                        2,
-                    )
-                    for _ in range(2)
-                ],
-            ),
-            False,
-        ),
-        (
-            (
-                {
-                    f"{i}": torch.ones(
-                        2,
-                    )
-                    for i in range(2)
-                },
-                {
-                    f"{i}": torch.ones(
-                        2,
-                    )
-                    for i in range(2)
-                },
-            ),
-            True,
-        ),
-        (
-            (
-                {
-                    f"{i}": torch.rand(
-                        2,
-                    )
-                    for i in range(2)
-                },
-                {
-                    f"{i}": torch.rand(
-                        2,
-                    )
-                    for i in range(2)
-                },
-            ),
-            False,
-        ),
+        ((torch.ones(2), torch.ones(2)), True),
+        ((torch.rand(2), torch.rand(2)), False),
+        (([torch.ones(2) for _ in range(2)], [torch.ones(2) for _ in range(2)]), True),
+        (([torch.rand(2) for _ in range(2)], [torch.rand(2) for _ in range(2)]), False),
+        (({f"{i}": torch.ones(2) for i in range(2)}, {f"{i}": torch.ones(2) for i in range(2)}), True),
+        (({f"{i}": torch.rand(2) for i in range(2)}, {f"{i}": torch.rand(2) for i in range(2)}), False),
     ],
 )
 def test_recursive_allclose(input, expected):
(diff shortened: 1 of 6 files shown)


tests/test_utilities.py
```diff
@@ -126,3 +127,20 @@ def test_bincount():
...
+    out = check_forward_no_full_state(
+        metric_class=metric_class,
+        input_args={
+            "preds": torch.randn(
+                100,
+            ),
+            "target": torch.randn(
+                100,
+            ),
+        },
```
Borda, line 144:
```suggestion
        input_args=dict(
            preds=torch.randn(100),
            target=torch.randn(100)
        ),
```

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,117 @@ def _check_retrieval_target_and_prediction_types(
...
+
+
+def _allclose_recursive(res1: Any, res2: Any, atol: float = 1e-8):
+    """Utility function for recursively asserting that two results are within a certain tolerance."""
+    # single output compare
+    if isinstance(res1, Tensor):
+        return torch.allclose(res1, res2, atol=atol)
+    elif isinstance(res1, Sequence):
+        return all(_allclose_recursive(r1, r2) for r1, r2 in zip(res1, res2))
+    elif isinstance(res1, Dict):
```
justusschock, line 617:
```suggestion
    elif isinstance(res1, Mapping):
```

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,117 @@ def _check_retrieval_target_and_prediction_types(
...
 
     return preds.flatten(), target.flatten()
+
+
+def _allclose_recursive(res1: Any, res2: Any, atol: float = 1e-8):
+    """Utility function for recursively asserting that two results are within a certain tolerance."""
+    # single output compare
+    if isinstance(res1, Tensor):
+        return torch.allclose(res1, res2, atol=atol)
+    elif isinstance(res1, Sequence):
```
justusschock, line 615:
```suggestion
    elif isinstance(res1, str):
        return res1 == res2
    elif isinstance(res1, Sequence):
```

special-casing strings here since they are also valid Sequences so you would run in infinite recursion issues :)

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,117 @@ def _check_retrieval_target_and_prediction_types(
...
+        return all(_allclose_recursive(r1, r2) for r1, r2 in zip(res1, res2))
+    elif isinstance(res1, Dict):
+        return all(_allclose_recursive(res1[k], res2[k]) for k in res1.keys())
+    else:
+        raise ValueError("Unknown format for comparison")
```
justusschock, line 620:
```suggestion
        return res1 == res2
```

will automatically raise errors if the equality cannot be assessed for these types

torchmetrics/metric.py
```diff
@@ -127,6 +127,19 @@ def __init__(
...
 
+        if self.full_state_update is None:
+            rank_zero_warn(
+                f"""Torchmetrics v0.9 introduced a new argument class property called ``full_state_update`` that has
+                not been set for this class ({self.__class__.__name__}). The property determines if ``update`` by
+                default needs access to the full metric state. If this is not the case, significant speedups can be
+                achived. We provide an checking function
+                ``from torchmetrics.utilities import check_forward_no_full_state``
+                that can be used to check if the ``full_state_update=True`` (old and potential slower behaviour,
+                default for now) or if ``full_state_update=False`` can be used safely
```
justusschock, line 138:
```suggestion
                default for now) or if ``full_state_update=False`` can be used safely.
```

torchmetrics/metric.py
```diff
@@ -127,6 +127,19 @@ def __init__(
         self._is_synced = False
         self._cache: Optional[Dict[str, Union[List[Tensor], Tensor]]] = None
 
+        if self.full_state_update is None:
+            rank_zero_warn(
+                f"""Torchmetrics v0.9 introduced a new argument class property called ``full_state_update`` that has
+                not been set for this class ({self.__class__.__name__}). The property determines if ``update`` by
+                default needs access to the full metric state. If this is not the case, significant speedups can be
+                achived. We provide an checking function
```
justusschock, line 135:
```suggestion
                achieved and we recommend setting this to ``False``.
                We provide an checking function
```

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,117 @@ def _check_retrieval_target_and_prediction_types(
     preds = preds.float()
 
     return preds.flatten(), target.flatten()
+
+
+def _allclose_recursive(res1: Any, res2: Any, atol: float = 1e-8):
```
justusschock, line 610: can we add test functions for this helper only? Without the need for the checking function below?

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,117 @@ def _check_retrieval_target_and_prediction_types(
...
+        full_state_update = True
+
+    class PartState(metric_class):
+        full_state_update = False
+
+    fullstate = FullState(**init_args)
+    partstate = PartState(**init_args)
+
+    equal = True
+    for _ in range(10):
```
justusschock, line 683:
```suggestion
    for _ in range(num_update_to_compare[0]):
```
why do you have a fixed number here?

SkafteNicki: does not really matter, just a number larger than 1 should be sufficient :]

torchmetrics/utilities/checks.py
SkafteNicki: done

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,119 @@ def _check_retrieval_target_and_prediction_types(
...
+            steps to compare over.
+        reps: number of repetitions of speedup test
+
+    Example (states in ``update`` are independent, save to set ``full_state_update=False``)
+        >>> from torchmetrics import ConfusionMatrix
+        >>> check_forward_no_full_state(
+        ...     ConfusionMatrix,
+        ...     init_args = {'num_classes': 3},
+        ...     input_args = {'preds': torch.randint(3, (10,)), 'target': torch.randint(3, (10,))},
+        ... ) # doctest: +SKIP
```
Borda, line 650: why is this skipped? if because of some rounding, lets use `+ELLIPSES`

SkafteNicki: its skipped due to the timings changing each time the test runs
Using `+ELLIPSES` you would propose the output instead looked like:
```
Full state for 10 steps took: ...
Partial state for 10 steps took: ...
```
and so on?

torchmetrics/metric.py
```diff
@@ -127,6 +127,20 @@ def __init__(
...
+            rank_zero_warn(
+                f"""Torchmetrics v0.9 introduced a new argument class property called ``full_state_update`` that has
+                not been set for this class ({self.__class__.__name__}). The property determines if ``update`` by
+                default needs access to the full metric state. If this is not the case, significant speedups can be
+                achieved and we recommend setting this to ``False``.
+                We provide an checking function
+                ``from torchmetrics.utilities import check_forward_no_full_state``
+                that can be used to check if the ``full_state_update=True`` (old and potential slower behaviour,
+                default for now) or if ``full_state_update=False`` can be used safely.
+                """,
```
Borda, line 140:
```suggestion
                f"""Torchmetrics v0.9 introduced a new argument class property called `full_state_update` that has
                not been set for this class ({self.__class__.__name__}). The property determines if `update` by
                default needs access to the full metric state. If this is not the case, significant speedups can be
                achieved and we recommend setting this to `False`.
                We provide an checking function
                `from torchmetrics.utilities import check_forward_no_full_state`
                that can be used to check if the `full_state_update=True` (old and potential slower behaviour,
                default for now) or if `full_state_update=False` can be used safely.
                """,
```
in python text (not docs) is single ` enought and easier to read

torchmetrics/utilities/checks.py
```diff
@@ -604,3 +605,119 @@ def _check_retrieval_target_and_prediction_types(
...
+    if isinstance(res1, Tensor):
+        return torch.allclose(res1, res2, atol=atol)
+    elif isinstance(res1, str):
+        return res1 == res2
+    elif isinstance(res1, Sequence):
+        return all(_allclose_recursive(r1, r2) for r1, r2 in zip(res1, res2))
+    elif isinstance(res1, Mapping):
+        return all(_allclose_recursive(res1[k], res2[k]) for k in res1.keys())
+    else:
+        return res1 == res2
```
Borda, line 622:
```suggestion
    return res1 == res2
```

Borda (approved):

justusschock (approved):

PR #1035, opened 2022-05-16
fix deterministic for Pytorch<1.8

### LINKED ISSUE #1034: Pytorch<1.8 not working despite requirements.txt
## 🐛 Bug

In `requirements.txt` you depend on torch>=1.3.1 but while trying to use torchmetrics with torch1.4 I encountered an error in https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/utilities/data.py#L242 because `are_deterministic_algorithms_enabled` function is only available for torch>=1.8

### To Reproduce

Install torchmetrics and torch<1.8 and run the provided script.

#### Code sample

```python
import torch
from torchmetrics.functional import confusion_matrix
x = torch.tensor([[1, 2, 3], [1, 2, 3]]).cuda()
y = torch.tensor([[1, 2, 3], [1, 2, 3]]).cuda()
confusion_matrix(x, y, num_classes=4)
```

### Expected behavior

Should run.

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): pip install torchmetrics~=0.7.2 
- Python & PyTorch Version (e.g., 1.0): python3.7 and pytorch1.4
- Any other relevant information such as OS (e.g., Linux): Ubuntu 20.04

charlielito:
In `requirements.txt` you depend on torch>=1.3.1 but while trying to use torchmetrics with torch1.4 I encountered an error in https://github.com/PyTorchLightning/metrics/blob/master/torchmetrics/utilities/data.py#L242 because `are_deterministic_algorithms_enabled` function is only available for torch>=1.8

## What does this PR do?

Fixes #1034 

## Before submitting

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 numpy>=1.17.2
-torch>=1.8.0
+torch>=1.3.1
 packaging
 typing-extensions
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 numpy>=1.17.2
-torch>=1.3.1
+torch>=1.8.0
 packaging
--- a/torchmetrics/utilities/data.py
+++ b/torchmetrics/utilities/data.py
@@ -16,12 +16,15 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Union
 import torch
 from torch import Tensor, tensor
 
-from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_8
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_7, _TORCH_GREATER_EQUAL_1_8
 
 if _TORCH_GREATER_EQUAL_1_8:
     deterministic = torch.are_deterministic_algorithms_enabled
-else:
+elif _TORCH_GREATER_EQUAL_1_7:
     deterministic = torch.is_deterministic
+else:
+    # Assume for older versions that computations need to be deterministic
+    deterministic = lambda: True
 
 METRIC_EPS = 1e-6
 
--- a/torchmetrics/utilities/data.py
+++ b/torchmetrics/utilities/data.py
@@ -16,15 +16,16 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Union
 import torch
 from torch import Tensor, tensor
 
-from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_7, _TORCH_GREATER_EQUAL_1_8
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_6, _TORCH_GREATER_EQUAL_1_7, _TORCH_GREATER_EQUAL_1_8
 
 if _TORCH_GREATER_EQUAL_1_8:
     deterministic = torch.are_deterministic_algorithms_enabled
 elif _TORCH_GREATER_EQUAL_1_7:
     deterministic = torch.is_deterministic
+elif _TORCH_GREATER_EQUAL_1_6:
+    deterministic = torch._is_deterministic
 else:
-    # Assume for older versions that computations need to be deterministic
-    deterministic = lambda: True
+    deterministic = lambda: False
 
 METRIC_EPS = 1e-6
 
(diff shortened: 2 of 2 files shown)


charlielito:
It seems that before 1.8 `torch.are_deterministic_algorithms_enabled` was called`torch.is_deterministic`.  See https://github.com/pytorch/pytorch/issues/58096
Would it be better to keep pytorch>=1.3.1 compatibility and handle that scenario in this PR?

Borda (changes requested):
let's adjust the condition but do not change in version

also curious how it is possible that all tests even for PT<1.8 pass correctly
https://github.com/PyTorchLightning/metrics/runs/6447823646?check_suite_focus=true

SkafteNicki:
> let's adjust the condition but do not change in version
> 
> also curious how it is possible that all tests even for PT<1.8 pass correctly https://github.com/PyTorchLightning/metrics/runs/6447823646?check_suite_focus=true

these never evaluate the `torch.are_deterministic_algorithms_enabled` becuase `x.is_cuda` evaluates to `False` in all cases

requirements.txt
```diff
@@ -1,4 +1,4 @@
 numpy>=1.17.2
-torch>=1.3.1
+torch>=1.8.0
```
SkafteNicki, line 2:
```suggestion
torch>=1.3.1
```

torchmetrics/utilities/data.py
```diff
@@ -16,6 +16,13 @@
 import torch
 from torch import Tensor, tensor
 
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_8
+
+if _TORCH_GREATER_EQUAL_1_8:
+    deterministic = torch.are_deterministic_algorithms_enabled
+else:
+    deterministic = torch.is_deterministic
```
SkafteNicki, line 24:
```suggestion
from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_7, _TORCH_GREATER_EQUAL_1_8

if _TORCH_GREATER_EQUAL_1_8:
    deterministic = torch.are_deterministic_algorithms_enabled
elif _TORCH_GREATER_EQUAL_1_7:
    deterministic = torch.is_deterministic
else:
    # Assume for older versions that computations need to be deterministic
    deterministic = lambda: True
```

torchmetrics/utilities/data.py
```diff
@@ -16,6 +16,16 @@
 import torch
 from torch import Tensor, tensor
 
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_7, _TORCH_GREATER_EQUAL_1_8
+
+if _TORCH_GREATER_EQUAL_1_8:
+    deterministic = torch.are_deterministic_algorithms_enabled
+elif _TORCH_GREATER_EQUAL_1_7:
+    deterministic = torch.is_deterministic
+else:
```
charlielito, line 25:
```suggestion
elif _TORCH_GREATER_EQUAL_1_7:
    deterministic = torch.is_deterministic
elif _TORCH_GREATER_EQUAL_1_6:
    deterministic = torch._is_deterministic
else:
```
v1.6 supports that https://github.com/pytorch/pytorch/blob/v1.6.0/torch/__init__.py#L322

torchmetrics/utilities/data.py
```diff
@@ -16,6 +16,16 @@
...
+    # Assume for older versions that computations need to be deterministic
+    deterministic = lambda: True
```
charlielito, line 27: Is this a safe assumption? AFAIK the default behavior is non-deterministic unless you set that explicitly, which for older versions there seems to be no way to do that.

Borda (approved):

torchmetrics/utilities/data.py
```diff
@@ -16,6 +16,17 @@
...
+from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_1_6, _TORCH_GREATER_EQUAL_1_7, _TORCH_GREATER_EQUAL_1_8
+
+if _TORCH_GREATER_EQUAL_1_8:
+    deterministic = torch.are_deterministic_algorithms_enabled
+elif _TORCH_GREATER_EQUAL_1_7:
+    deterministic = torch.is_deterministic
+elif _TORCH_GREATER_EQUAL_1_6:
+    deterministic = torch._is_deterministic
+else:
+    deterministic = lambda: False
```
justusschock, line 28:
```suggestion
    def deterministic() -> bool:
        return True
```

from [pep8](https://peps.python.org/pep-0008/#programming-recommendations): 
> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier

PR #1038, opened 2022-05-19
Use PyTorch's `any` implementation

alanhdu:
Python's `any` is much slower than PyTorch's `any` (especially if the data lives on the GPU and you can avoid unnecessary GPU -> CPU copy of `nans`), so this will be much faster.

### DIFF
--- a/torchmetrics/aggregation.py
+++ b/torchmetrics/aggregation.py
@@ -69,7 +69,7 @@ class BaseAggregator(Metric):
             x = torch.as_tensor(x, dtype=torch.float32, device=self.device)
 
         nans = torch.isnan(x)
-        if nans.any().item():
+        if nans.any():
             if self.nan_strategy == "error":
                 raise RuntimeError("Encounted `nan` values in tensor")
             if self.nan_strategy == "warn":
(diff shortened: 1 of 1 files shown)


Borda:
@alanhdu interesting, mind include some benchmark? :rabbit:

alanhdu:
The effect's pretty obvious -- for example, I did:
```python
sm = SumMetric()
X = torch.randn(16, 200, 10)
```
and then
```
%%time
for __ in range(50):
    sm.update(X)
print(sm.compute())
```

Before, this took 3.8 seconds to run on my laptop -- afterwards, this takes 6.31 ms (a 500x speedup). I assume this would be even more drastic if this was running on a GPU.

justusschock:
Which makes sense, as python's any requires boolean values and therefore needs to cast every tensor element to a boolean (which includes a potential GPU->CPU transfer and thereby a synchronization point.

Borda (approved):

torchmetrics/aggregation.py
```diff
@@ -69,7 +69,7 @@ def _cast_and_nan_check_input(self, x: Union[float, Tensor]) -> Tensor:
             x = torch.as_tensor(x, dtype=torch.float32, device=self.device)
 
         nans = torch.isnan(x)
-        if any(nans.flatten()):
+        if nans.any().item():
```
justusschock, line 72:
```suggestion
        if nans.any():
```

Issue #1047, opened 2022-05-25
CUDA TENSOR RuntimeERROR


ravikakaiya:
## 🐛 Bug

While Passing a cuda tensor we encounter a RuntimeERROR : Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

### To Reproduce

Run the following code.
#### Code sample
```py
from torchmetrics import PeakSignalNoiseRatio
import torch
preds = torch.rand([3, 128, 128], generator=torch.manual_seed(42))
target = preds * 0.75
preds = preds.cuda()
target =target.cuda()

psnr_loss = PeakSignalNoiseRatio()
avg_psnr = psnr_loss(preds, target)
print(avg_psnr)
```
### Expected behavior

Should return a tensor of a value. 

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): TorchMetrics version 0.8.2
- Python & PyTorch Version (e.g., 1.0): python 3.9.12 , pytorch 1.11.0
- Any other relevant information such as OS (e.g., Linux):OS : Linux

Borda:
it is expected, that you need to have also the metric on the sam device or use PL logging which does it for you... :rabbit: 
see: https://github.com/PyTorchLightning/metrics/issues/608#issuecomment-963153184

ravikakaiya:
Thank You @Borda

lsecerovic:
@Borda can you please comment how to use PL logging so that it does this for us?

Issue #917, opened 2022-03-28
Accuracy, Precision, Recall raises RuntimeError when device="cuda"


Voleraa:
## 🐛 Bug

Hi,

when I run the following code,
```py
import torch
import torchmetrics

preds  = torch.tensor([[0.6,0.9,0.3,0,0,0,0,0.6], 
                       [0.1,0.6,0.1,0,0,0,0,0]])
target = torch.tensor([[0,1,1,0,0,0,0,0],
                       [1,1,0,0,0,0,0,1]])
accuracy = torchmetrics.Accuracy()
accuracy(preds.cuda(), target.cuda())
```
I get the following RuntimeError:
```
RuntimeError                              Traceback (most recent call last)
/tmp/ipykernel_54291/197549698.py in <module>
      7                        [1,1,0,0,0,0,0,1]])
      8 accuracy = torchmetrics.Accuracy()
----> 9 accuracy(preds.cuda(), target.cuda())

/opt/conda/lib/python3.8/site-packages/torch/nn/modules/module.py in _call_impl(self, *input, **kwargs)
    887             result = self._slow_forward(*input, **kwargs)
    888         else:
--> 889             result = self.forward(*input, **kwargs)
    890         for hook in itertools.chain(
    891                 _global_forward_hooks.values(),

~/.local/lib/python3.8/site-packages/torchmetrics/metric.py in forward(self, *args, **kwargs)
    204 
    205         # global accumulation
--> 206         self.update(*args, **kwargs)
    207 
    208         if self.compute_on_step:

~/.local/lib/python3.8/site-packages/torchmetrics/metric.py in wrapped_func(*args, **kwargs)
    265             self._update_called = True
    266             with torch.set_grad_enabled(self._enable_grad):
--> 267                 return update(*args, **kwargs)
    268 
    269         return wrapped_func

~/.local/lib/python3.8/site-packages/torchmetrics/classification/accuracy.py in update(self, preds, target)
    258             # Update states
    259             if self.reduce != "samples" and self.mdmc_reduce != "samplewise":
--> 260                 self.tp += tp
    261                 self.fp += fp
    262                 self.tn += tn

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

```
This error also occurs when I use Recall and Precision.

### Environment

- OS: Linux
- Python Version: 3.8
- PyTorch Version: 1.8.1

Borda:
[...]

PR #1056, opened 2022-05-31
Better error message on wrong device

### LINKED ISSUE #1055: PearsonCorrCoef throws device when tensors are on cuda device
### To Reproduce

```pthon
m = PearsonCorrCoef()
t1 = torch.tensor(range(10), device="cuda", dtype=float)
t2 = 2 * torch.tensor(range(10), device="cuda", dtype=float)
```
### Expected behavior

Compite pesron corr coef

- TorchMetrics version: installed with pip , verison 0.8.2

SkafteNicki:
## What does this PR do?

Fixes #1055, #1047, #917 (And probably a lot more)
Common mistake by users is not moving the metric object to the same device as the input. I think this already pretty clear from the [documentation](https://torchmetrics.readthedocs.io/en/stable/pages/overview.html#metrics-and-devices) (and the fact that metrics are just `nn.Module` which also needs to be moved) but users keep making it. This is a proposed solution that can hopefully catch this mistake.

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/torchmetrics/metric.py
+++ b/torchmetrics/metric.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import functools
 import inspect
-import traceback
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from copy import deepcopy
@@ -382,13 +381,12 @@ class Metric(Module, ABC):
                     update(*args, **kwargs)
                 except RuntimeError as err:
                     if "Expected all tensors to be on" in str(err):
-                        traceback.print_exc()
                         raise RuntimeError(
                             f"{str(err)}. \n"
                             "This could be due to the metric class not being on the same device as input.\n"
                             "Instead of `metric=Metric()` try to do `metric=Metric().to(device)` where"
                             " device corresponds to the device of the input."
-                        )
+                        ) from err
                     raise err
 
             if self.compute_on_cpu:
--- a/torchmetrics/metric.py
+++ b/torchmetrics/metric.py
@@ -383,8 +383,8 @@ class Metric(Module, ABC):
                 except RuntimeError as err:
                     if "Expected all tensors to be on" in str(err):
                         raise RuntimeError(
-                            f"{str(err)}."
-                            "Encountered different devices in metric calculation."
+                            "Encountered different devices in metric calculation"
+                            " (see stacktrace for details)."
                             "This could be due to the metric class not being on the same device as input."
                             f"Instead of `metric={self.__class__.__name__}(...)` try to do"
                             f" `metric={self.__class__.__name__}(...).to(device)` where"
--- a/torchmetrics/metric.py
+++ b/torchmetrics/metric.py
@@ -383,9 +383,9 @@ class Metric(Module, ABC):
                 except RuntimeError as err:
                     if "Expected all tensors to be on" in str(err):
                         raise RuntimeError(
-                            f"{str(err)}. \n"
-                            "Encountered different devices in metric calculation.\n"
-                            "This could be due to the metric class not being on the same device as input.\n"
+                            f"{str(err)}."
+                            "Encountered different devices in metric calculation."
+                            "This could be due to the metric class not being on the same device as input."
                             f"Instead of `metric={self.__class__.__name__}(...)` try to do"
                             f" `metric={self.__class__.__name__}(...).to(device)` where"
                             " device corresponds to the device of the input."
--- a/torchmetrics/metric.py
+++ b/torchmetrics/metric.py
@@ -380,16 +380,16 @@ class Metric(Module, ABC):
             with torch.set_grad_enabled(self._enable_grad):
                 try:
                     update(*args, **kwargs)
-                except RuntimeError as e:
-                    if "Expected all tensors to be on" in str(e):
+                except RuntimeError as err:
+                    if "Expected all tensors to be on" in str(err):
                         traceback.print_exc()
                         raise RuntimeError(
-                            f"{str(e)}. \n"
+                            f"{str(err)}. \n"
                             "This could be due to the metric class not being on the same device as input.\n"
                             "Instead of `metric=Metric()` try to do `metric=Metric().to(device)` where"
                             " device corresponds to the device of the input."
                         )
-                    raise e
+                    raise err
 
             if self.compute_on_cpu:
                 self._move_list_states_to_cpu()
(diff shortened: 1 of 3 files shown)


torchmetrics/metric.py
```diff
@@ -377,7 +378,19 @@ def wrapped_func(*args: Any, **kwargs: Any) -> None:
...
+                except RuntimeError as e:
+                    if "Expected all tensors to be on" in str(e):
+                        traceback.print_exc()
+                        raise RuntimeError(
+                            f"{str(e)}. \n"
+                            "This could be due to the metric class not being on the same device as input.\n"
+                            "Instead of `metric=Metric()` try to do `metric=Metric().to(device)` where"
+                            " device corresponds to the device of the input."
+                        )
+                    raise e
```
Borda, line 392:
```suggestion
                try:
                    update(*args, **kwargs)
                except RuntimeError as err:
                    if "Expected all tensors to be on" in str(err):
                        traceback.print_exc()
                        raise RuntimeError(
                            f"{str(err)}. \n"
                            "This could be due to the metric class not being on the same device as input.\n"
                            "Instead of `metric=Metric()` try to do `metric=Metric().to(device)` where"
                            " device corresponds to the device of the input."
                        )
                    raise err
```
single letter var is tricky... :)

torchmetrics/metric.py
```diff
@@ -377,7 +378,19 @@ def wrapped_func(*args: Any, **kwargs: Any) -> None:
             self._computed = None
             self._update_count += 1
             with torch.set_grad_enabled(self._enable_grad):
-                update(*args, **kwargs)
+                try:
+                    update(*args, **kwargs)
+                except RuntimeError as e:
+                    if "Expected all tensors to be on" in str(e):
+                        traceback.print_exc()
```
Borda, line 385: is this needed while you raise the error below?

Borda (approved):

torchmetrics/metric.py
SkafteNicki: To still get the original traceback. It could be that the error was raised for some other reason than the metric was on the wrong device and in that case I think it is great to still have the original traceback.

carmocca: In that case, you should be using the `raise RuntimeError(...) from err` construct which will do what you intend and then you don't need to print the previous exception manually. One example in PL:

https://github.com/PyTorchLightning/pytorch-lightning/blob/7938293cd9bb27e5ad97aabc7261fb29027dc6b3/pytorch_lightning/utilities/data.py#L201

SkafteNicki: cool, will do :]

torchmetrics/metric.py
```diff
@@ -377,7 +377,18 @@ def wrapped_func(*args: Any, **kwargs: Any) -> None:
...
             with torch.set_grad_enabled(self._enable_grad):
-                update(*args, **kwargs)
+                try:
+                    update(*args, **kwargs)
+                except RuntimeError as err:
+                    if "Expected all tensors to be on" in str(err):
+                        raise RuntimeError(
+                            f"{str(err)}. \n"
+                            "This could be due to the metric class not being on the same device as input.\n"
+                            "Instead of `metric=Metric()` try to do `metric=Metric().to(device)` where"
```
carmocca, line 387: This should use self.__class__.__name__

carmocca (approved):

awaelchli (approved):

torchmetrics/metric.py
```diff
@@ -377,7 +377,19 @@ def wrapped_func(*args: Any, **kwargs: Any) -> None:
...
+                try:
+                    update(*args, **kwargs)
+                except RuntimeError as err:
+                    if "Expected all tensors to be on" in str(err):
+                        raise RuntimeError(
+                            "Encountered different devices in metric calculation.\n"
+                            "This could be due to the metric class not being on the same device as input.\n"
+                            f"Instead of `metric={self.__class__.__name__}(...)` try to do"
+                            f" `metric={self.__class__.__name__}(...).to(device)` where"
+                            " device corresponds to the device of the input."
```
justusschock, line 389: can we still include the old error message here to give the actual devices to see where the mismatch is from?

SkafteNicki: done :]

justusschock (approved):

stancld (approved):

torchmetrics/metric.py
carmocca: @justusschock In a previous comment, I advocated for removing this because it's already shown by the `raise ... from ...` construct. Now it's duplicated:


```python
Traceback (most recent call last):
  File "/home/carmocca/git/pytorch-lightning/kkperf.py", line 5, in <module>
    raise RuntimeError("Expected all tensors to be on")
RuntimeError: Expected all tensors to be on

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/carmocca/git/pytorch-lightning/kkperf.py", line 8, in <module>
    raise RuntimeError(
RuntimeError: Expected all tensors to be on. 
Encountered different devices in metric calculation.
This could be due to the metric class not being on the same device as input.
Instead of `metric=FooMetric(...)` try to do `metric=FooMetric(...).to(device)` where device corresponds to the device of the input.
```

justusschock: But with the stacktrace it doesn't show the actual devices, does it?

carmocca: It does:

```python
name = "FooMetric"
import torch

if __name__ == "__main__":
    try:
        torch.tensor(1, device="cuda:1") +  torch.tensor(1, device="cuda:0")
    except RuntimeError as err:
        if "Expected all tensors to be on" in str(err):
            raise RuntimeError(
                "Encountered different devices in metric calculation.\n"
                "This could be due to the metric class not being on the same device as input.\n"
                f"Instead of `metric={name}(...)` try to do"
                f" `metric={name}(...).to(device)` where"
                " device corresponds to the device of the input."
            ) from err
        raise
```

```python
Traceback (most recent call last):
  File "pl_examples/bug_report/bug_report_model.py", line 6, in <module>
    torch.tensor(1, device="cuda:1") +  torch.tensor(1, device="cuda:0")
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "pl_examples/bug_report/bug_report_model.py", line 9, in <module>
    raise RuntimeError(
RuntimeError: Encountered different devices in metric calculation.
This could be due to the metric class not being on the same device as input.
Instead of `metric=FooMetric(...)` try to do `metric=FooMetric(...).to(device)` where device corresponds to the device of the input.
```

PR #1058, opened 2022-06-01
Fix v0.9 integration with PL

### LINKED ISSUE #1064: UserWarning at functional confusion_matrix function
## 🐛 Bug

When executing the function 'confusion_matrix' of the functional module, a `UserWarning` appears:

```
UserWarning: Torchmetrics v0.9 introduced a new argument class property called `full_state_update` that has
                not been set for this class (_ResultMetric). The property determines if `update` by                                                                                                                                                                                                                            
                default needs access to the full metric state. If this is not the case, significant speedups can be                                                                                                                                                                                                            
                achieved and we recommend setting this to `False`.                                                                                             
                We provide an checking function                                                                                                                
                `from torchmetrics.utilities import check_forward_no_full_state`                                                                               
                that can be used to check if the `full_state_update=True` (old and potential slower behaviour,                                                                                                                                                                                                                 
                default for now) or if `full_state_update=False` can be used safely.
```

This call does not allow the `full_state_update` parameter to be passed, so as users we have no way to change this behaviour. So, we should not show a user warning.

### To Reproduce

Simply run the `confusion_matrix` function of the functional module

#### Code sample

`confusion_matrix = torchmetrics.functional.confusion_matrix(
            preds, targets, num_classes=len(self.classes))`

### Expected behavior

A `UserWarning` is not displayed when executing this function.

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): pip
- Python & PyTorch Version (e.g., 1.0): python3.8, pytorch 1.11
- Any other relevant information such as OS (e.g., Linux): Linux (ubuntu20.04)

SkafteNicki:
## What does this PR do?

Fixes #1064 
Removed the v0.9 warning if user override `forward` themself (like we do in PL).

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/tests/bases/test_metric.py
+++ b/tests/bases/test_metric.py
@@ -423,3 +423,18 @@ def test_warning_on_not_set_full_state_update(metric_class):
         match="Torchmetrics v0.9 introduced a new argument class property called.*",
     ):
         UnsetProperty()
+
+
+@pytest.mark.parametrize("metric_class", [DummyListMetric, DummyMetric, DummyMetricMultiOutput, DummyMetricSum])
+def test_no_warning_on_custom_forward(metric_class):
+    """If metric is using custom forward, full_state_update is irrelevant."""
+
+    class UnsetProperty(metric_class):
+        full_state_update = None
+
+        def forward(self, *args, **kwargs):
+            self.update(*args, **kwargs)
+
+    with pytest.warns(None) as record:
+        UnsetProperty()
+    assert len(record) == 0
--- a/tests/bases/test_metric.py
+++ b/tests/bases/test_metric.py
@@ -25,6 +25,7 @@ from torch.nn import Module
 
 from tests.helpers import seed_all
 from tests.helpers.testers import DummyListMetric, DummyMetric, DummyMetricMultiOutput, DummyMetricSum
+from tests.helpers.utilities import no_warning_call
 from torchmetrics.utilities.imports import _TORCH_LOWER_1_6
 
 seed_all(42)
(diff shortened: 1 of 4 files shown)


Borda (approved):

torchmetrics/metric.py
```diff
@@ -40,6 +40,13 @@ def jit_distributed_available() -> bool:
     return torch.distributed.is_available() and torch.distributed.is_initialized()
 
 
+def is_overridden(method: str, instance: object, parent: object) -> bool:
```
carmocca, line 43: In PL we have a more complex implementation that also supports mocks and partials. Considerin copying it: https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/utilities/model_helpers.py#L21

tests/bases/test_metric.py
```diff
@@ -423,3 +423,18 @@ class UnsetProperty(metric_class):
...
+
+    class UnsetProperty(metric_class):
+        full_state_update = None
+
+        def forward(self, *args, **kwargs):
+            self.update(*args, **kwargs)
+
+    with pytest.warns(None) as record:
+        UnsetProperty()
+    assert len(record) == 0
```
carmocca, line 440: We have a nicer utility for testing the lack of warnings in PL: https://github.com/PyTorchLightning/pytorch-lightning/blob/7938293cd9bb27e5ad97aabc7261fb29027dc6b3/tests/helpers/utils.py#L118 in case you want to copy and use it.

carmocca (approved):
LGTM. Minor comments about utilities

torchmetrics/metric.py
SkafteNicki: thanks :]

tests/bases/test_metric.py
SkafteNicki: thanks :]

awaelchli (approved):

tcoln:
The warnings apper in v0.92 again.

Issue #1061, opened 2022-06-02
Missing documentation for mask format for segmentation IOU type in mAP detection metric


JeroenDelcour:
The [documentation](https://torchmetrics.readthedocs.io/en/v0.9.0/detection/mean_average_precision.html) seems to be missing the required format for masks when using the "segm" option for `iou_type`.

I'm assuming there has to be a dict entry with key "masks", but what should the shape and dtype of the masks be? I tried a few different options, but always got cryptic errors from pycocotools. The [example script](https://github.com/PyTorchLightning/metrics/blob/567fa3405d5a60ba0a6052e57afe5a5d34c376a7/tm_examples/detection_map.py) doesn't mention masks at all. I tried to figure it out from the [unit tests](https://github.com/PyTorchLightning/metrics/blob/567fa3405d5a60ba0a6052e57afe5a5d34c376a7/tests/detection/test_map.py), but I couldn't really follow what was going on.

I'd be happy to create a PR and improve the documentation myself if someone can help me figure out the required format.

Related question: I was very much looking forward to the torchmetrics implementation of mAP with masks because I thought it meant not having to deal with the poorly documented pycocotools. I've been following #822 and saw pycocotools is still required. Besides improving documentation, are there plans for removing this dependency entirely?

SkafteNicki:
cc: @gianscarpe can you comment what the format should be then @JeroenDelcour can do the actual PR? :]

gianscarpe:
Hi! let me clarify this.
Predictions are a list of dictionaries. Each dictionary has 3 keys, "masks", "labels", and "scores". Labels and scores are the same as in bounding-boxes IOU. "masks" is a tensor of shape N_predxHxW, as the example below
```
{'masks': tensor([[[False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False],
         ...,
         [False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False]],

        [[False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False],
         ...,
         [False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False],
         [False, False, False,  ..., False, False, False]]]), 'scores': tensor([0.3180, 0.7260]), 'labels': tensor([3, 2], dtype=torch.int32)}
```
[...]

PR #1080, opened 2022-06-08
Fix mAP calculation for areas with 0 predictions

23pointsNorth:
The issue was first dicussed here: https://github.com/PyTorchLightning/metrics/issues/1061#issuecomment-1150106306

In cases of segm mAP calculation with 0 predictions, the standard `box_area` calculation is needed. This is fixed to use the updated `compute_area` method.

Sidenotes:
- if `gt_label_mask` is with len>0, accessing a python list would fail -> need to casting
- `compute_area` assumes a list of areas to be computed, thus the need to expand dims in cases a single element is selected from the array.

## What does this PR do?

Fixes part of a bug first highlighted as in #1061. 

## Before submitting

## PR review

## Did you have fun?


Borda (commented):
Can you add a test for this configuration? :rabbit:

torchmetrics/detection/mean_ap.py
```diff
@@ -482,9 +482,11 @@ def __evaluate_image_gt_no_preds(
     ) -> Dict[str, Any]:
         """Some GT but no predictions."""
         # GTs
-        gt = gt[gt_label_mask]
+        gt = np.asarray(gt)[gt_label_mask]
+        if len(gt_label_mask) == 1:
+            gt = np.expand_dims(gt, axis=0)
```
stancld, line 487: Shouldn't we use directly `torch` instead of `numpy`?

23pointsNorth: `gt` is a list like [(width, height), b"some_RLE_encoding"]. the binary part is not torch encodable + the input is later on given to pycocotools, which assumes np/python.

Happy to learn how to convert things to torch if you give an example.

23pointsNorth:
Hi @Borda ,

I am having issues running the test locally (make test assumes to have `metrics/_data/detection/instance_segmentation_inputs.json`, and those I assume are fetched seperately not through e.g. `make download-data`).

I've added what I expect to be a sensible test (0 gt, 0 predictions), but hard to evaluate if it's the right spot or if it impacts some of the numerical test you have already set.

Borda:
> I am having issues running the test locally (make test assumes to have `metrics/_data/detection/instance_segmentation_inputs.json`, and those I assume are fetched separately not through e.g. `make download-data`).

good point we will add it to our make file...
the download link is https://pl-public-data.s3.amazonaws.com/metrics/data.zip

23pointsNorth:
Thanks! I assume also in .requirements/text.txt one needs to add tensorflow (because of bert -> huggingface_hub -> keras_mixin -> tf)

```
ERROR tests/text/test_bertscore.py - ModuleNotFoundError: No module named 'tensorflow'
```

Is there a way to run just detection tests?

torchmetrics/detection/mean_ap.py
stancld: Oh, was not aware of this. In that case, sticking to `numpy` seems to fine to me :]

23pointsNorth: Turns out it was still breaking some bbox assumptions about torch.stack (it had to be np dtype=object), so final solution that I feel is actually cleaner is to do list comprehension (avoids some if statements)

Borda (approved):

23pointsNorth:
Do you folks have any pointers to when the next patch version would be released?

dcela:
Ran into the same issue, for now will build from current master!

dcela:
If there is interest in getting people just to improve documentation, I would be willing to help!

Borda:
> If there is interest in getting people just to improve documentation, I would be willing to help!

please do :rabbit:

PR #884, opened 2022-03-10
Fixed bug in computing map metric for images with either no ground tr…

mtailanian:
…uth or no prediction

## What does this PR do?

Fixes #794 

When either the prediction or the ground truth is empty, the image was not taken into account for the metric computation.
This PR fixes that, by returning proper values

## Before submitting

- [x] Was this **discussed/approved** via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/metrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [x] Did you make sure to **update the docs**?
- [x] Did you write any new **necessary tests**?

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #1097, opened 2022-06-18
MeanAveragePrecision: Skip box conversion if no boxes are present

### LINKED ISSUE #1096: MeanAveragePrecision does not handle empty predictions or ground truth boxes not in xyxy format
## 🐛 Bug

When empty ground truth or prediction `boxes` tensor are provided, `MeanAveragePrecision` crashes during the conversion to `xyxy` from other formats.

### To Reproduce

1. Have empty `boxes` tensor in ground truth or prediction
2. instantiate `MeanAveragePrecision` with `box_format='xywh'`
3. run `update` with the ground truth and prediction boxes
4. conversion error

#### Code sample

```python
import torch
from torchmetrics.detection.mean_ap import MeanAveragePrecision
gts = [
        dict(boxes=torch.Tensor([[10, 20, 15, 25]]), labels=torch.IntTensor([0])),
    ]
preds = [
    dict(boxes=torch.Tensor([]), scores=torch.Tensor([]), labels=torch.IntTensor([])),
]
metric = MeanAveragePrecision(box_format='xywh')
metric.update(preds, gts)
```

### Expected behavior

torchmetrics should skip the conversion for empty boxes.

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): `0.9.1` from `nix`
- Python & PyTorch Version (e.g., 1.0): Python 3.9, PyTorch 1.11.0

kouyk:
The `box_convert` function from torchvision expects the input to be a
Tensor[N, 4], where N > 0. Should N == 0 and in_fmt != out_fmt, `unbind`
will error out on the boxes tensor during the conversion process.

The workaround is therefore to skip the box conversion if boxes is an
empty tensor.

## What does this PR do?

Fixes #1096 

## Before submitting

## PR review

## Did you have fun?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed mAP calculation for areas with 0 predictions ([#1080](https://github.com/PyTorchLightning/metrics/pull/1080))
 
+
 - Fixed bug where avg precision state and auroc state was not merge when using MetricCollections ([#1086](https://github.com/PyTorchLightning/metrics/pull/1086))
+
+
+- Skip box conversion if no boxes are present in `MeanAveragePrecision` ([#1097](https://github.com/PyTorchLightning/metrics/pull/1097))
 
 
 ## [0.9.1] - 2022-06-08
--- a/src/torchmetrics/detection/mean_ap.py
+++ b/src/torchmetrics/detection/mean_ap.py
@@ -416,7 +416,8 @@ class MeanAveragePrecision(Metric):
 
         if self.iou_type == "bbox":
             boxes = _fix_empty_tensors(item["boxes"])
-            boxes = box_convert(boxes, in_fmt=self.box_format, out_fmt="xyxy")
+            if boxes.numel() > 0:
+                boxes = box_convert(boxes, in_fmt=self.box_format, out_fmt="xyxy")
             return boxes
         elif self.iou_type == "segm":
             masks = []
--- a/tests/unittests/detection/test_map.py
+++ b/tests/unittests/detection/test_map.py
@@ -342,12 +342,8 @@ def test_empty_preds():
     metric = MeanAveragePrecision()
 
     metric.update(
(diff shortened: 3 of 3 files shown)


Borda (dismissed):

justusschock (commented):
Programming-Wise it looks fine to me. 

What is the expected outcome if there are no boxes at all? Is there a test for that?
Also, Could you add a changelog entry?

Borda:
@SkafteNicki @twsl @tkupek pls check ^^ :rabbit:

kouyk:
> What is the expected outcome if there are no boxes at all? Is there a test for that? Also, Could you add a changelog entry?

Should there be no boxes for both predictions and ground truth, the outcome is the same as any other case, i.e. no crashes. My understanding is that `update` only stores the boxes and does no computation of the metric, hence other than ensuring that the conversion process is correct, there shouldn't be anything to verify at this step.

Even for the current tests, it is somewhat redundant to test for both empty ground truth and prediction, since the problem lies in `_get_safe_item_values` that is applicable to both.

justusschock:
what I was asking for is, if there are no boxes in any update step, what should be the output of compute?

kouyk:
> what I was asking for is, if there are no boxes in any update step, what should be the output of compute?

It depends on the scenario, it could be considered false positive (empty ground truth) or false negative (empty predictions) and affect the calculation that way. If there are totally no boxes in any given update step (or more specifically any pair of ground truth and prediction boxes), the result of compute should be exactly the same as without them based on my understanding of the mAP metric definition.

That aside, I think #884 already addressed the case where there are no ground truth or prediction boxes, the bug in question is only present when using box formats other than xyxy and requires a conversion, it doesn't change the existing behaviour since empty boxes can already be provided in xyxy format.

tkupek (approved):
Fully agree with @kouyk, this does not change the behavior, only adds a failsafe for empty boxes.

Nice tests, thanks for taking care!

LGTM

justusschock (approved):

SkafteNicki (approved):

Issue #1224, opened 2022-09-14
MS-SSIM crashes for different sized images


Queuecumber:
## 🐛 Bug

My validation set has images of many different sizes. Currently the MS-SSIM module seems to cache all of these images and then computing only at the end of validation. That's probably not a good idea for a couple of reasons. 

1. It makes the code crash when the images have different sizes which I think is a pretty common situation. 
2. Some of the images could be quite large so caching them is a good way to run out of memory

I think most people would compute SSIM over each image in the batch and sum the results over all batches and divide by the total number of images. Why isn't that being done here? Or did I miss an option for it?

### To Reproduce

1. Take two images of different sizes
2. Call update on them separately
3. Call compute

#### Code sample

```
import torch
from torchmetrics import StructuralSimilarityIndexMeasure

ssim = StructuralSimilarityIndexMeasure()

a = torch.rand(5, 3, 10, 10)
b = torch.rand(5, 3, 20, 20)

ssim.update(a, a)
ssim.update(b, b)

print(ssim.compute())
```

output: 

```
Traceback (most recent call last):
  File "repro.py", line 12, in <module>
    print(ssim.compute())
  File "/mnt/fsx-outputs-chipdesign/mehrlich/poetry_cache/metabit-wl5Ts0fA-py3.8/lib/python3.8/site-packages/torchmetrics/metric.py", line 531, in wrapped_func
    value = compute(*args, **kwargs)
  File "/mnt/fsx-outputs-chipdesign/mehrlich/poetry_cache/metabit-wl5Ts0fA-py3.8/lib/python3.8/site-packages/torchmetrics/image/ssim.py", line 117, in compute
    preds = dim_zero_cat(self.preds)
  File "/mnt/fsx-outputs-chipdesign/mehrlich/poetry_cache/metabit-wl5Ts0fA-py3.8/lib/python3.8/site-packages/torchmetrics/utilities/data.py", line 42, in dim_zero_cat
    return torch.cat(x, dim=0)
RuntimeError: Sizes of tensors must match except in dimension 0. Expected size 10 but got size 20 for tensor number 1 in the list.
```

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): git master, installed with pip
- Python & PyTorch Version (e.g., 1.0): 3.8/1.12
- Any other relevant information such as OS (e.g., Linux):

Queuecumber:
[...]

PR #1231, opened 2022-09-19
Online SSIM and MS-SSIM Computation

Queuecumber:
## What does this PR do?

Working on #1224 
Fixes https://github.com/Lightning-AI/metrics/issues/1213

This PR fixes SSIM and MS-SSIM by allowing them to compute using running stats. This will make them actually usable in real situations.

Also I fixed several bugs that I found in these modules and changed the defaults to match pytorch-msssim

All tests should pass but to be honest I'm not sure how good those tests were anyway

I verified that this matches the known-good pytorch-mssim library to ~6 decimal places with default settings which was not the case before and I will be using this code in production in the coming week to see how it does.

## Before submitting

## PR review

## Did you have fun?


Queuecumber:
I'm not sure what the best way to fix the mypy failure is, would appreciate some guidance on that. Explicitly annotating the variable as a `Tensor` didn't do it

Queuecumber:
@SkafteNicki I could use some eyes on this, I've been using it in production and it seems to work well

src/torchmetrics/image/ssim.py
```diff
@@ -55,9 +55,9 @@ class StructuralSimilarityIndexMeasure(Metric):
     Example:
         >>> from torchmetrics import StructuralSimilarityIndexMeasure
         >>> import torch
-        >>> preds = torch.rand([16, 1, 16, 16])
+        >>> preds = torch.rand([3, 3, 256, 256])
         >>> target = preds * 0.75
-        >>> ssim = StructuralSimilarityIndexMeasure()
+        >>> ssim = StructuralSimilarityIndexMeasure(data_range=1.0)
```
justusschock, line 60: why can't we stick to `None` here?

src/torchmetrics/functional/image/ssim.py
```diff
@@ -468,10 +462,10 @@ def multiscale_structural_similarity_index_measure(
 
     Example:
         >>> from torchmetrics.functional import multiscale_structural_similarity_index_measure
-        >>> preds = torch.rand([1, 1, 256, 256], generator=torch.manual_seed(42))
+        >>> preds = torch.rand([3, 3, 256, 256], generator=torch.manual_seed(42))
         >>> target = preds * 0.75
-        >>> multiscale_structural_similarity_index_measure(preds, target)
-        tensor(0.9558)
+        >>> multiscale_structural_similarity_index_measure(preds, target, data_range=1.0)
```
justusschock, line 467: why can't we stick to `None` here? that seems to be the most intuitive one for me.

justusschock (commented):
It seems strange to me, that `compute` is called in `update` (this kind of breaks our paradigm), although the implementation seems to be correct.

justusschock:
Could you also add tests against pytorch-mssim as this seems to also support batching (which is lacking in sklearn)?

src/torchmetrics/image/ssim.py
Queuecumber: I guess there's no reason why we couldn't since this is just a test case that we want to come out the same every time. Technically the range for `rand` is [0, 1) though

src/torchmetrics/functional/image/ssim.py
Queuecumber: Actually I think `None` is the least intuitive since it estimates the min and max from the data. 

If this was real data (image for example) you know what the theoretical lower bound and upper bound on the values are and you wouldnt want to estimate them from the data. Otherwise you might get different ranges over different batches and the SSIM values wouldn't be directly comparable.

It's counterintuitive because who would want that?

Queuecumber:
Yeah I will add tests against pytorch_msssim, do you want to leave the sklearn tests in place?

BTW we found an interesting issue with pytorch_msssim while testing this PR internally. It really doesn't play nice with fp16 AMP. Torchmetrics handles it much better. Irrelevant for the test cases but thought I would point it out

Queuecumber:
> It seems strange to me, that `compute` is called in `update` (this kind of breaks our paradigm), although the implementation seems to be correct.

There should be a way to refactor it so that it looks more like the PSNR metric which does an `_update` call which is computing sum of squared error and then compute is more or less just reducing it. 

I'm fine implementing that refactor if you want

Queuecumber:
Just following up, my plan is to do the following 

1. Rename `_ssim_update` to `_ssim_check_inputs` (open to suggestions on that name)
2. Rename `_ssim_compute` to `_ssim_update` and remove the reduction logic (but unsure of where to put the sum)
3. Add `_ssim_compute` which does the reduction (a little weird since in this case it's literally just calling reduce)
4. Update `structural_similarity_index_measure` to call these new functions correctly

justusschock:
> do you want to leave the sklearn tests in place?

Ideally yes, as this is the most trusted lib for non-batched stuff. Maybe we can then reduce the tests compared to sklearn and add the more thorough testing with the other lib, but some sklearn tests should stay.


> Just following up, my plan is to do the following
>
>Rename _ssim_update to _ssim_check_inputs (open to suggestions on that name)
Rename _ssim_compute to _ssim_update and remove the reduction logic (but unsure of where to put the sum)
Add _ssim_compute which does the reduction (a little weird since in this case it's literally just calling reduce)
Update structural_similarity_index_measure to call these new functions correctly


Sounds good to me!

Queuecumber:
OK I did the refactoring we discussed and added test cases for SSIM again pytorch-msssim (the MS-SSIM tests were already using it)

requirements/test.txt
```diff
@@ -14,3 +14,4 @@ fire
 
 cloudpickle>=1.3
 scikit-learn>1.0, <1.1.1
+pytorch-msssim==0.2.1
```
SkafteNicki, line 17: update this file instead:
https://github.com/Lightning-AI/metrics/blob/master/requirements/image_test.txt
```suggestion
```

SkafteNicki (approved):
LGTM

justusschock (approved):
Thanks for doing. this refactor! LGTM!

PR #1236, opened 2022-09-26
[Segmentation] Add mean IoU

### LINKED ISSUE #1124: mIoU
Where is the semantic segmentation mIou?

NielsRogge:
## What does this PR do?

This PR adds the mean Intersection over Union (mIoU) metric, especially useful for semantic segmentation (where the goal is to label each pixel of an image with a certain class).

I first tried to use the existing Jaccard Index metric for this, but it's not ideal; one needs to set `average=None`, and even then, you can't just easily calculate the mIoU as you need to take the union of the labels present in the predicted segmentation map and the ground truth segmentation map.

Hence, this PR proposes to add a new "Segmentation" section, to which metrics like mIoU and panoptic quality (PQ) can be added.

Fixes #1124 

The implementation is based on the one in [mmsegmentation](https://github.com/open-mmlab/mmsegmentation/blob/master/mmseg/core/evaluation/metrics.py) by OpenMMLab.

I just created a functional variant for now, if this gets approved I can proceed with making a module variant, as well as implementing the tests.

## Before submitting

## PR review

## Did you have fun?


src/torchmetrics/functional/segmentation/mean_iou.py
```diff
@@ -0,0 +1,233 @@
...
+    nan_to_num: Optional[int] = None,
+    label_map: Optional[Dict[int, int]] = None,
+    reduce_labels: bool = False,
+) -> Tuple[Tensor, int]:
+    """Updates and returns variables required to compute Mean Intersection over Union.
+
+    Checks for same shape of each element of the ``preds`` and ``target`` lists.
+
+    Args:
+        preds (list[torch.Tensor]):
```
justusschock, line 144: remove typehints here as you already added them above. Also: They don't match. The annotations says `torch.Tensor`, the docstring `list[torch.Tensor]`. WHich one is it?

src/torchmetrics/functional/segmentation/mean_iou.py
```diff
@@ -0,0 +1,233 @@
...
+        target (list[torch.Tensor]):
+            List of ground truth segmentation maps.
+        num_classes (int):
+            Number of categories.
+        ignore_index (int):
+            Index that will be ignored in evaluation.
+        label_map (dict):
+            Mapping old labels to new labels. Default: dict().
+        reduce_zero_label (bool):
+            Whether ignore zero label. Default: False.
```
justusschock, line 155: remove :)

src/torchmetrics/functional/segmentation/mean_iou.py
```diff
@@ -0,0 +1,233 @@
...
+        num_labels:
+            number of labels
+        ignore_index:
+            index that will be ignored in evaluation
+        nan_to_num:
+            If specified, NaN values will be replaced by the numbers defined by the user. Default: None.
+        label_map:
+            Mapping old labels to new labels. Default: None.
+        reduce_labels:
+            Whether to ignore the zero label and reduce all labels by one. Default: False.
```
justusschock, line 218: remove :)

justusschock (commented):
Hi, thanks for opening this PR. I have left a few comments:

1.) Why didn't you rely on the jaccard index as pointed out in the issue you linked?
2.) You only implemented a multiclass case. Could you also implement a binary case (see our recent classification refactor) and comment on a multilabel case?

NielsRogge:
Thanks for the quick review.

> 1.) Why didn't you rely on the jaccard index as pointed out in the issue you linked?

I made a quick Colab notebook to showcase what I'd like to achieve: https://colab.research.google.com/drive/1O8KlOdiz7JXAIKLh2TsKs11cwH5B4ZDK?usp=sharing.

I implemented the mIoU metric in HF evaluate, but as it's in NumPy it's rather slow. As can be seen, the `per_category_iou` returns `nan` for categories which occur at locations which are `ignore_index` in the ground truth.

Could we achieve this using the existing implementation? I think the current implementation doesn't take `ignore_index` into account in the same way as the implementation in `evaluate` does.

justusschock:
@NielsRogge fair enough. Just wanted to get some confirmation that we actually need this and don't implement something again we already have on a different name :)

robmarkcole:
@NielsRogge why close this?

lantiga:
Let's reopen this one. I agree that the output is not handled correctly here when `ignore_index` is provided.

justusschock (approved):

src/torchmetrics/functional/segmentation/utils.py
```diff
@@ -24,6 +24,13 @@
 from torchmetrics.utilities.imports import _SCIPY_AVAILABLE
 
 
+def _ignore_background(preds: Tensor, target: Tensor) -> Tuple[Tensor, Tensor]:
+    """Ignore the background class in the computation."""
```
Borda, line 28:
```suggestion
    """Ignore the background class in the computation assuming it is the first, ihned 0."""
```

Borda (approved):

PR #1259, opened 2022-10-08
Vectorize `_find_best_gt_match` in `MeanAveragePrecision`

### LINKED ISSUE #1024: MeanAveragePrecision is slow
## 🐛 Bug

It's extremely slow to compute the mean-average-precision since torchmetrics > 0.6.0.

### To Reproduce

I noticed that my training times have almost doubled since I upgraded torchmetrics from 0.6.0, because validation using the MAP / MeanAveragePrecision metric is so much slower. During validation steps I call `update()`, and in the end of a validation epoch I call `compute()` on the `MeanAveragePrecision` object.

I calculated the time that spent inside `compute()` with different torchmetrics versions:

- torchmetrics 0.6.0: 12 s
- torchmetrics 0.6.1: didn't work for some reason
- torchmetrics 0.6.2: 9.5 min
- torchmetrics 0.7.0: 9.4 min
- torchmetrics 0.7.1: 1.9 min
- torchmetrics 0.7.2: 2.0 min
- torchmetrics 0.7.3: 1.9 min
- torchmetrics 0.8.0: 4.5 min
- torchmetrics 0.8.1: 4.6 min
- torchmetrics 0.8.2: 4.6 min

It seems that after 0.6.0 the time to run `compute()` has increased from 10 seconds to 9.5 minutes. In 0.7.1 it was improved and took 2 minutes. Then in 0.8.0 things got worse again and it took 4.5 minutes to run `compute()`. This is more than 20x slower than with 0.6.0 and for example when training 100 epochs adds another 7 hours to the training time.

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): 0.6.0 through 0.8.2, installed using pip
- Python & PyTorch Version (e.g., 1.0): Python 3.8.11, PyTorch 1.10.0

stancld:
## What does this PR do?

Part of #1024

## Before submitting

<hr>

Not familiar with `MeanAveragePrecision`, but I conducted some measurements (on my CPU (M1)):

```python
import torch
from torchmetrics.detection.mean_ap import MeanAveragePrecision

d = dict(
    boxes=torch.rand(20,4),
    scores=torch.rand(20),
    labels=torch.randint(3, (20,)),
)
dp = dict(
    boxes=torch.rand(20,4),
    labels=torch.randint(3, (20,)),
)
preds = 2000 * [d]
target = 2000 * [dp]

mape = MeanAveragePrecision()
mape.update(preds, target)
mape.compute()
```

Before this change, `mape.compute()` takes ~17s, now it's ~6s.

old_mape

new_mape


src/torchmetrics/detection/mean_ap.py
```diff
@@ -579,10 +579,6 @@ def _evaluate_image(
         det = [det[i] for i in det_label_mask]
         if len(gt) == 0 and len(det) == 0:
             return None
-        if isinstance(det, dict):
-            det = [det]
-        if isinstance(gt, dict):
-            gt = [gt]
```
stancld, line 585: `det` and `gt` are stored as `torch.Tensor`, not `dict`. Thus this checking is redundant.

Borda (approved):

SkafteNicki (approved):

src/torchmetrics/detection/mean_ap.py
```diff
@@ -604,23 +600,23 @@ def _evaluate_image(
...
-                        continue
-                    det_ignore[idx_iou, idx_det] = gt_ignore[m]
-                    det_matches[idx_iou, idx_det] = 1
-                    gt_matches[idx_iou, m] = 1
+            best_matches = self._find_best_gt_matches(iou_thresholds, gt_matches, gt_ignore, ious)
+            _zero_tensor = torch.tensor(0, dtype=torch.bool, device=self.device)
+            _one_tensor = torch.tensor(1, dtype=torch.bool, device=self.device)
+            det_ignore = torch.where(
+                best_matches != -1, gt_ignore[best_matches.clamp(max=gt_ignore.shape[0] - 1)], _zero_tensor
+            )
```
SkafteNicki, line 616: this does not work in torch 1.7 (apparently) because the 
```RuntimeError: "where_cpu" not implemented for 'Bool'```
the simple solution is casting to float before and bool afterwards, but not sure if we should just skip 1.7
```suggestion
            det_ignore = torch.where(
                best_matches != -1, gt_ignore[best_matches.clamp(max=gt_ignore.shape[0] - 1)].float(), _zero_tensor.float()
            ).bool()
```

stancld: Oh, good catch. Maybe I can encapsulate this into a method and within that method differentiate between `torch==1.7` and other versions? Not sure how much this affects the performance (likely not much).

justusschock: At this point I think we could also drop 1.7 compatibility. It is already quite old...

justusschock (approved):
Thanks a lot for doing this @stancld this is huge!

src/torchmetrics/detection/mean_ap.py
SkafteNicki: If we settle for 1.8 being minimum we would still be compatible with 5 releases of pytorch.

justusschock: Yes, I think aiming for 5 major releases would be a good compromise (in PL we do fewer). This would also make it easier to implement new features since at some point you are quite limited with the older versions of torch

Borda: yes, I am fine with set min version 1.8, just lets do it in separate PR :)

SkafteNicki: just created issue for future reference: https://github.com/Lightning-AI/metrics/issues/1262

SkafteNicki: and pr: https://github.com/Lightning-AI/metrics/pull/1263

senarvi:
Hi @stancld , great to see that you're working on this issue!

First I tested this code with the speed test script from my issue. I got these results:

**torchmetrics 0.6.0 (pycocotools)**

Total time in init: 2.7128050999999687
Total time in update: 0.09905929999513319
Total time in compute: 6.811695100001089

**torchmetrics 0.10.0**

Total time in init: 0.0006100000027799979
Total time in update: 0.08251599999130121
Total time in compute: 167.1733286000017

**this pull request**

Total time in init: 1.2181383000024653
Total time in update: 0.07762909999655676
Total time in compute: 92.89537899999777

This pull request improves the speed a lot! Although as you said, there's still more room for improvement.

Then I compared this PR to torchmetrics 0.6.0 on real data, by running a test epoch using Lightning. torchmetrics 0.6.0 was 7x faster. Mostly the scores were quite close, but there's a very large difference in map_50 - 9 % absolute. So that's strange.

Borda:
@senarvi, would you be interested in participating in the speed/performance improvements? :rabbit:

senarvi:
@Borda, sorry, I'm currently working on two other pull requests, so I don't have time to dig into this.

Borda:
> @Borda, sorry, I'm currently working on two other pull requests, so I don't have time to dig into this.

no problem we are fine to wait, just wondering if I to reserve it for you :rabbit:

senarvi:
@Borda I really need to focus on other things.

PR #1295, opened 2022-10-27
Fix typo in LPIPS class docstring and in an error message

bryant1410:


Borda (approved):

Borda:
@stancld seems we have one more/new error message for HF model not available:
`OSError: Can't load the model for 'roberta-large'. If you were trying to load it from 'https://huggingface.co/models', make sure you don't have a local directory with the same name. Otherwise, make sure 'roberta-large' is the correct path to a directory containing a file named pytorch_model.bin, tf_model.h5, model.ckpt or flax_model.msgpack.
`

stancld:
@Borda I think we have this kind of failure handled for unit tests. Do you know if there is a way how to skip a particular doctest if this `OSError` is encountered? 🤔

Borda:
> @Borda I think we have this kind of failure handled for unit tests. Do you know if there is a way how to skip a particular doctest if this `OSError` is encountered? thinking

It shall be skipped with this:
https://github.com/Lightning-AI/metrics/blob/a42a25a5e82094b20f4782e5789d8918fdc79cbb/src/torchmetrics/image/lpip.py#L31

PR #1296, opened 2022-10-27
Make LPIPS code example use the interval [-1, 1]

bryant1410:
LPIPS input range is `[-1, 1]` by default (not `[0, 1]`). I think the example as it is now is confusing because it generates numbers in `[0, 1]`, so I changed it to be to `[-1, 1]`.

### DIFF
--- a/src/torchmetrics/image/lpip.py
+++ b/src/torchmetrics/image/lpip.py
@@ -78,7 +78,7 @@ class LearnedPerceptualImagePatchSimilarity(Metric):
         >>> _ = torch.manual_seed(123)
         >>> from torchmetrics.image.lpip import LearnedPerceptualImagePatchSimilarity
         >>> lpips = LearnedPerceptualImagePatchSimilarity(net_type='vgg')
-        >>> img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
+        >>> img1 = (torch.rand(10, 3, 100, 100) * 2) - 1  # LPIPS needs the images to be in the [-1, 1] range.
         >>> img2 = (torch.rand(10, 3, 100, 100) * 2) - 1
         >>> lpips(img1, img2)
         tensor(0.3493, grad_fn=<SqueezeBackward0>)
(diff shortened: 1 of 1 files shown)


src/torchmetrics/image/lpip.py
```diff
@@ -78,10 +78,10 @@ class LearnedPerceptualImagePatchSimilarity(Metric):
         >>> _ = torch.manual_seed(123)
         >>> from torchmetrics.image.lpip import LearnedPerceptualImagePatchSimilarity
         >>> lpips = LearnedPerceptualImagePatchSimilarity(net_type='vgg')
-        >>> img1 = torch.rand(10, 3, 100, 100)
-        >>> img2 = torch.rand(10, 3, 100, 100)
+        >>> img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
```
Borda, line 81: lets add comment with an explanation of what the -1 is for :)

bryant1410: Just added a comment

src/torchmetrics/image/lpip.py
```diff
@@ -78,10 +78,10 @@ class LearnedPerceptualImagePatchSimilarity(Metric):
         >>> _ = torch.manual_seed(123)
         >>> from torchmetrics.image.lpip import LearnedPerceptualImagePatchSimilarity
         >>> lpips = LearnedPerceptualImagePatchSimilarity(net_type='vgg')
-        >>> img1 = torch.rand(10, 3, 100, 100)
-        >>> img2 = torch.rand(10, 3, 100, 100)
+        >>> img1 = (torch.rand(10, 3, 100, 100) * 2) - 1  # LPIPS needs the images to be in the [-1, 1] range.
```
Borda, line 81:
```suggestion
        >>> # LPIPS needs the images to be in the [-1, 1] range.
        >>> img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
```

Borda (approved):

stancld (approved):

PR #1309, opened 2022-11-03
`BinaryPrecisionRecallCurve` for large datasets (>100 million samples)

jpcbertoldo:
## What does this PR do?

This PR provides an alternative to [this warning](https://github.com/Lightning-AI/metrics/blob/8b02cd55bf319cdca3488ca54228fb9c549274cd/src/torchmetrics/classification/precision_recall_curve.py#L537): 

```python
rank_zero_warn(
    "Metric `PrecisionRecallCurve` will save all targets and predictions in buffer."
    " For large datasets this may lead to large memory footprint."
)
```

### Context

Currently, `BinaryPrecisionRecallCurve` (and related methods by consequence) has two operation modes, which I will call "computed-thresholds" (arg `thresholds=None`) and "given-thresholds" (otherwise).

#### "computed-thresholds" 

All possible thresholds in the sample ares used in `compute()`. It has a high memory consumption if the number of instances is high because all the preds and targets are kept in memory during the updates. 

How much is "high" memory consumption?

I will consider 100 million samples the order of magnitude where "high" starts.
This roughly corresponds to ~750Mb of for the tensor `preds` under the "computed-thresholds" mode: `1e8 (samples) * 4 (bytes/float32) = 375Mb`. 

> In my research I'd like to compute roc curves of pixel-wise classifications with `1024 x 1024` resolution images, and I have ~2000 in test, which would account for `1024 * 1024 * 2000 * 4 (byte/float32) ~= 7.8Gb` at least .  

#### "given-thresholds"

Thresholds are pre-defined, so all possible binarizations are known in advance.  At each update all are tested, giving as many confusion matrices. 

The memory consumption is low because and a function of the number of thresholds
If `num_thresholds = 10e6`, then it uses roughly `10e6 * (2 *2) (confmat shape) * 8 (bytes/long) ~= 300mb` which is much lower even with 1 million thresholds. 
However, it requires the user to know meaningful threshols in advance, which is often not the case. 

A possible (inconvenient) solution would be to estimate the thresholds on a first call then compute the curve.

### Solution

I propose to create a hybrid strategy where the update switches from "computed-thresholds" to "given-thresholds" dinamically.

Given a `budget` (number of instances), the keeps preds and targets in the state until that budget is reached, then estimates meaningful thresholds and compute the confusion matrices at that point. From there it behaves like "given-thresholds" mode.

### API sketch

```python
# the arg `thresholds` can be a str 
# `budget` is the max number of instances seen before the mode switch
metric = BinaryPrecisionRecallCurve(thresholds="dynamic", budget=10e8)  # 100 million for instance
```

## Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/metrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?

## PR review


SkafteNicki:
Hi @jpcbertoldo, thanks for proposing this enhancement (I have also seen your thread on slack)
I do like the idea, but I wonder if it is possible to still fit this feature into the `thresholds` argument, so we do not have to change the outer interface again. We could reserve `str` input to this feature such that:
```python
metric = BinaryPrecisionRecallCurve(thresholds="12345")
metric = BinaryPrecisionRecallCurve(thresholds="250mb")
metric = BinaryPrecisionRecallCurve(thresholds="3gb")
```
all enabled this feature. What do you think about that?

jpcbertoldo:
> Hi @jpcbertoldo, thanks for proposing this enhancement (I have also seen your thread on slack) I do like the idea, but I wonder if it is possible to still fit this feature into the `thresholds` argument, so we do not have to change the outer interface again. We could reserve `str` input to this feature such that:
> 
> ```python
> metric = BinaryPrecisionRecallCurve(thresholds="12345")
> metric = BinaryPrecisionRecallCurve(thresholds="250mb")
> metric = BinaryPrecisionRecallCurve(thresholds="3gb")
> ```
> 
> all enabled this feature. What do you think about that?

Yeah makes sense. I was actually thinking of considering just the case `threshols="200mb"`, I think it makes a lot more sense than thinking about numbers of instances.

jpcbertoldo:
@SkafteNicki by the way, I just saw this 

> New Features:
> 1. Submit a github issue - describe what is the motivation of such feature (adding the use case or an example is helpful).
> 2. Let's discuss to determine the feature scope.

Could we skip this step for this time? : )

SkafteNicki:
> @SkafteNicki by the way, I just saw this
> 
> > New Features:
> > 
> > 1. Submit a github issue - describe what is the motivation of such feature (adding the use case or an example is helpful).
> > 2. Let's discuss to determine the feature scope.
> 
> Could we skip this step for this time? : )

That is normally how we like it. But since you already had implemented some, no need to open a issue about this. It is recommended because we do not want people to implement a lot before talking with us, because we may decide it is not a feature we want and therefore their work would go to waste.

src/torchmetrics/classification/precision_recall_curve.py
```diff
@@ -104,9 +129,25 @@ class BinaryPrecisionRecallCurve(Metric):
     higher_is_better: Optional[bool] = None
     full_state_update: bool = False
 
+    class _ComputationMode(Enum):
+        """Internal state of the dynamic mode."""
+
+        BINNED = "binned"
+        NON_BINNED = "non-binned"
+        NON_BINNED_DYNAMIC = "non-binned-dynamic"
```
SkafteNicki, line 137: it seems weird to me having a class inside a class def?

src/torchmetrics/classification/precision_recall_curve.py
```diff
@@ -41,6 +45,26 @@
...
+    if budget <= 0:
+        raise ValueError("Budget must be larger than 0.")
+
+    if _budget_bytes_to_nsamples(budget) <= _DYNAMIC_THRESHOLDS_MIN_NSAMPLES:
+        warnings.warn(
+            f"Budget is relatively small ({humanfriendly.format_size(budget, binary=True)}). "
+            "The dynamic mode is recommended for bigger samples."
+        )
+
+    return budget
```
SkafteNicki, line 66: if we only have the `thresholds` argument i guess all this logic can be moved to the `_adjust_threshold_arg` function

SkafteNicki (commented):
Have you made any thoughts on extending this to multiclass / multilabel?

src/torchmetrics/functional/classification/precision_recall_curve.py
```diff
@@ -14,6 +14,7 @@
 
 from typing import List, Optional, Sequence, Tuple, Union
 
+import humanfriendly
```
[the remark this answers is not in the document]
jpcbertoldo: I saw that one coming haha.

So, i put it anyway because it feels like the kind of functionality pruned to silly mistakes while a tiny library like this has it neatly packed in.

I can try to make a minimal version of it based on the library's source code. Is that a better solution?

src/torchmetrics/classification/precision_recall_curve.py
jpcbertoldo: In the multiclass/label cases the estimation (number of samples) <-> (memory consumption) would be different.

I don't have a very strong opinion on this, i will put some more thought on the multi* cases first.

jpcbertoldo:
> Have you made any thoughts on extending this to multiclass / multilabel?

Just as much as in my reply to your comment 😬. 

I'll be putting some effort on that this weekend :)

jpcbertoldo:
@SkafteNicki i'm having a hard time to make my self available to invest some time on this.
Maybe could you give me a hand and tell me what/where I have to change things to adapt it for the multilabel case?

Borda:
@stancld could you help here to finish this PR? :otter:

src/torchmetrics/classification/precision_recall_curve.py
```diff
@@ -11,14 +11,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import warnings
+from enum import Enum
 from typing import Any, List, Optional, Tuple, Union
 
+import humanfriendly
```
Borda, line 18: lets have this as optional

src/torchmetrics/classification/precision_recall_curve.py
Borda: yes please

src/torchmetrics/classification/precision_recall_curve.py
```diff
@@ -101,9 +126,25 @@ class BinaryPrecisionRecallCurve(Metric):
...
+        NON_BINNED_DYNAMIC = "non-binned-dynamic"
+
+    @staticmethod
+    def _deduce_computation_mode(thresholds: Optional[Union[int, List[float], Tensor, str]]) -> _ComputationMode:
+        if isinstance(thresholds, str):
+            return BinaryPrecisionRecallCurve._ComputationMode.NON_BINNED_DYNAMIC
+        elif thresholds is None:
+            return BinaryPrecisionRecallCurve._ComputationMode.NON_BINNED
+        else:
+            return BinaryPrecisionRecallCurve._ComputationMode.BINNED
```
Borda, line 143:
```suggestion
        if isinstance(thresholds, str):
            return BinaryPrecisionRecallCurve._ComputationMode.NON_BINNED_DYNAMIC
        if thresholds is None:
            return BinaryPrecisionRecallCurve._ComputationMode.NON_BINNED
        return BinaryPrecisionRecallCurve._ComputationMode.BINNED
```

src/torchmetrics/functional/classification/precision_recall_curve.py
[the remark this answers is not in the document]
Borda: lets make it conditional, if user already have it, then use it
```py
if module_available("humanfriendly"):
    import humanfriendly
else:
    humanfriendly = None
```

src/torchmetrics/functional/classification/precision_recall_curve.py
```diff
@@ -80,13 +81,20 @@ def _binary_clf_curve(
...
+
+    - list and int -> tensor
+    - None -> None
+    - str -> int (memory budget) in Mb
+    """
     if isinstance(thresholds, int):
         thresholds = torch.linspace(0, 1, thresholds, device=device)
     if isinstance(thresholds, list):
         thresholds = torch.tensor(thresholds, device=device)
+    if isinstance(thresholds, str):
```
Borda, line 96:
```suggestion
    if isinstance(thresholds, str) and humanfriendly:
```

Borda:
@jpcbertoldo, how is it going here? I think we are on good path... :)

jpcbertoldo:
> @jpcbertoldo, how is it going here? I think we are on good path... :)

Hi @Borda, I lost track of this since a while (vacation, other priorities...). I will try to come back to it soon!

jpcbertoldo:
@Borda I was thinking if we could do a simpler solution for this by computing the metric in 2 rounds ("epochs").

1st round: just find the min/max, then, with a given maximum number of points, linearly  -- or some smarter way ? -- space the thresholds between the min/max. 

2nd round: compute the actual TPR/FPRs in an online way (because now the thresholds are known in advance).

---

An alternative for the 1st round: instead of just min/max, keep track of unique values, which may provide information for eventually having parts of the threshold range more or less dense. The set of unique values could be in `float16` instead of `float32` to increase the redundancy of values (i'm guessing it shouldn't affect so much the accuracy of the AUC).

Borda:
@jpcbertoldo apology for the late reply...
 @SkafteNicki what do you think about the suggestion above?

Borda:
@SkafteNicki do you think we could finish this one? 🐿️

SkafteNicki:
Lets close this for now. I get that is would be nice feature to have, that torchmetrics would be able to switch between the modes, however there has not really been other users requesting this so I guess it is not a large problem in reality.

PR #1320, opened 2022-11-08
Prevent iteration over metrics

### LINKED ISSUE #1319: Iteration over a metric never ends
## 🐛 Bug
If you try to iterate over a metric (by mistake in my case), it just spins forever, making it appear that the program is hung. I'm not sure internally to torchmetrics exactly why this behavior occurs and if it is desired, but it would be nice if it errored out or something, unless there is a use for iterating over a metric that I am not aware of?

```py
from torchmetrics.classification import Accuracy
acc = Accuracy()
for i, item in enumerate(acc):
    print(i, item)
```

The above code will spin forever, printing something like this
```
8144 CompositionalMetric(
  <lambda>(
    Accuracy(),
    None
  )
)
```

### Expected behavior
I would expect an error to be raised if I try to iterate over a metric.

### Environment
- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source):  `0.10.2`

SkafteNicki:
## What does this PR do?

Fixes #1319 
Prevents users from accidentally end up in a infinite loop if they try to iterate over a single metric.
The reason currently is that we implement the `__getitem__` method to create compositional metrics for indexing and we therefore need to explicit prevent iteration.

## Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/Lightning-AI/metrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?

## PR review

## Did you have fun?

### DIFF
--- a/src/torchmetrics/metric.py
+++ b/src/torchmetrics/metric.py
@@ -858,6 +858,9 @@ class Metric(Module, ABC):
     def __getnewargs__(self) -> Tuple:
         return (Metric.__str__(self),)
 
+    def __iter__(self):
+        raise NotImplementedError("Metrics does not support iteration.")
+
 
 def _neg(x: Tensor) -> Tensor:
     return -torch.abs(x)
--- a/tests/unittests/bases/test_metric.py
+++ b/tests/unittests/bases/test_metric.py
@@ -464,3 +464,10 @@ def test_custom_availability_check_and_sync_fn():
     acc.compute()
     dummy_availability_check.assert_called_once()
     assert dummy_dist_sync_fn.call_count == 4  # tp, fp, tn, fn
+
+
+def test_no_iteration_allowed():
+    metric = DummyMetric()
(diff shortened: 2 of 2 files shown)


src/torchmetrics/metric.py
```diff
@@ -858,6 +858,9 @@ def __getitem__(self, idx: int) -> "Metric":
     def __getnewargs__(self) -> Tuple:
         return (Metric.__str__(self),)
 
+    def __iter__(self):
```
justusschock, line 861: this dies not apply to metric collection right?

justusschock (approved):

src/torchmetrics/metric.py
SkafteNicki: No, because metric collection is actually not a metric:
https://github.com/Lightning-AI/metrics/blob/a3dc40c89e79622474290e8813ff27b92ffd8268/src/torchmetrics/collections.py#L29

stancld (approved):

Borda (approved):

PR #1328, opened 2022-11-10
Feature: Plotting of metrics

SkafteNicki:
## What does this PR do?

Fixes https://github.com/Lightning-AI/metrics/issues/652
Fixes https://github.com/Lightning-AI/metrics/issues/1265

This PR adds basic plotting functionality to torchmetrics. Idea is that every modular implementation should have an `plot` method that can create a basic plot of the metric. All `plot` methods should have the signature:

```python
def compute(self) -> TypeT:
    ...

def plot(self, val: Optional[TypeT, Sequence[TypeT]])
    val = val | self.compute()
    ...
```
meaning that if no value is provided, `self.compute` is called. Additionally, either a single value can be provided or a list of values to plot a metric as it has progressed. Please try executing the `examples/plotting` for examples.

This PR is meant as an proof-on-concept and when it (hopefully) gets merge we can follow up with adding the functionality to other metrics.

## Before submitting

## PR review

## Did you have fun?


SkafteNicki:
Executing `examples/plotting.py` should give something like this:

Single metric value
Figure_a

Single metric value, one per class
Figure_b

Multiple metric values
Figure_c

Multiple metric values, one per class
Figure_d

src/torchmetrics/classification/accuracy.py
```diff
@@ -212,11 +213,21 @@ class MulticlassAccuracy(MulticlassStatScores):
     is_differentiable = False
     higher_is_better = True
     full_state_update: bool = False
+    plot_options = {"lower_bound": 0.0, "upper_bound": 1.0, "legend_name": "Class"}
 
     def compute(self) -> Tensor:
         tp, fp, tn, fn = self._final_state()
         return _accuracy_reduce(tp, fp, tn, fn, average=self.average, multidim_average=self.multidim_average)
 
+    def plot(self, val: Optional[Union[Tensor, Sequence[Tensor]]] = None, show: bool = False) -> _PLOT_OUT_TYPE:
```
justusschock, line 222: I think this should be a static method as all the things accessed here should actually be on a per-class and not per-instance level, no? Then you could also use plotting without instantiating. This would go for all metrics accordingly.

src/torchmetrics/utilities/plot.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def plot_single_or_multi_val(
+    val: Optional[Union[Tensor, Sequence[Tensor]]],
+    higher_is_better: Optional[bool] = None,
+    lower_bound: Optional[float] = None,
+    upper_bound: Optional[float] = None,
+    legend_name: Optional[str] = None,
+    name: Optional[str] = None,
+    fig_ax: Optional[_PLOT_OUT_TYPE] = None,
```
justusschock, line 50: we only need the axes here and can retrieve the figure with `ax.get_figure()` (if we need it at all). This makes it easier to pass things over as plotting should ideally only be done on the axes

src/torchmetrics/utilities/plot.py
```diff
@@ -0,0 +1,264 @@
...
+        for ax in axs[N:]:
+            ax.remove()
+        return axs[:N]
+
+
+def plot_confusion_matrix(
+    confmat: Tensor,
+    add_text: bool = True,
+    labels: Optional[List[str]] = None,
+    cmap: Optional[matplotlib.colors.Colormap] = None,
```
justusschock, line 155: also add option for axes? Should be consistent :)

src/torchmetrics/utilities/plot.py
```diff
@@ -0,0 +1,264 @@
...
+                ax.text(jj, ii, str(val.item()), ha="center", va="center", fontsize=15)
+
+    return fig, axs
+
+
+def _plot_curve(
+    input: Union[Tuple[Tensor, Tensor, Tensor], Tuple[List[Tensor], List[Tensor], List[Tensor]]],
+    auc: bool = False,
+    single_plot: bool = False,
+    xy_labels: Tuple[str, str] = ("X axis", "Y axis"),
```
justusschock, line 203: also support to pass axes? What if I by choice want multiple plots on the same axes?

justusschock:
Something more general: Do we want to support smoothing? Especially for multiple timepoints this is often done to get a more general idea of the progress instead of spiky individual values (See tensorboard for example).

src/torchmetrics/utilities/plot.py
SkafteNicki: will update

src/torchmetrics/utilities/plot.py
SkafteNicki: will update

SkafteNicki:
> Something more general: Do we want to support smoothing? Especially for multiple timepoints this is often done to get a more general idea of the progress instead of spiky individual values (See tensorboard for example).

I would be fine with supporting this :)
We would need to add `scipy` as an dependency, but thats the only "problem".

src/torchmetrics/classification/accuracy.py
justusschock: That's true. Ideally, we want both options...

So I should be able to do `my_acc.plot()` as well as `MulticlassAccuracy.plot([val1, val2])`.

We could do instance-checks on `val` and allow it to also be a Metric, but then we would have to go with `my_acc.plot(my_acc)` which is also a bit weird.

Or we just keep it as is, then people would have to instantiate the metric, which  would also be fine the more I think about it :)

docs/source/conf.py
```diff
@@ -101,8 +101,12 @@ def _transform_changelog(path_in: str, path_out: str) -> None:
     "sphinx_paramlinks",
     "sphinx.ext.githubpages",
     "pt_lightning_sphinx_theme.extensions.lightning",
+    "matplotlib.sphinxext.plot_directive",
 ]
 
+# something
```
stancld, line 107: Guess here should be something different :D

requirements/integrate.txt
```diff
@@ -1 +1,2 @@
 pytorch-lightning>=1.5
+matplotlib
```
stancld, line 2: Any minimal version required here?

src/torchmetrics/classification/accuracy.py
```diff
@@ -107,11 +112,45 @@ class BinaryAccuracy(BinaryStatScores):
...
 
     def compute(self) -> Tensor:
         tp, fp, tn, fn = self._final_state()
         return _accuracy_reduce(tp, fp, tn, fn, average="binary", multidim_average=self.multidim_average)
 
+    def plot(self, val: Optional[Union[Tensor, Sequence[Tensor]]] = None) -> _PLOT_OUT_TYPE:
+        """Plot a single or multiple values from the metric.
+
+        Args:
+            val: Either single result from calling `metric.forward` or `metric.compute` or an list of these results.
```
stancld, line 125:
```suggestion
            val: Either a single result from calling `metric.forward` or `metric.compute` or a list of these results.
```

src/torchmetrics/classification/accuracy.py
```diff
@@ -212,11 +251,45 @@ class MulticlassAccuracy(MulticlassStatScores):
...
 
     def compute(self) -> Tensor:
         tp, fp, tn, fn = self._final_state()
         return _accuracy_reduce(tp, fp, tn, fn, average=self.average, multidim_average=self.multidim_average)
 
+    def plot(self, val: Optional[Union[Tensor, Sequence[Tensor]]] = None) -> _PLOT_OUT_TYPE:
+        """Plot a single or multiple values from the metric.
+
+        Args:
+            val: Either single result from calling `metric.forward` or `metric.compute` or an list of these results.
```
stancld, line 264:
```suggestion
            val: Either a single result from calling `metric.forward` or `metric.compute` or a list of these results.
```

src/torchmetrics/classification/confusion_matrix.py
```diff
@@ -220,6 +225,39 @@ def compute(self) -> Tensor:
         """
         return _multiclass_confusion_matrix_compute(self.confmat, self.normalize)
 
+    def plot(self, val: Optional[Tensor] = None) -> _PLOT_OUT_TYPE:
+        """Plot a single or multiple values from the metric.
+
+        Args:
+            val: Either single result from calling `metric.forward` or `metric.compute` or an list of these results.
```
stancld, line 232:
```suggestion
            val: Either a single result from calling `metric.forward` or `metric.compute` or a list of these results.
```

src/torchmetrics/regression/mse.py
```diff
@@ -71,3 +73,47 @@ def update(self, preds: Tensor, target: Tensor) -> None:  # type: ignore
     def compute(self) -> Tensor:
         """Computes mean squared error over state."""
         return _mean_squared_error_compute(self.sum_squared_error, self.total, squared=self.squared)
+
+    def plot(self, val: Optional[Union[Tensor, Sequence[Tensor]]] = None) -> _PLOT_OUT_TYPE:
+        """Plot a single or multiple values from the metric.
+
+        Args:
+            val: Either single result from calling `metric.forward` or `metric.compute` or an list of these results.
```
stancld, line 81:
```suggestion
            val: Either a single result from calling `metric.forward` or `metric.compute` or a list of these results.
```

stancld (commented):
That's great, looking forward to seeing this merged! :]

Maybe can we also update the `README.md` with info that we offer some plots out-of-the-box?

examples/plotting.py
```diff
@@ -0,0 +1,84 @@
...
+import torch
+
+if __name__ == "__main__":
+    list_of_choices = ["accuracy", "mean_squared_error", "confusion_matrix"]
+    parser = argparse.ArgumentParser(description="Example script for plotting metrics.")
+    parser.add_argument("metric", choices=list_of_choices)
+    args = parser.parse_args()
+
+    if args.metric == "accuracy":
+        from torchmetrics.classification import MulticlassAccuracy
```
Borda, line 26: let's move these blocks to functions so it is easier to read... :)

requirements/integrate.txt
Borda:
```suggestion
matplotlib>=3.0.0
```

src/torchmetrics/classification/accuracy.py
```diff
@@ -107,11 +112,45 @@ class BinaryAccuracy(BinaryStatScores):
...
+
+        Returns:
+            fig: Figure object
+            ax: Axes object
+
+        Raises:
+            ModuleNotFoundError:
+                If `matplotlib` is not installed
+
+        .. plot::
```
Borda, line 136: Does that do what I assume? that would be cool!

Borda (commented):
Shall we include `plot` to the base class and set it as an unimplemented exception?

src/torchmetrics/classification/accuracy.py
stancld: image
Yep :roc

requirements/devel.txt
```diff
@@ -4,6 +4,9 @@
 # add the testing dependencies
 -r test.txt
 
+# add integration dependencies
+-r integrate.txt
+
```
Borda, line 9:
```suggestion
```
lets d not do this as PL ould reinstall also TM and not sure what the order is...

src/torchmetrics/classification/accuracy.py
```diff
@@ -107,12 +112,58 @@ class BinaryAccuracy(BinaryStatScores):
     is_differentiable = False
     higher_is_better = True
     full_state_update: bool = False
+    plot_options = {"lower_bound": 0.0, "upper_bound": 1.0}
```
Borda, line 115:
```suggestion
    plot_options: dict = {"lower_bound": 0.0, "upper_bound": 1.0}
```

src/torchmetrics/classification/accuracy.py
```diff
@@ -317,6 +414,7 @@ class MultilabelAccuracy(MultilabelStatScores):
     is_differentiable = False
     higher_is_better = True
     full_state_update: bool = False
+    plot_options = {"lower_bound": 0.0, "upper_bound": 1.0, "legend_name": "Label"}
```
Borda, line 417:
```suggestion
    plot_options: dict = {"lower_bound": 0.0, "upper_bound": 1.0, "legend_name": "Label"}
```

requirements/integrate.txt
```diff
@@ -1 +1,2 @@
 pytorch-lightning>=1.5
+matplotlib>=3.0.0
```
Borda, line 2: lets make it as `extras.txt`

src/torchmetrics/regression/mse.py
```diff
@@ -42,6 +47,7 @@ class MeanSquaredError(Metric):
     is_differentiable = True
     higher_is_better = False
     full_state_update = False
+    plot_options = {"lower_bound": 0.0}
```
Borda, line 50:
```suggestion
    plot_options: dict = {"lower_bound": 0.0}
```

Borda (approved):

src/torchmetrics/classification/accuracy.py
```diff
@@ -107,12 +112,58 @@ class BinaryAccuracy(BinaryStatScores):
...
+    plot_options: dict = {"lower_bound": 0.0, "upper_bound": 1.0}
 
     def compute(self) -> Tensor:
         """Computes accuracy based on inputs passed in to ``update`` previously."""
         tp, fp, tn, fn = self._final_state()
         return _accuracy_reduce(tp, fp, tn, fn, average="binary", multidim_average=self.multidim_average)
 
+    def plot(self, val: Optional[Union[Tensor, Sequence[Tensor]]] = None) -> _PLOT_OUT_TYPE:
+        """Plot a single or multiple values from the metric.
+
```
awaelchli, line 124: Would it make sense to take the axis as input here, so that one can plot to an existing axes/figure?

awaelchli (approved):

justusschock (approved):

Borda:
@SkafteNicki could we check the failing tests? :)

PR #1418, opened 2022-12-29
cleaning ref. metric

Borda:
## What does this PR do?

Cleaning the name of reference metric in args and setting pool as nb of CPU

## Before submitting

## PR review

## Did you have fun?


tests/unittests/helpers/testers.py
```diff
@@ -33,8 +33,13 @@
 except RuntimeError:
     pass
 
-NUM_PROCESSES = 2
-NUM_BATCHES = 4  # Need to be divisible with the number of processes
+if torch.cuda.is_available():
```
justusschock, line 36: please let's keep it with 2 all the time, everything else will slow things down and to test ddp we only need two processes

justusschock (changes requested):

tests/unittests/helpers/testers.py
Borda: I have not seen any slowdown... :confused:

SkafteNicki (approved):
Fine by me, but agree with @justusschock that for simplicity that we should keep the `n_process` and `num_batches` size fixed.

Borda:
> Fine by me, but agree with @justusschock that for simplicity that we should keep the `n_process` and `num_batches` size fixed.

yes, already reverted

justusschock (approved):

PR #1419, opened 2022-12-30
Return also classes for MAP metric

SkafteNicki:
## What does this PR do?

Fixes https://github.com/Lightning-AI/metrics/issues/1417
Returns also `classes` for MAP metric.

## Before submitting

## PR review

## Did you have fun?


Borda (approved):

justusschock (approved):

adamjstewart:
Unfortunately this makes it much more difficult to use this metric in Lightning.

The following code:
```python
class Foo(LightningModule):
    def __init__(self):
        self.metric = MeanAveragePrecision()

    def on_validation_epoch_end(self):
        self.log_dict(self.metric.compute())
```
used to work in 0.11.4. However, this PR adds a new `classes` key-pair to the metrics dict containing a list of all observed classes. `log_dict` now complains about this entry:
```
ValueError: `self.log(classes, tensor([0, 1], dtype=torch.int32))` was called, but the tensor must have a single element. You can try doing `self.log(classes, tensor([0, 1], dtype=torch.int32).mean())`
```
Here, `tensor([0, 1])` is the `classes` value. Current workaround is to delete this key from the dict before logging, but it would be good to do this automatically.

PR #1493, opened 2023-02-09
Improve speed and memory consumption of binned `PrecisionRecallCurve`

### LINKED ISSUE #1492: Binned implementation of PrecisionRecallCurve is extremely slow for large batch sizes
## 🐛 Bug

Since the classification metrics refactor in torchmetrics 0.10, the update step of `PrecisionRecallCurve` and metrics based on it (such as `AveragePrecision`) is extremely slow and memory-consuming when using the binned implementation with many samples. This is, for example, the case in semantic segmentation scenarios, where every pixel is a sample.

The binned implementation is intended to be both faster and memory efficient than computing the exact precision-recall curve. This is currently not the case.

### To Reproduce

```python
import torch
from torchmetrics.classification import BinaryAveragePrecision

targets = torch.randint(0, 2, (16, 512, 512), device="cuda")
preds = torch.rand((16, 512, 512), device="cuda")

metric_exact = BinaryAveragePrecision().cuda()
metric_binned = BinaryAveragePrecision(thresholds=200).cuda()

with torch.inference_mode():
    start_exact_evt = torch.cuda.Event(enable_timing=True)
    update_exact_evt = torch.cuda.Event(enable_timing=True)
    compute_exact_evt = torch.cuda.Event(enable_timing=True)
    start_binned_evt = torch.cuda.Event(enable_timing=True)
    update_binned_evt = torch.cuda.Event(enable_timing=True)
    compute_binned_evt = torch.cuda.Event(enable_timing=True)
    
    start_exact_evt.record()
    metric_exact.update(preds, targets)
    update_exact_evt.record()
    ap_exact = metric_exact.compute()
    compute_exact_evt.record()
    
    start_binned_evt.record()
    metric_binned.update(preds, targets)
    update_binned_evt.record()
    ap_binned = metric_exact.compute()
    compute_binned_evt.record()

torch.cuda.synchronize()

print("AP (exact):", ap_exact)
print("AP (binned):", ap_binned)
print("Time (exact): {} ms + {} ms".format(
    start_exact_evt.elapsed_time(update_exact_evt),
    update_exact_evt.elapsed_time(compute_exact_evt)
))
print("Time (binned): {} ms + {} ms".format(
    start_binned_evt.elapsed_time(update_binned_evt),
    update_binned_evt.elapsed_time(compute_binned_evt)
))
```

Output (using an RTX 3090):
```
AP (exact): tensor(0.5008, device='cuda:0')
AP (binned): tensor(0.5008, device='cuda:0')
Time (exact): 1.8867520093917847 ms + 4.845024108886719 ms
Time (binned): 10816.78515625 ms + 0.002047999994829297 ms
```

Binned average precision was 1600x slower than exact average precision.
The memory consumption of exact AP on the GPU was 260 MB, while binned AP consumed 20,000 MB.

### Expected behavior

The binned implementation of `PrecisionRecallCurve`, i.e., with `thresholds != None`, should be faster and more memory-efficient.

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): 0.11.1 (pip)
- Python & PyTorch Version (e.g., 1.0): Python 3.7.9, PyTorch 1.12.1

### Additional context

The [current implementation of `_binary_precision_recall_curve_update`](https://github.com/Lightning-AI/metrics/blob/b95d482b81a78086764ee89cde0c211eb67dec4c/src/torchmetrics/functional/classification/precision_recall_curve.py#L185) compares all predictions with all thresholds at once, which is very memory consuming:

```python
len_t = len(thresholds)
preds_t = (preds.unsqueeze(-1) >= thresholds.unsqueeze(0)).long()  # num_samples x num_thresholds
unique_mapping = preds_t + 2 * target.unsqueeze(-1) + 4 * torch.arange(len_t, device=target.device)
bins = _bincount(unique_mapping.flatten(), minlength=4 * len_t)
return bins.reshape(len_t, 2, 2)
```

Computing `preds_t` and `unique_mapping` is quite fast, but both consume large amounts of memory (number of samples * number of thresholds * 8 byte). `_bincount`, on the other hand, does not consume much memory, but a lot of time.

Iterating over the thresholds one by one is significantly faster (0.6 sec instead of 10 sec) and only consumes 150 MB memory:

```python
confmat = thresholds.new_empty((len(thresholds), 2, 2), dtype=torch.int64)
for i in range(len(thresholds)):
    preds_t = (preds >= thresholds[i]).long()
    unique_mapping = preds_t + 2 * targets
    bins = torch.bincount(unique_mapping, minlength=4)
    confmat[i] = bins.reshape(2, 2)
```

Using the [implementation of `BinnedPrecisionRecallCurve` from torchmetrics 0.9](https://github.com/Lightning-AI/metrics/blob/ff61c482e5157b43e647565fa0020a4ead6e9d61/torchmetrics/classification/binned_precision_recall.py#L145) is even faster (0.1 sec) and uses only 60 MB:

```python
confmat = thresholds.new_empty((len(thresholds), 2, 2), dtype=torch.int64)
targets_t = targets == 1
for i in range(len(thresholds)):
    preds_t = preds >= thresholds[i]
    confmat[i, 1, 1] = (targets_t & preds_t).sum()
    confmat[i, 0, 1] = ((~targets_t) & preds_t).sum()
    confmat[i, 1, 0] = (targets_t & (~preds_t)).sum()
    confmat[i, 0, 0] = len(preds_t) - confmat[i, 0, 1] - confmat[i, 1, 0] - confmat[i, 1, 1]
```

Callidior:
## What does this PR do?

Fixes #1492 by reverting the implementation of the update step of binary and multiclass PrecisionRecallCurve with a fixed number of thresholds to the implementation from torchmetrics 0.9.

I did not touch the multilabel case, because it is difficult to adapt the `ignore_idx` filtering logic, a first attempt of implementing it on a per-threshold basis was much slower, and multilabel classification is not common in semantic segmentation scenarios, which is where we typically have to deal with huge numbers of samples.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Performance Comparison

Hardware: NVIDIA RTX 3090

### `BinaryPrecisionRecallCurve.update()`

`thresholds=200`

|  #Samples  | Current Speed | New Speed | Current Memory | New Memory |
|-----------:|--------------:|----------:|---------------:|-----------:|
|        100 |        0.2 ms |   23.8 ms |         < 1 MB |     < 1 MB |
|      1,000 |        1.1 ms |   24.4 ms |          20 MB |     < 1 MB |
|     25,000 |       24.0 ms |   25.8 ms |         200 MB |     < 1 MB |
|    100,000 |      115.3 ms |   25.5 ms |         722 MB |       2 MB |
|  1,000,000 |    1,012.1 ms |   26.7 ms |       7,156 MB |       2 MB |
|  3,000,000 |    7,948.3 ms |   56.5 ms |      18,886 MB |      44 MB |

### `MulticlassPrecisionRecallCurve.update()`

`num_classes=100, thresholds=200`

|  #Samples  | Current Speed | New Speed | Current Memory | New Memory |
|-----------:|--------------:|----------:|---------------:|-----------:|
|         10 |        0.2 ms |   15.5 ms |          20 MB |     < 1 MB |
|        100 |        0.5 ms |   15.2 ms |          52 MB |     < 1 MB |
|      1,000 |        3.5 ms |   15.2 ms |         484 MB |       4 MB |
|     10,000 |       30.6 ms |   30.8 ms |       4,770 MB |      22 MB |
|     25,000 |       76.6 ms |   70.6 ms |      12,404 MB |      60 MB |

The implementation proposed here is not faster in all cases but always more memory efficient. In particular, it is up to 100x slower for few numbers of samples such as in a typical image-level classification scenario. However, it is up to 100x faster for scenarios with many samples per batch such as in semantic segmentation use-cases. In these scenarios, the extreme memory consumption of the previous implementation is the largest deal breaker.

## Did you have fun?


Callidior:
Hi @justusschock! Thanks for your feedback. I completely agree that there are also common use cases (such as image-level classification with < 100 samples per batch), for which the current implementation would be the better choice. An automatic switch that selects the more suitable algorithm would be my preferred solution as well.

For the setup I benchmarked above for binary classification, one would probably switch to the old implementation (this PR) from 10k samples and above. I could run some more tests with different numbers of thresholds, but the result might even by hardware-dependent.

What is more predictable, though, is the memory consumption. I would definitely not spend much more than 100 MB for computing a metric if I could do the same with 1-2 MB. We could estimate the memory needed by the current implementation based on `number of samples * number of thresholds (* number of classes) * constant` and switch implementations based on that.

What do you think?

Callidior:
A more detailed speed benchmark on the binary case (time in ms):

image

- The loop implementation scales linearly with the number of thresholds but is almost insensitive to the number of samples.
- The vectorized implementation scales linearly with the number of samples.
- Keep in mind that memory consumption also scales almost linearly with number of samples for the vectorized implementation. The case in the bottom right corner takes 24 GB of VRAM, while the loop implementation has a constant memory consumption of 2 MB.

The documentation of `PrecisionRecallCurve` currently states;

> The implementation both supports calculating the metric in a non-binned but accurate version and a binned version that is less accurate but more memory efficient. Setting the thresholds argument to None will activate the non-binned version that uses memory of size O(n_samples) whereas setting the thresholds argument to either an integer, list or a 1d tensor will use a binned version that uses memory of size O(n_thresholds) (constant memory).

That is true for the metric's state but not for intermediate computations during the update step. The documentation does not mention speed but puts a promise on memory efficiency in the foreground, which is currently not kept.

With the current implementation, I don't see any benefit of using the binned version of the metric. If you have few samples, the accurate version (`thresholds=None`) is still very tractable and more accurate than the binned one. If you have many samples, both the binned and the accurate version of unusable due to high memory requirements and computation time.

@justusschock: Given that users can already switch between the accurate and the binned version, I don't think an explicit second switch for two different implementations of the binned version makes sense.

SkafteNicki:
@Callidior thank you for investigating this issue.

> With the current implementation, I don't see any benefit of using the binned version of the metric. If you have few samples, the accurate version (thresholds=None) is still very tractable and more accurate than the binned one. If you have many samples, both the binned and the accurate version of unusable due to high memory requirements and computation time.

The thing your analysis is missing is that you are assuming that all samples are evaluated at the same time, however this is not how it is done in practise in any deep learning pipeline where samples arrive in mini-batches. Thus, there is a significant difference in memory between doing
```python
preds, target = torch.rand(1e6), torch.randint(2, (1e6,))
output = metric(preds, target)
```
and
```python
preds, target = torch.rand(1e6), torch.randint(2, (1e6,))
for i in range(1000):
    metric.update(preds[1000*i:1000*(i+1)], target[1000*i:1000*(i+1)])
output = metric.compute()
```
this is the reason for the binned version existing, because the non-binned version would have to save all samples in memory in both cases whereas the binned version only needs to save reduced state (the `[n_threshold, n_class, n_class]` tensor) which is significant less memory.

SkafteNicki:
I would agree with @justusschock that the correct approach is implement some kind of heuristic that changes between `bincount` and a `for-loop` depending on how many samples and threshold the user are using. To settle on this heuristic, would it be possible for you @Callidior do share the benchmark script you have developed. In particular I am also interested in what the difference between doing this on CPU is (it seems you have only tried out GPU).

Callidior:
@SkafteNicki Fair point. I updated the PR to implement the suggested heuristic switch between the two implementations.

Further benchmark results for the binary and multi-class case on both CPU and GPU can be found below, as well as the benchmark script. The picture on CPU doesn't look too differently from the situation on the GPU.

Based on those benchmarks, I'd propose the following heuristic:
- For binary classification, use the loop implementation if `num_samples > 50,000`.
- For multiclass classification, use the loop implementation if `num_samples * num_classes > 1,000,000`.

# Speed Benchmark

GPU: NVIDIA GeForce RTX 3090
CPU: AMD Ryzen 9 5900X (12-Core)

## Binary

### GPU

pr_benchmark_binary_gpu

### CPU

pr_benchmark_binary_cpu

## Multi-Class

### GPU

#### `num_classes==10`

pr_benchmark_10cls_gpu

#### `num_classes==100`

pr_benchmark_100cls_gpu

#### `num_classes==1000`

pr_benchmark_1000cls_gpu

### CPU

#### `num_classes==10`

pr_benchmark_10cls_cpu

#### `num_classes==100`

pr_benchmark_100cls_cpu

#### `num_classes==1000`

My 96 GB of main memory are not sufficient to run the vectorized implementation.

## Benchmark Script

<details>
<summary>Click to show code</summary>

```python
import argparse
import numpy as np
import torch
import matplotlib.colors
import matplotlib.pyplot as plt
from time import perf_counter_ns
from functools import partial
from tqdm import tqdm


def binary_pr_curve_update_vectorized(preds, targets, thresholds):
    preds_t = (preds.unsqueeze(-1) >= thresholds.unsqueeze(0)).long()
    unique_mapping = preds_t + 2 * targets.unsqueeze(-1) + 4 * torch.arange(len(thresholds), device=targets.device)
    bins = torch.bincount(unique_mapping.flatten(), minlength=4 * len(thresholds))
    confmat = bins.reshape(len(thresholds), 2, 2)
    return confmat

def binary_pr_curve_update_loop(preds, targets, thresholds):
    confmat = thresholds.new_empty((len(thresholds), 2, 2), dtype=torch.int64)
    targets_t = targets == 1
    for i in range(len(thresholds)):
        preds_t = preds >= thresholds[i]
        confmat[i, 1, 1] = (targets_t & preds_t).sum()
        confmat[i, 0, 1] = ((~targets_t) & preds_t).sum()
        confmat[i, 1, 0] = (targets_t & (~preds_t)).sum()
    confmat[:, 0, 0] = len(preds_t) - confmat[:, 0, 1] - confmat[:, 1, 0] - confmat[:, 1, 1]
    return confmat


def multiclass_pr_curve_update_vectorized(preds, targets, thresholds, num_classes):
    len_t = len(thresholds)
    preds_t = (preds.unsqueeze(-1) >= thresholds.unsqueeze(0).unsqueeze(0)).long()
    target_t = torch.nn.functional.one_hot(targets, num_classes=num_classes)
    unique_mapping = preds_t + 2 * target_t.unsqueeze(-1)
    unique_mapping += 4 * torch.arange(num_classes, device=preds.device).unsqueeze(0).unsqueeze(-1)
    unique_mapping += 4 * num_classes * torch.arange(len_t, device=preds.device)
    bins = torch.bincount(unique_mapping.flatten(), minlength=4 * num_classes * len_t)
    return bins.reshape(len_t, num_classes, 2, 2)


def multiclass_pr_curve_update_loop(preds, targets, thresholds, num_classes):
    len_t = len(thresholds)
    target_t = torch.nn.functional.one_hot(targets, num_classes=num_classes)
    confmat = thresholds.new_empty((len_t, num_classes, 2, 2), dtype=torch.int64)
    for i in range(len_t):
        preds_t = preds >= thresholds[i]
        confmat[i, :, 1, 1] = (target_t & preds_t).sum(dim=0)
        confmat[i, :, 0, 1] = ((~target_t) & preds_t).sum(dim=0)
        confmat[i, :, 1, 0] = (target_t & (~preds_t)).sum(dim=0)
    confmat[:, :, 0, 0] = len(preds_t) - confmat[:, :, 0, 1] - confmat[:, :, 1, 0] - confmat[:, :, 1, 1]
    return confmat


def benchmark_gpu(fn1, fn2, num_samples_list, num_thresholds_list, num_classes=1, num_loops=100):
    times1 = np.empty((len(num_samples_list), len(num_thresholds_list)))
    times2 = np.empty((len(num_samples_list), len(num_thresholds_list)))

    evt1 = torch.cuda.Event(enable_timing=True)
    evt2 = torch.cuda.Event(enable_timing=True)
    evt3 = torch.cuda.Event(enable_timing=True)

    for i, num_samples in enumerate(tqdm(num_samples_list)):
        for j, num_thresholds in enumerate(tqdm(num_thresholds_list, leave=False)):
            targets = torch.randint(0, max(2, num_classes), (num_samples,), device="cuda")
            preds = torch.rand((num_samples, num_classes), device="cuda").squeeze()
            thresholds = torch.linspace(0, 1, num_thresholds, device="cuda")

            evt1.record()
            try:
                for _ in range(num_loops):
                    confmat1 = fn1(preds, targets, thresholds)
            except RuntimeError:
                confmat1 = None
            evt2.record()
            try:
                for _ in range(num_loops):
                    confmat2 = fn2(preds, targets, thresholds)
            except RuntimeError:
                confmat2 = None
            evt3.record()

            torch.cuda.synchronize()
            assert confmat1 is None or confmat2 is None or torch.allclose(confmat1, confmat2)

            times1[i,j] = evt1.elapsed_time(evt2) / num_loops if confmat1 is not None else np.nan
            times2[i,j] = evt2.elapsed_time(evt3) / num_loops if confmat2 is not None else np.nan

    return times1, times2


def benchmark_cpu(fn1, fn2, num_samples_list, num_thresholds_list, num_classes=1, num_loops=100):
    times1 = np.empty((len(num_samples_list), len(num_thresholds_list)))
    times2 = np.empty((len(num_samples_list), len(num_thresholds_list)))

    for i, num_samples in enumerate(tqdm(num_samples_list)):
        for j, num_thresholds in enumerate(tqdm(num_thresholds_list, leave=False)):
            targets = torch.randint(0, max(2, num_classes), (num_samples,), device="cpu")
            preds = torch.rand((num_samples, num_classes), device="cpu").squeeze()
            thresholds = torch.linspace(0, 1, num_thresholds, device="cpu")

            evt1 = perf_counter_ns()
            for _ in range(num_loops):
                confmat1 = fn1(preds, targets, thresholds)
            evt2 = perf_counter_ns()
            for _ in range(num_loops):
                confmat2 = fn2(preds, targets, thresholds)
            evt3 = perf_counter_ns()

            assert torch.allclose(confmat1, confmat2)

            times1[i,j] = (evt2 - evt1) / (num_loops * 1e6)
            times2[i,j] = (evt3 - evt2) / (num_loops * 1e6)

    return times1, times2


def plot_timings(*times, num_samples_list, num_thresholds_list, titles=None):
    fig, axes = plt.subplots(1, len(times), figsize=(6*len(times), 8), constrained_layout=True, sharey=True)

    vmin = min([np.nanmin(t) for t in times])
    vmax = max([np.nanmax(t) for t in times])
    scaler = matplotlib.colors.LogNorm(vmin=vmin, vmax=vmax)
    for ax, t in zip(axes, times):
        ax.imshow(t.T, norm=scaler)

    if titles is not None:
        for ax, title in zip(axes, titles):
            ax.set_title(title)

    for i in range(len(num_samples_list)):
        for j in range(len(num_thresholds_list)):
            for ax, t in zip(axes, times):
                if not np.isnan(t[i,j]):
                    ax.annotate(
                        f"{t[i,j]:.1f}",
                        (i, j),
                        ha="center",
                        va="center",
                        color="white" if scaler(t[i,j]) < 0.8 else "black"
                    )

    axes[0].set_ylabel("Number of Thresholds")
    for ax in axes.ravel():
        ax.set_xlabel("Number of Samples")
        ax.set_xticks(np.arange(len(num_samples_list)), [f"{x:,d}" for x in num_samples_list])
        ax.set_yticks(np.arange(len(num_thresholds_list)), [f"{x:,d}" for x in num_thresholds_list])

    return fig


if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--num-classes", type=int, default=1, help="Number of classes. Set to 1 for binary classification.")
    parser.add_argument("--num-loops", type=int, default=100, help="Number of repetitions per experiment.")
    parser.add_argument("--cpu", action="store_true", default=False, help="Run on CPU.")
    args = parser.parse_args()
    
    num_thresholds_list = [10, 50, 100, 200, 500, 1000]
    if args.num_classes == 1:
        fn1 = binary_pr_curve_update_vectorized
        fn2 = binary_pr_curve_update_loop
        num_samples_list = [10, 100, 1000, 10000, 100000, 1000000]
    else:
        fn1 = partial(multiclass_pr_curve_update_vectorized, num_classes=args.num_classes)
        fn2 = partial(multiclass_pr_curve_update_loop, num_classes=args.num_classes)
        num_samples_list = [10, 100, 1000, 10000, 25000]

    benchmark_fn = benchmark_cpu if args.cpu else benchmark_gpu
    
    times = benchmark_fn(
        fn1,
        fn2,
        num_samples_list,
        num_thresholds_list,
        num_classes=args.num_classes,
        num_loops=args.num_loops,
    )
    plot_timings(
        *times,
        num_samples_list=num_samples_list,
        num_thresholds_list=num_thresholds_list,
        titles=["Vectorized", "Loop"],
    )
    plt.show()
```

</details>

justusschock:
@Callidior this heuristic is fine with me. Thanks for all the detailed benchmarks!

src/torchmetrics/functional/classification/precision_recall_curve.py
```diff
@@ -194,13 +194,54 @@ def _binary_precision_recall_curve_update(
     """
     if thresholds is None:
         return preds, target
+    if preds.numel() <= 50000:
+        update_fn = _binary_precision_recall_curve_update_vectorized
+    else:
+        update_fn = _binary_precision_recall_curve_update_loop
```
Borda, line 200:
```suggestion
    update_fn = _binary_precision_recall_curve_update_vectorized if preds.numel() <= 50_000 else _binary_precision_recall_curve_update_loop
```

src/torchmetrics/functional/classification/precision_recall_curve.py
```diff
@@ -409,8 +450,25 @@ def _multiclass_precision_recall_curve_update(
     """
     if thresholds is None:
         return preds, target
+    if preds.numel() * num_classes <= 1000000:
+        update_fn = _multiclass_precision_recall_curve_update_vectorized
+    else:
+        update_fn = _multiclass_precision_recall_curve_update_loop
```
Borda, line 456:
```suggestion
    size_thr = preds.numel() * num_classes <= 1_000_000
    update_fn = _multiclass_precision_recall_curve_update_vectorized if size_thr else _multiclass_precision_recall_curve_update_loop
```

Borda (approved):

SkafteNicki (approved):

Issue #1591, opened 2023-03-05
Add Modified Panoptic Quality metric


marcocaccin:
## 🚀 Feature

Add Modified Panoptic Quality metric introduced in [Seamless Scene Segmentation](https://arxiv.org/abs/1905.01220), Section 4 "Revisiting Panoptic Segmentation".

### Motivation

The Modified Panoptic Quality is a subtle modification of the original, and it comes in handy to have it available when dealing with Panoptic Segmentation problems: it's not easy to say whether the original or the modified one is a better choice in any given situation, so one would ideally track both. 
To prove the point, some public datasets such as [nuScenes](https://www.nuscenes.org/panoptic/?externalData=all&mapData=all&modalities=Any) use both variants for evaluation on the leaderboard.

### Pitch

I'm happy to continue the work on PQ to extend its functionality 😉 

### Additional context

Question to the maintainers: would you rather see this implemented in its own class and functional versions, or as a switch (e.g., `implementation: Literal['original', 'modified'] = 'original'`) of the already existing metrics?
The code is almost identical besides a couple of extra `if` statements in the `update()` step, and so are the signatures, so I would lean on the latter option to avoid additional copy-paste.

SkafteNicki:
Hi @marcocaccin, sound like a great enhancement.
After looking through the paper you linked, it looks to me that the easier solution would be to just have this as an argument `implementation: Literal['original', 'modified'] = 'original'` in the current PQ implementation as the authors does not really consider this a new metric but only an adjusted version of the original.

stancld:
@SkafteNicki Just my two cents. On the other hand, we have `BLEUScore` and `SacreBLEUScore`, which basically differ in tokenization. But both options seem to be acceptable.

SkafteNicki:
@stancld true. To me, it actually does not matter what option we go with, like you said but options are acceptable :]

marcocaccin:
Thanks for the feedback! After thinking about it, maybe the least confusing solution is to have a common private module where all the functionality is implemented (`torchmetrics/functional/detection/_panoptic_quality_common.py`) and separate super-shallow modules containing the individual functional and class implementations.
The only reason I have for that is the docstring: I find that linking a single paper and writing a single docstring test is the least confusing.
Feel free to assign the issue to me btw 🙂

marcocaccin:
[...]

PR #1596, opened 2023-03-06
Refactor panoptic quality

marcocaccin:
## What does this PR do?

Preparation work for  #1591: breaks apart the big messy update() function, and moves all the helper functions to a new `src/torchmetrics/functional/detection/_panoptic_quality_common.py` file.

To make the review a lot simpler, the PR contains 2 commits: I suggest to look at the first commit 7cbf54b6743aa251a3e69ae801788dc958aebe60 which contains all the new code and it can be `diff`'d against `master`, whereas the 2nd commit d873aabc3140f390b4eaef2fe4180c7b1734c8e0 is doing nothing else than moving code around.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?


marcocaccin:
Not sure how to interpret the codecov failure, the PR doesn't add any "new" code. Is it a CI glitch? Should I do something about it?

src/torchmetrics/detection/panoptic_quality.py
```diff
@@ -11,13 +11,12 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import warnings
 from typing import Any, Collection, Optional, Sequence, Union
 
 import torch
 from torch import Tensor
 
-from torchmetrics.functional.detection.panoptic_quality import (
+from torchmetrics.functional.detection._panoptic_quality_common import (
```
Borda, line 19: What is the motivation for moving it to a separate file when common still means shared only for this one metric? it generates a much larger diff so it is not so easy to track real changes... :chipmunk:

Borda:
> I suggest to look at the first commit [7cbf54b](https://github.com/Lightning-AI/metrics/commit/7cbf54b6743aa251a3e69ae801788dc958aebe60) which contains all the new code and it can be `diff`'d against `master`

let's just stay with this one commit pls :otter:

Borda:
> Not sure how to interpret the codecov failure, the PR doesn't add any "new" code. Is it a CI glitch? Should I do something about it?

it is a sanity check that the new code is covered, but it also has some false negative alarms when the coverage change is almost zero...

src/torchmetrics/detection/panoptic_quality.py
marcocaccin: The common functionality is being pulled out into a separate file according to what I mentioned in #1591 . This PR is the first of 2, so that you don’t get hit with an unreviewable one… if there’s a better option let me know!

Borda: I understand common, but while it is used only by one metric it seems rather specific than common :)

marcocaccin: Should I call it utils in this PR and rename it to common in the next for consistency? Alternatively all of the common code can stay in panoptic_quality.py and the modified implementation will import from there, but the asymmetry would be slightly confusing

Borda: it is not about the name, but what other function/metric would use it?

marcocaccin: The modified PQ metric that is mentioned in the issue and I’m implementing on the side of the standard PQ that’s already there

Borda: ok, I see now, sorry :innocent:

Borda (approved):

src/torchmetrics/detection/panoptic_quality.py
marcocaccin: No worries, it was clear we had a misunderstanding but I wasn't sure where 🙂

marcocaccin: Btw you had a very good reason to be confused, I am now too... https://github.com/Lightning-AI/metrics/issues/1591#issuecomment-1456985706. Let me know if you have a preference either way, we don't need to merge this PR if there's a preference in keeping just 1 metric with 2 flavours.

PR #1676, opened 2023-03-31
Classification: option to disable input formatting [wip]

### LINKED ISSUE #1526: Sigmoid function in PrecisionRecallCurve leads to information loss
## 🐛 Bug
Hello, first of all, thank you for the awesome library! I am a maintainer of the [Anomalib library](https://github.com/openvinotoolkit/anomalib), and we are using TorchMetrics extensively throughout our code base to evaluate our models.

The most recent version of TorchMetrics introduced some changes to the `PrecisionRecallCurve` metric, which are causing some problems in one of our components. The problems are related to the re-mapping of the prediction values to the [0,1] range by applying a sigmoid function.

### Some context
The goal of the models in our library is to detect anomalous samples in a dataset that contains both normal and anomalous samples. The task is similar to a classical binary classification problem, but instead of generating a class label and a confidence score, our models generate an anomaly score, which quantifies the distance of the sample to the distribution of normal samples seen during training. The range of possible anomaly score values is unbounded and may differ widely between models and/or datasets, which makes it tricky to set a good threshold for mapping the raw anomaly scores to a binary class label (normal vs. anomalous). This is why we apply an adaptive thresholding mechanism as a post-processing step. The adaptive threshold mechanism returns the threshold value that maximizes the F1 score over the validation set.

Our [adaptive thresholding class](https://github.com/openvinotoolkit/anomalib/blob/main/anomalib/utils/metrics/anomaly_score_threshold.py#L15) inherits from TorchMetrics' `PrecisionRecallCurve` class. After TorchMetrics computes the precision and recall values, our class computes the F1 scores for the range of precision and recall values, and finally returns the threshold value that corresponds to the highest observed F1 score.

### The problem
In the latest version of the `PrecisionRecallCurve` metric, the `update` method now re-maps the predictions to the [0, 1] range by [applying a sigmoid function](https://github.com/Lightning-AI/metrics/blob/v0.11.1/src/torchmetrics/functional/classification/precision_recall_curve.py#L181). As a result, the `thresholds` variable returned by `compute` is now not in the same domain as the original predictions, and the values are not usable for our purpose of finding the optimal threshold value.

In addition, the sigmoid function squeezes the higher and lower values, which leads to lower resolution at the extremes of the input range, and in some cases even information loss.

### To Reproduce

Here's an example to illustrate the problem. Let's say we have a set of binary targets and a set of model predictions in the range [12, 17]. Previously, the `PrecisionRecallCurve` metric would return the values of precision and recall for the different thresholds that occur naturally in the data. 

#### v0.10.3
```python
>>> from torchmetrics import PrecisionRecallCurve
>>> from torch import Tensor
>>>
>>> targets = Tensor([0, 0, 1, 0, 1, 1]).int()
>>> predictions = Tensor([12, 13, 14, 15, 16, 17])
>>>
>>> metric = PrecisionRecallCurve()
>>> metric.update(predictions, targets)
>>> precision, recall, thresholds = metric.compute()
>>>
>>> precision
tensor([0.7500, 0.6667, 1.0000, 1.0000, 1.0000])
>>> recall
tensor([1.0000, 0.6667, 0.6667, 0.3333, 0.0000])
>>> thresholds
tensor([14., 15., 16., 17.])
```

Given these outputs it is straightforward to obtain the F1 scores for the different threshold values and use this to find the optimal threshold that maximizes F1.

After the recent changes, the predictions are now re-mapped by the sigmoid function. While we can still compute the F1 scores, we can no longer find the value of the threshold that yields the highest F1 score, because the values of the `thresholds` variable are no longer in the same domain as the original predictions.

#### v0.11.1
```python
>>> from torchmetrics import PrecisionRecallCurve
>>> from torch import Tensor
>>>
>>> targets = Tensor([0, 0, 1, 0, 1, 1]).int()
>>> predictions = Tensor([12, 13, 14, 15, 16, 17])
>>>
>>> metric = PrecisionRecallCurve(task="binary")
>>> metric.update(predictions, targets)
>>> precision, recall, thresholds = metric.compute()
>>>
>>> precision
tensor([0.7500, 0.6667, 1.0000, 1.0000, 1.0000])
>>> recall
tensor([1.0000, 0.6667, 0.6667, 0.3333, 0.0000])
>>> thresholds
tensor([1.0000, 1.0000, 1.0000, 1.0000])
```
Note that the elements of the `thresholds` variable all appear as 1.0000 because the numerical differences between the threshold candidates are minimized due to the squeezing effect of the sigmoid.

It gets even worse when we increase the absolute values of the predictions to [22, 27]. The output of the sigmoid now evaluates to 1.0 for all predictions due to rounding, and the metric is not able to compute any meaningful precision and recall values.

#### v0.11.1
```python
>>> from torchmetrics import PrecisionRecallCurve
>>> from torch import Tensor
>>>
>>> targets = Tensor([0, 0, 1, 0, 1, 1]).int()
>>> predictions = Tensor([22, 23, 24, 25, 26, 27])
>>>
>>> metric = PrecisionRecallCurve(task="binary")
>>> metric.update(predictions, targets)
>>> precision, recall, thresholds = metric.compute()
>>>
>>> precision
tensor([0.5000, 1.0000])
>>> recall
tensor([1., 0.])
>>> thresholds
tensor(1.)
```

I guess this change was made to accommodate classical binary classification problems, where the predictions are generally confidence scores in the [0, 1] range, but I feel this is too restricting for other problem classes. Mathematically there is no reason why the precision-recall curve cannot be computed using predictions that fall outside of this range.

### Expected behavior

The re-mapping of the prediction values to [0,1] by applying a sigmoid function should be optional.

### Environment

- TorchMetrics 0.11.1 (pip)
- Python 3.8
- PyTorch 1.13.1

### LINKED ISSUE #1604: BinaryAccuracy() sometimes gives incorrect answers due to non-deterministic sigmoiding
## 🐛 Bug

`torchmetrics.classification.BinaryAccuracy` will apply a sigmoid to some inputs but not others leading to incorrect behavior.

## Details
The current behavior of BinaryAccuracy() is to apply a sigmoid transformation if the inputs are outside of [0, 1] before binarizing   

> If preds is a floating point tensor with values outside [0,1] range we consider the input to be logits and will auto apply sigmoid per element. 

i.e. 
y_hat = 1(sigmoid(z) >= threshold) if z outside [0, 1] 
y_hat = 1(z >= threshold) if z inside [0, 1] 

I assume z inside [0, 1] is checked for then entire batch (i.e. if one element of the batch is outside [0, 1] then we apply the sigmoid to everyone).

This will cause silent errors. In particular, if the user inputs logits then they expect the logits to always be sigmoided. However, it is totally possible for all of the logits to lie in [0, 1] for some batches in which case the input will **not** be sigmoided which will cause incorrect thresholding.

### To Reproduce

Here is a simple example. Support our network outputs logits.

```python
from torchmetrics.classification import BinaryAccuracy
from scipy.special import expit # expit = sigmoid
import numpy as np
import torch
```

This example should lead to a correct prediction
``` python
probability_thresh = 0.5 
logits = np.array([0.49]) # network output
target = np.array([1])

# logits of 0.49 give a probability of 0.62 indicating class 1, the correct prediction
expit(logits)
array([0.62010643])
int(expit(logits) >= probability_thresh) == target
True
```

`BinaryAccuracy()` however thinks it's an incorrect prediction~
``` python
# torchmetrics, however, thinks we have the inccorect prediction because it does NOT sigmoid the logits
ba = BinaryAccuracy(threshold=probability_thresh) 
ba.forward(preds=torch.tensor(logits), target=torch.tensor(target))
tensor(0.)
```

### Suggested Fix

I suggest adding an argument indicating whether or not the input predictions are sigmoided so the inputs are either always sigmoided or never sigmoided

### LINKED ISSUE #1989: AUROC reported wrong value when called differently
## 🐛 Bug

AUROC calculation is different when being called differently. 

### To Reproduce

Copy the following code and run. 

Note that when .update() is called, values that are greater than 1.0 will be scaled using the sigmoid activation, values that is less than 1.0 will not be scaled. This result in the compute step to compute with the wrong input values. However, directly calling the auroc will not rescale the value

<details>
  <summary>
```
pp = tensor([tensor([1.2]), tensor([0.9678]), tensor([1.7])])
tt = tensor([tensor([1]), tensor([0]), tensor([1])])

auroc = AUROC(num_classes=1, task="binary")
print("Forward:", auroc(pp, tt))


for p, t in zip(pp, tt):
    auroc.update(p, t)
    print(auroc.preds, auroc.target)

print("Update and Compute:", auroc.compute())
```
</summary>

</details>

### Expected behavior

The reported AUROC metrics for both cases should be the same.

### Environment

- TorchMetrics 1.0.3
- Python 3.9.16
- PyTorch 1.13.1

### LINKED ISSUE #2195: Conversion from logits to probabilities happens on a batch by batch basis
## 🐛 Bug

I'm using `BinaryAUROC` and calling `.update(logits, y)` after each batch (`validation_step`), expecting the following behavior:

"If preds has values outside [0,1] range we consider the input to be logits and will auto apply sigmoid per element."

I noticed some discrepancies in the results, and after investigating realized that the rule above happens on a batch by batch basis rather than across the entire dataset (it happens in `_binary_precision_recall_curve_format`, which is called by `update`). 

Therefore, if some batch has all logits in [0, 1], then the conversion doesn't happen for that batch.

My use case has batch size = 1, so this happens often.

### Expected behavior

I'd expect that the conversion would happen consistently across the entire dataset (i.e., when calling `.compute()`). If not that, this behavior should be documented more prominently.

### Environment

```
torchmetrics==1.0.3
```

### LINKED ISSUE #2329: `BinaryPrecisionRecallCurve` computes wrong value if used with logits, even though the docstring says this is supported
## 🐛 Bug

The object oriented `BinaryPrecisionRecallCurve` can compute substantially incorrect values if logits are passed as predictions instead of probabilities, even though the docstring says this is ok. The underlying functional version `binary_precision_recall_curve` seems to work correctly in both cases. Both versions attempt to convert the logits to probabilities by passing them through a sigmoid *if any values are outside of the range [0, 1]*. In the object oriented case this condition is incorrectly checked independently for each batch, rather than for the metric as a whole. Consequently some batches may have sigmoid applied to their scores, while others do not, resulting in an incorrect curve for the dataset as a whole.

### To Reproduce

<details>

import torch
import torchmetrics

b1 = torchmetrics.AUROC(task="binary")
b2 = torchmetrics.AUROC(task="binary")

score = torch.tensor([-1.0, 0.0, 1.0])
label = torch.tensor([1.0, 0.0, 1.0])

b1.update(score, label)  # pass all three score/label pairs in as a single batch

for s, l in zip(score, label):
    b2.update(torch.tensor([s]), torch.tensor([l]))  # pass score/label pairs in one at a time

assert b1.compute().item() == 0.5
assert b2.compute().item() == 1.0

</details>

### Expected behavior

I would expect both AUCs to equal 0.5, as computed with scikit-learn using `sklearn.metrics.roc_auc_score(label, score)`.

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source): 1.3.0.post0
- Python & PyTorch Version (e.g., 1.0): 3.10.13
- Any other relevant information such as OS (e.g., Linux): MacOS

### Additional context

The bug is on line https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/classification/precision_recall_curve.py#L165 -- as it is currently written the function `_binary_precision_recall_curve_format` should only be called on the full dataset, not on individual batches. Otherwise the behavior is wrong if some batches have all scores in the range [0, 1] but other batches do not.

Some possible solutions are: (1) update the docs not to allow for logits in the object oriented interface, since the behavior is correct for probabilities; (2) don't try to automatically infer whether to apply sigmoid -- my choice, but would be a breaking change (3) refactor `_binary_precision_recall_curve_format` and accept that if any values are found which require sigmoid, then all values from past batches need to have sigmoid applied (this would be tricky in the case where thresholds are specified because the scores are not kept around).

SkafteNicki:
## What does this PR do?

Fixes #1526
Fixes #1604
Fixes #1989 
Fixes #2195 
Fixes #2329 

<details>

</details>

- [x] accuracy.py
- [x] auroc.py
- [x] average_precision.py
- [ ] calibration_error.py
- [ ] cohen_kappa.py
- [x] confusion_matrix.py
- [ ] dice.py
- [ ] exact_match.py
- [x] f_beta.py
- [ ] group_fairness.py
- [x] hamming.py
- [ ] hinge.py
- [ ] jaccard.py
- [ ] matthews_corrcoef.py
- [ ] precision_fixed_recall.py
- [x] precision_recall.py
- [x] precision_recall_curve.py
- [ ] ranking.py
- [ ] recall_fixed_precision.py
- [x] roc.py
- [x] specificity.py
- [ ] specificity_sensitivity.py
- [x] stat_scores.py

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?


Borda:
@SkafteNicki how is it going here? :rabbit:

samet-akcay:
Hi @SkafteNicki, do you guys have any update on this one please?

Borda:
> do you guys have any update on this one please?

We need to address failing checks

Borda:
@SkafteNicki how is it going here? :rabbit:

src/torchmetrics/functional/classification/precision_recall_curve.py
```diff
@@ -329,6 +349,16 @@ def binary_precision_recall_curve(
             Specifies a target value that is ignored and does not contribute to the metric calculation
         validate_args: bool indicating if input arguments and tensors should be validated for correctness.
             Set to ``False`` for faster computations.
+        input_format: str or bool specifying the format of the input preds tensor. Can be one of:
```
Borda, line 352: yes, this looks good to me...
cc: @Lightning-AI/core-metrics @awaelchli

Borda:
regarding #1989 I would consider drop auto as it is if you have first inputs accidentally in range (0, 1) but your true nature is logits so you would end up with treading inputs wrongly as prob... right?

justusschock:
@SkafteNicki what's left here? :)

justusschock:
I'd probably name the things a bit differently and stick to string arguments only e.g no True and instead of False have a "none" arg

the resulting docstring would then be
```
input_format: str or bool specifying the format of the input preds tensor. Can be one of:
            - ``'auto'``: automatically detect the format based on the values in the tensor. If all values
                are in the [0,1] range, we consider the tensor to be probabilities and only thresholds the values.
                If all values are non-float we consider the tensor to be labels and does nothing. Else we consider the
                tensor to be logits and will apply sigmoid to the tensor and threshold the values.
            - ``'probs'``: preds tensor contains values in the [0,1] range and is considered to be probabilities. Only
                thresholding will be applied to the tensor and values will be checked to be in [0,1] range.
            - ``'logits'``: preds tensor contains values outside the [0,1] range and is considered to be logits. We
                will apply sigmoid to the tensor and threshold the values before calculating the metric.
            - ``'labels'``: preds tensor contains integer values and is considered to be labels. No formatting will be
                applied to preds tensor.
            - ``none'``: will disable all input formatting. This is the fastest option but also the least safe.
```

Borda:
@SkafteNicki, how is it going? What is missing... :flags:

Borda:
setting as ready PR just to enable full testing...

idc9:
Thanks for working on the bug in #1604! I suspect this bug is silently affecting people currently using torchmetrics.

It looks like the current solution is a new argument `input_format`and gives the user the option to specify if they are providing probabilities, logits, etc  (see below documentation). This *almost* fixes the problem, but this argument defaults to `input_format='auto'` which will cause exactly the bug discussed in #1604 (i.e. if your logits happen to all be in [0, 1] this will incorrectly fail to convert them to probabilities). 

I see two options

1) Change the default to `input_format='logits'` (or `probs` though I think logits is the most common input). This default options is explicit about behavior that is a but subtle and  will prevent the bug from happening by defaults

2) Get rid of the `auto` option all together. The `auto` option will always be buggy if the user inputs logits. 

I personally see no reason to include the `auto` option and would vote for option 2. 

```
input_format: str specifying the format of the input preds tensor. Can be one of:

    - ``'auto'``: automatically detect the format based on the values in the tensor. If all values
        are in the [0,1] range, we consider the tensor to be probabilities and only thresholds the values.
        If all values are non-float we consider the tensor to be labels and does nothing. Else we consider the
        tensor to be logits and will apply sigmoid to the tensor and threshold the values.
    - ``'probs'``: preds tensor contains values in the [0,1] range and is considered to be probabilities. Only
        thresholding will be applied to the tensor and values will be checked to be in [0,1] range.
    - ``'logits'``: preds tensor contains values outside the [0,1] range and is considered to be logits. We
        will apply sigmoid to the tensor and threshold the values before calculating the metric.
    - ``'labels'``: preds tensor contains integer values and is considered to be labels. No formatting will be
        applied to preds tensor.
    - ``'none'``: will disable all input formatting. This is the fastest option but also the least safe.
```

SkafteNicki:
@idc9 thanks for giving your opinions on this issue.
I already laid out on slack some of my opinions, but just for transparency here are the two main reasons for having `auto` as the default:
1. To keep everything backwards compatible 
2. To say *"The auto option will always be buggy if the user inputs logits."* is simply wrong. For reasonably sized input, any model should really output logit values outside the [0,1] range which after sigmoid transformation corresponds to probabilities in the [0.5, 0.73] range. If a model is only outputting values in the 0.5-0.73 probability range, this will mean it is never going to predict the negative class, and it is never confident about the positive class. Both seems highly unlikely for any well trained model.

NoahAtKintsugi:
@SkafteNicki In cases where GPU memory limits batch sizes to be very small, there will be no "reasonably-sized input", so "auto" will be buggy. What about not having "auto", but rather having only "probs" and "logits", with "probs" being the default and verifying that its input is in [0, 1]? This will be backwards compatible, except raise an error in the case that the old version was doing the wrong thing. The error message can suggest the user might mean `input_format = "logits"`.

idc9:
Sorry I should have been more careful with my language! Let me rephrase  “The auto option will always be buggy if the user inputs logits” to “The ‘auto’ option leaves open the possibility for the bug to happen when the user inputs logits”.

Since the auto option leaves open the possibility of this bug occurring I suspect we don’t we want to preserve backwards compatibility. This is a subtle/unexpected issue many users won’t immediately realize it can be an issue. Given it’s straightforward for the user to specify logits/prob, I’m curious if there is any case where ‘auto’ is useful (i.e. should it just be removed)?

As @SkafteNicki  pointed out the logits bug will only occur if all of the predicted probabilities lie in [0.5, 0.73]. It’s true this is likely not the usual scenario but it probably does come up. A few examples where it may be more common include: small batches, early phases of model training, difficult applications where hold-out probabilities are (sadly) closeish to 0.5, unbalanced classes, situations where your label prediction rule thresholds the probability at some value different from 0.5.

SkafteNicki:
@NoahAtKintsugi and @idc9 Alright, I am willing to give me on this, but we are then going to break backwards compatibility, which I do not take as a light issue. If we do this change we need to do it right. You are probably right that it is better to break stuff now and then fix this issue that may be a potential hidden problem for some users.

@Borda and @justusschock please give some opinions on this. Do we 
* continue with `input_format="auto"` as the default keeping everything backwards compatibility
* set `input_format="auto"` for v1.4 (next release), include user warning that it will be changing to `input_format="logits"/"probs"` (which one we choose does not matter to me) in v1.5 
* something else...

Borda:
> * set `input_format="auto"` for v1.4 (next release), include user warning that it will be changing to `input_format="logits"/"probs"` (which one we choose does not matter to me) in v1.5

I think due to stability, we may add a warning but being having a hard switch from 2.0
cc: @lantiga

idc9:
Defaulting to `logits` might make the most sense. I suspect is that it's more common for users to default to outputting logits than outputting probabilities (e.g. https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html, https://github.com/huggingface/pytorch-image-models/blob/main/train.py, https://github.com/pytorch/examples/blob/main/imagenet/main.py) probably since this avoids a small bit of extra computation.

Borda:
@SkafteNicki, how is this one going? :)

Borda:
@SkafteNicki how about the list in the description move to an issue and start to merging smaller PRs with just a few fixed metrics so we keep it rolling and laos eventually others can contribute... also we would know we can'o't release new features release until all is done?

PR #1762, opened 2023-05-06
Add `MultitaskWrapper`

### LINKED ISSUE #1741: Add a wrapper adapted to multi-task learning
## 🚀 Feature

Add a wrapper adapted to multi-task learning. Given a list of model outputs and a list of labels, it should be able to compute a different metric (or several different metrics) on each output.

### Motivation

When doing multi-task learning, we have several outputs, and the metric that we want to compute on each output is generally different. 

For instance, in image detection with K classes, the output could consist of 4 neurons for bounding box location (pos_x, pos_y, height, width), and K neurons for the class of the detected object. We would like to have a way to compute some regression metrics (like MSE, MAE, etc...) on the outputs corresponding to the bounding box locations, and to compute some classification metrics (like Accuracy) on the outputs corresponding to the class. 

An idea could be to simply generalize `MultioutputWrapper` to use a list of `base_metric` rather than a single, duplicated, `base_metric`. But in the previous example, the output and the label are `tuple[Tensor, Tensor]` (the first element is a `Tensor` of shape [4], the second element is a `Tensor` of shape [K]), and `MultioutputWrapper` works with `Tensor` only.

### Pitch

Add a `MultitaskWrapper` class that would take a list of `Metric`s at initialization (or maybe a `MetricCollection`, or an `Iterable[Metric]`), and whose `update` signature would look like:
```python
def update(self, preds: Iterable[Tensor], targets: Iterable[Tensor]) -> None:
```

ValerianRey:
## What does this PR do?

Fixes #1741 

<details>

</details>

<details>
  <summary>PR review</summary>

This is the first draft of my work on #1741. Before going further in the development, I want to discuss several points:

1. I made the choice to make this wrapper work entirely with dictionaries. In the `__init__` method, `task_metrics` is a dict associating metrics to each task. In the `update` method, `task_preds` and `task_targets` are dictionaries associating a pred and a target to each task. Lastly, the output of the `compute` method is also a dictionary associating a computed metric to each task. Arguably, we could make it work entirely with Sequences (of metrics, of preds, of targets, of outputs). Another alternative would be to use a Sequence of metrics along with a Sequence of str in the `__init__` method, for the metrics and the task names. The preds and targets could be Sequences of Tensors, and the output could be a Dictionary, mapping each task name to the computed output. Lastly, we could have both options supported, but I think this will make the code messy and the interface harder to understand. 
So I think the current version (using dicts only) is the best, but I want an external opinion on this before going further.

2. I have a hard time understanding the `full_state_update` bool defined in `Metric`. What should I do with it?
3. In `MultioutputWrapper` (which has many common points with `MultitaskWrapper`), I have seen that you have overriden the following methods as identity, with the docstring "Overwrite to do nothing.":
    - `_wrap_update` ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/wrappers/multioutput.py#L143-L145))
    - `_wrap_compute` ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/wrappers/multioutput.py#L147-L149))

    What was the intention, and should I do the same in `MultitaskWrapper`?
4. I don't know whether I need to override `forward` or not (it is overriden in `MultioutputWrapper` ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/wrappers/multioutput.py#L123-L135)))
5. The `plot` method is greatly inspired from `MetricCollection.plot` ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/collections.py#L492-L577)). I decided to only support plotting each task's metrics separately, as they are heterogeneous in general. Arguably, each `MetricCollection` in the `task_metrics` would need to have a `together` bool parameter associated to it when calling `plot`, but that's very messy to implement and probably useless (the users could simply call `plot` separately and with the right value of `together`, for each metric in `task_metric`).

Please do not hesitate to give negative feedback!
</details>

<details>
  <summary>Next steps</summary>

- [x] Add unit tests
- [x] Add docstring for `plot`
- [x] Make the necessary changes based on answers to 1., 2., 3., 4. and 5.
- [x] Make the necessary changes based on feedback

</details>


ValerianRey:
@SkafteNicki I'd love to have some of your feedback on this before going further!

src/torchmetrics/wrappers/multitask.py
```diff
@@ -0,0 +1,110 @@
...
+        task_metrics: Dict[str, Union[Metric, MetricCollection]],
+    ) -> None:
+        self._check_task_metrics_type(task_metrics)
+        super().__init__()
+        self.task_metrics = task_metrics
+
+    @staticmethod
+    def _check_task_metrics_type(task_metrics: Dict[str, Union[Metric, MetricCollection]]) -> None:
+        if not isinstance(task_metrics, dict):
+            raise TypeError(f"Expected task_metrics to be a dict. Found task_metrics = {task_metrics}")
```
SkafteNicki, line 56:
```suggestion
            raise TypeError(f"Expected argument `task_metrics` to be a dict. Found task_metrics = {task_metrics}")
```

src/torchmetrics/wrappers/multitask.py
```diff
@@ -0,0 +1,110 @@
...
+
+        for metric in task_metrics.values():
+            if not (isinstance(metric, (Metric, MetricCollection))):
+                raise TypeError(
+                    "Expected each task's metric to be a Metric or a MetricCollection. "
+                    f"Found a metric of type {type(metric)}"
+                )
+
+    def update(self, task_preds: Dict[str, Tensor], task_targets: Dict[str, Tensor]) -> None:
```
SkafteNicki, line 65:
```suggestion
    def update(self, *args: Tuple[Dict[str, Tensor]], **kwargs: Dict[str, Dict[str, Tensor]) -> None:
```
I wonder if we can generalize to `*args, **kwargs`. The wast majority of all torchmetrics metrics follows the `preds, target` naming, but a few does not. If we could do this, it would be cool, but if not I am okay with that.

src/torchmetrics/wrappers/multitask.py
```diff
@@ -0,0 +1,110 @@
...
+
+    is_differentiable = False
+
+    def __init__(
+        self,
+        task_metrics: Dict[str, Union[Metric, MetricCollection]],
+    ) -> None:
+        self._check_task_metrics_type(task_metrics)
+        super().__init__()
+        self.task_metrics = task_metrics
```
SkafteNicki, line 51:
```suggestion
        self.task_metrics = nn.ModuleDict(task_metrics)
```
I am 99% sure we need to do a wrapping here, because if not the other metrics will not be considered "children" of the wrapper and therefore calling `.cuda()`, `.to()` etc. method will not move the states of the sub-modules to the correct device

src/torchmetrics/wrappers/multitask.py
```diff
@@ -0,0 +1,110 @@
...
+         >>> regression_preds = torch.tensor([3.0, 5.0, 2.5])
+         >>> preds = {"Classification": classification_preds, "Regression": regression_preds}
+         >>> metrics = MultitaskWrapper({
+         ...     "Classification": BinaryAccuracy(),
+         ...     "Regression": MeanSquaredError()
+         ... })
+         >>> metrics.update(preds, targets)
+         >>> metrics.compute()
+         {'Classification': tensor(0.3333), 'Regression': tensor(0.8333)}
+    """
```
SkafteNicki, line 41: An second example using a metric collection would be nice to communicate that the output in that case will be a nested dict structure.

SkafteNicki (commented):
Already looking good.
Of cause the testing is missing, but I think the interface is nice :]

src/torchmetrics/wrappers/multitask.py
ValerianRey: I just quickly tested it and you're right, I'll fix that!

src/torchmetrics/wrappers/multitask.py
ValerianRey: I don't understand the signature that you proposed, could you please explain it?

From my understanding, if we want to support some metrics which take arbitrary arguments in their `update` method, we need the `update` method of `MultitaskWrapper` to have the following signature:
```python
def update(self, task_positional_args: [Dict[str, Any]], task_kwargs: Dict[str, Dict[str, Any]) -> None:
```
i.e. we allow some arbitrary positional args and key-word args, different for each task.
I think a problem with this would be the use-case. We want the users to be able to do:
```python
preds = model(x)  # preds is of the form {"Classification": tensor(...), "Regression": tensor(...)}
multitask_metric.update(preds, targets)
```
with the above signature, we would have to do:
```python
preds = model(x)  # preds is of the form {"Classification": tensor(...), "Regression": tensor(...)}
positional_args = {
    "Classification": [preds["Classification"], targets["Classification"]], 
    "Regression": [preds["Regression"], targets["Regression"]]
}
multitask_metric.update(positional_args, {})
```
This makes this wrapper have a different usage than single-task metrics, which is precisely what I want to avoid by implementing the multi-task metric (so that we can change between single-task and multi-task problems without changing the code).

Again, I'm not sure that I understood what you meant, so please correct me if you had a different idea in mind!

ValerianRey:
Thanks for the feedback! I made most of the changes that you suggested!

src/torchmetrics/wrappers/multitask.py
SkafteNicki: Cool, then lets just keep it as it is, but then lets specify in the documentation docstring that it is only metrics with the `preds, target` signature that is supported

ValerianRey:
> @ValerianRey cool. Could you try to write some unit tests for the wrapper? I do not think that is should be that hard as the metric is not that complex. We basically just need to check that inputs gets correctly send to their respective task.

Yes I will do that this week!
Could you please have a look at the questions that I asked in the original post of this PR and try to answer those for which you know the answer / have an opinion? It would help me a lot!

SkafteNicki:
Answers to your questions

> I made the choice to make this wrapper work entirely with dictionaries. In the __init__ method, task_metrics is a dict associating metrics to each task. In the update method, task_preds and task_targets are dictionaries associating a pred and a target to each task. Lastly, the output of the compute method is also a dictionary associating a computed metric to each task. Arguably, we could make it work entirely with Sequences (of metrics, of preds, of targets, of outputs). Another alternative would be to use a Sequence of metrics along with a Sequence of str in the __init__ method, for the metrics and the task names. The preds and targets could be Sequences of Tensors, and the output could be a Dictionary, mapping each task name to the computed output. Lastly, we could have both options supported, but I think this will make the code messy and the interface harder to understand. So I think the current version (using dicts only) is the best, but I want an external opinion on this before going further.

Agree that this is a fine interface.

> I have a hard time understanding the full_state_update bool defined in Metric. What should I do with it?

Instead of setting this, please add a custom `forward` method:

```python
def forward(self, task_preds: Dict[str, Tensor], task_targets: Dict[str, Tensor]) -> Dict[str, Any]:
    return {metric(task_preds[task_name], tast_targets[task_name]) for task_name, metric in self.task_metrics.items()}:
```

The default forward function is not needed here because all computations are handled by the task specific metrics. 

> In MultioutputWrapper (which has many common points with MultitaskWrapper), I have seen that you have overriden the following methods as identity, with the docstring "Overwrite to do nothing.":
        _wrap_update ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/wrappers/multioutput.py#L143-L145))
        _wrap_compute ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/wrappers/multioutput.py#L147-L149))
> What was the intention, and should I do the same in MultitaskWrapper?

`_wrap_update` and `_wrap_compute` for normal metrics takes care of keeping the states up to date and making sure that everything gets correctly synced in distributed settings. As this wrapper (and `MultioutputWrapper`) does not have states of their own, they rely on other metrics, we can safely override these because all the logic is dealt with the child metrics.
Short answer: yes, please overwrite them to do nothing.

> I don't know whether I need to override forward or not (it is overriden in MultioutputWrapper ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/wrappers/multioutput.py#L123-L135)))

See previous answer above

> The plot method is greatly inspired from MetricCollection.plot ([code](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/collections.py#L492-L577)). I decided to only support plotting each task's metrics separately, as they are heterogeneous in general. Arguably, each MetricCollection in the task_metrics would need to have a together bool parameter associated to it when calling plot, but that's very messy to implement and probably useless (the users could simply call plot separately and with the right value of together, for each metric in task_metric).

I agree that this is the way to go. The plot interface is meant for quick and dirty plotting of metrics, not to produce highly custom one. If users want that they can probably figure out to code it themselves :]

SkafteNicki:
A lot of the observations made here probably indicate that we should make a common interface for wrappers if possible since there seems to be some duplication going on. Lets tackle that in a future PR.

ValerianRey:
> Instead of setting this, please add a custom `forward` method:

I added the `forward` method, with a docstring and a comment explaining this implementation choice, with what I understood. I hope that this is correct! Also, I'm not very familiar with the `@torch.jit.unused` decorator, and I'm unsure whether I should add it (like in `Metric.forward`, `MultioutputWrapper.forward` and `MetricCollection.forward`) or not (like in `ClasswiseWrapper.forward`, `Running.forward` and `MetricTracker.forward`).

The last 2 steps for me are the docstring of `plot` and the unit tests (I already did some quick tests locally but nothing clean yet, so I didn't commit them yet).

SkafteNicki:
> > Instead of setting this, please add a custom `forward` method:
> 
> I added the `forward` method, with a docstring and a comment explaining this implementation choice, with what I understood. I hope that this is correct! Also, I'm not very familiar with the `@torch.jit.unused` decorator, and I'm unsure whether I should add it (like in `Metric.forward`, `MultioutputWrapper.forward` and `MetricCollection.forward`) or not (like in `ClasswiseWrapper.forward`, `Running.forward` and `MetricTracker.forward`).
> 
> The last 2 steps for me are the docstring of `plot` and the unit tests (I already did some quick tests locally but nothing clean yet, so I didn't commit them yet).

Cool, the code still looks good to me. Going to wait for the unittest before reviewing the code again :]

SkafteNicki (approved):
LGTM :]

tests/unittests/wrappers/test_multitask.py
```diff
@@ -0,0 +1,182 @@
...
+            "Classification": BinaryAccuracy(),
+        }
+    )
+
+    with pytest.raises(ValueError, match="Expected *"):
+        multitask_metrics.update(wrong_key_preds, _multitask_targets)
+    with pytest.raises(ValueError, match="Expected *"):
+        multitask_metrics.update(_multitask_preds, wrong_key_targets)
+    with pytest.raises(ValueError, match="Expected *"):
+        wrong_key_multitask_metrics.update(_multitask_preds, _multitask_targets)
```
Borda, line 99: Could the matching message be more specific?

Borda (approved):

tests/unittests/wrappers/test_multitask.py
ValerianRey: Fixed

PR #1832, opened 2023-06-08
Revert MAP implementation to `pycocotools` backend

### LINKED ISSUE #1024: MeanAveragePrecision is slow

### LINKED ISSUE #1164: Discrepancy between values with metrics and detectron2 cocoeval
## 🐛 Bug

Training a model with detectron2 and evaluating it with the built-in cocoeval results in different mAP estimates for both `box` and `segm` metric cases.

### To Reproduce

1. Train model with detectron2
2. Use CocoEval to generate mAP, mAR, etc 
3. Use torchmetrics to calculate mAP, mAR, etc
4. Compare for a detection and segmentation models

### Expected behavior

The values generated to be the same given both use underlying pycocotools (at least for segm). Or up to some floating point precision.

See below for `box` case:

<html>
<body>

metricName | torchmetrics | Detectron2 Cocoeval (x100)
-- | -- | --
map | 0.6738 | 66.01083445431875
map_50 | 0.9600 | 97.36921342328598
map_75 | 0.8268 | 83.13779622663151
map_small | 0.5943 | 64.11872772143319
map_medium | 0.6192 |  64.02383668112452
map_large | 0.8431 |  77.85958548685628
mar_1 | 0.7376
mar_10 | 0.7394
mar_100 | 0.7394
mar_small | 0.6466
mar_medium | 0.7293
mar_large | 0.8795
map_per_class | 0.6738 | NaN, NaN, NaN, NaN,  66.01083
mar_100_per_class | 0.7394

</body>
</html>

### Environment

- TorchMetrics version 9.3 (`pip`)

SkafteNicki:
## What does this PR do?

Fixes #1024
Fixes #1164

Reverts the `MeanAveragePrecision` implementation back to its v0.6 version for two primary reasons:
* Our current implementation is orders of magnitude slower and even with improvements in https://github.com/Lightning-AI/torchmetrics/pull/1389 it is still slower
* Nobody on the primary team is actually able to maintain the metric because we do not have experience with object detection. Therefore it is better to rely on `pycocotools` during the major calculations and we "just" provide an interface

The consequence for now is that `pycocotools` will be (re-)introduced as an require dependency for the metric.

TODO list:
- [x] Revert implementation
- [x] Make sure implementation supports argument `iou_type` (introduced after v0.6)
- [x] Make sure implementation supports argument `iou_thresholds, rec_thresholds, max_detection_thresholds` (introduced after v0.6)
- [x] Benchmark implementation to make sure it is still faster
- [x] Move from static testing to dynamic testing (for most tests) = call `pycocotools` directly during testing to check if output is correct
- [ ] ~Check if implementation fixes https://github.com/Lightning-AI/torchmetrics/issues/1774~
- [ ] ~Check if implementation fixes https://github.com/Lightning-AI/torchmetrics/issues/1793~
- [ ] ~Check if implementation fixes https://github.com/Lightning-AI/torchmetrics/issues/1184~
- [x] Check if implementation fixes https://github.com/Lightning-AI/torchmetrics/issues/1164

Benchmark results below (measured on my laptop)

Time | This PR | Master | PR https://github.com/Lightning-AI/torchmetrics/pull/1389 (improves current implementation) |
------|--------|-------|---------------------------------------------------------------------------------------------------|
Total time in init | 0.0007 | 0.0010 | 0.0012
Total time in update | 0.2445 | 0.2472 | 0.2769
Total time in compute | 10.5286 | 324.6941 | 220.8867

As it been stated many times the old implementation is orders of magnitude faster than our current implementation.

<details>
  <summary>Benchmark code</summary>

```python
import time

import torch

try:
    from torchmetrics.detection import MeanAveragePrecision
except ImportError:
    from torchmetrics.detection import MAP
    MeanAveragePrecision = MAP

total_time = {}

class UpdateTime:
    def __init__(self, name) -> None:
        self._name = name

    def __enter__(self):
        self._start_time = time.perf_counter()

    def __exit__(self, exc_type, exc_val, exc_tb):
        end_time = time.perf_counter()
        if self._name in total_time:
            total_time[self._name] += end_time - self._start_time
        else:
            total_time[self._name] = end_time - self._start_time
        return True

def generate(n):
    boxes = torch.rand(n, 4) * 1000
    boxes[:, 2:] += boxes[:, :2]
    labels = torch.randint(0, 10, (n,))
    scores = torch.rand(n)
    return {"boxes": boxes, "labels": labels, "scores": scores}

with UpdateTime("init"):
    map = MeanAveragePrecision()

for _batch_idx in range(100):
    with UpdateTime("update"):
        detections = [generate(100) for _ in range(10)]
        targets = [generate(10) for _ in range(10)]
        map.update(detections, targets)

with UpdateTime("compute"):
    map.compute()

for name, time in total_time.items():
    print(f"Total time in {name}: {time}")
```

</details>

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
:books: Documentation preview :books:: https://torchmetrics--1832.org.readthedocs.build/en/1832/


Borda:
can we rather preserve both so have `MAPCoco`?

SkafteNicki:
> can we rather preserve both so have `MAPCoco`?

Yes, we could do that, but then I would prefer we just add a `backend="pycocotools"/"pytorch"` argument to the class. That said, I am in the process of improving the testing and if the old implementation does not pass, then I would propose to still remove it.

CHANGELOG.md
```diff
@@ -149,6 +149,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Changed calculation in `PearsonCorrCoeff` to be more robust in certain cases  ([#1729](https://github.com/Lightning-AI/torchmetrics/pull/1729))
 
+
+- Changed `MeanAveragePrecision` to `pycocotools` backend ([#1832](https://github.com/Lightning-AI/torchmetrics/pull/1832))
```
Borda, line 153: I thought we would keep both backends...

Borda (approved):

justusschock (approved):

ddelange:
hi 👋  does this PR include breaking API changes?

can we still `from torchmetrics.detection.mean_ap import MeanAveragePrecision` and the API will be the same as before, just with the pycocotools backend?

Borda:
> can we still `from torchmetrics.detection.mean_ap import MeanAveragePrecision` and the API will be the same as before, just with the pycocotools backend?

yes, it shall be compatible :flamingo: 
cc: @SkafteNicki

adamjstewart:
> no breaking changes :]

One breaking change :]

The following code:
```python
class Foo(LightningModule):
    def __init__(self):
        self.metric = MeanAveragePrecision()

    def on_validation_epoch_end(self):
        self.log_dict(self.metric.compute())
```
used to work in 0.11.4. However, this PR adds a new `classes` key-pair to the metrics dict containing a list of all observed classes. `log_dict` now complains about this entry:
```
ValueError: `self.log(classes, tensor([0, 1], dtype=torch.int32))` was called, but the tensor must have a single element. You can try doing `self.log(classes, tensor([0, 1], dtype=torch.int32).mean())`
```
Here, `tensor([0, 1])` is the `classes` value. Current workaround is to delete this key from the dict before logging, but it would be good to do this automatically.

adamjstewart:
Good catch, thanks!

ddelange:
No worries, I happened to know because I hit it in in a recent [PR](https://github.com/autogluon/autogluon/pull/3190/files#diff-b23bc4f5266e60588f3af28e99b8b77c74400fdf88c337aaad85f4d9ef6ac80e) during a particularly representative case of [dephell](https://medium.com/knerd/the-nine-circles-of-python-dependency-hell-481d53e3e025)

SkafteNicki:
@adamjstewart @ddelange 
Yeah, I was probably too fast with saying no breaking changed :/
Would you prefer that we introduced an argument `include_classes` (or something similar) when initializing the class to enable/disable the inclusion of the `class` key.

For now you can also get around by a simple subclass:
```python
from torchmetrics.detection.mean_ap import MeanAveragePrecision as MAP
class MeanAveragePrecision(MAP):
    def compute(self):
        res = super().compute()
        res.pop("classes")
        return res
```

SkafteNicki:
> An argument would be good. I would vote for it to default to False so that the above example works out-of-the-box. I can submit a PR if you'd like.

Please feel free to open a PR. I would prefer if we set the default to be `True` such that we do not break backwards compatible once again.

adamjstewart:
If it defaults to True then the above code will continue to not work. Are you sure?

PR #1886, opened 2023-07-06
Add `EvaluationDistributedSampler` and examples on distributed evaluation

SkafteNicki:
## What does this PR do?

Fixes https://github.com/Lightning-AI/torchmetrics/issues/1338

The original issue is about if we should implement a [join context](https://pytorch.org/tutorials/advanced/generic_join.html) such that metrics could be evaluated on uneven number of samples in distributed settings. Just to remind, we normally discourage users from evaluating in distributed because the default [distributed sampler](https://github.com/pytorch/pytorch/blob/main/torch/utils/data/distributed.py) from Pytorch will add additional samples to make all processes do even work, which messes with results.

After investigating this issue, it seems that we do not need a join context at all due to the custom synchronization we have for metrics. To understand this we need to look at the two different states we can have: tensor state and list of tensor states.

1. For tensor states the logic is fairly simple: even if rank 0 is evaluated on more samples or more batches than rank 1, we still only need to do one all-gather operation regardless of how many samples/batches each rank has seen.
2. For list states we need are saved by the custom logic we have. Imaging that rank 0 state is a list of two tensors `[t_01, t_02]` and rank 1 state is a list of one tensor `[t_11]` (rank 0 have seen one more batch than rank 1). We list states are encountered internally we make sure to concatenate the states into one tensor to not need to call `allgather` for each tensor in the list
https://github.com/Lightning-AI/torchmetrics/blob/879595de67ab35d891a227e3254e0b5a26a050f0/src/torchmetrics/metric.py#L418-L419
such after this each state is a single tensor `t_0` and `t_1` but clearly `t_0.shape != t_1.shape`. Again, internally we deal with this by padding to same size and then doing a all gather:
https://github.com/Lightning-AI/torchmetrics/blob/879595de67ab35d891a227e3254e0b5a26a050f0/src/torchmetrics/utilities/distributed.py#L136-L148

Thus in both cases, even if one rank sees more samples/batches, we still do the same number of distributed operations per rank, which should mean that everything works. 

To highlight this feature of TM this PR does a couple of things:
* Introduce a new `EvaluationDistributedSampler` that does not add extra samplers. Thus, users can use this as a drop in replacement for any `DistributedSampler` if they want to do proper distributed evaluation (else they just need to secure that number of samples are even divisible by the number of processes).
* Add unittests that supports the above
* Add example on how to do this distributed evaluation in both lightning + standard torch

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
:books: Documentation preview :books:: https://torchmetrics--1886.org.readthedocs.build/en/1886/


examples/distributed_evaluation.py
```diff
@@ -0,0 +1,191 @@
...
+    x, y = dataset.tensors
+    preds = model(x)
+    return (preds.argmax(dim=1) == y).float().mean()
+
+
+def use_lightning(
+    model: Module, dataset: Dataset, batch_size: int, use_standard: bool, num_processes: int, gpu: bool
+) -> None:
+    """Use lightning to evaluate a model on a dataset."""
+    from lightning.pytorch import LightningModule, Trainer
```
Borda, line 69:
```suggestion
    if module_available("lightning"):
        from lightning.pytorch import LightningModule, Trainer
    else:
        from pytorch_lightning import LightningModule, Trainer
```

examples/distributed_evaluation.py
```diff
@@ -0,0 +1,191 @@
...
+        accelerator="cpu" if not gpu else "gpu",
+    )
+
+    res = trainer.test(model)
+    manual_res = calculate_accuracy_manually(dataset, model)
+    print(manual_res)
+    if torch.allclose(torch.tensor(res[0]["test_acc"]), manual_res):
+        print("success! result matched manual calculation")
+    else:
+        print("failure! result did not match manual calculation")
```
Borda, line 110: shall this rather we logging or `exit()`?

examples/distributed_evaluation.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def use_torch(
+    model: Module, dataset: Dataset, batch_size: int, use_standard: bool, num_processes: int, gpu: bool
+) -> None:
+    """Use torch.distributed to evaluate a model on a dataset."""
+    mp.spawn(_use_torch_worker_fn, nprocs=2, args=(model, dataset, batch_size, use_standard, num_processes, gpu))
+
+
+def main() -> None:
```
Borda, line 163: let's consider using `fire` or `jsonargapsrse` as CLI for examples... :)

Borda:
@SkafteNicki, how is it going here? do you think we could land it for the next 1.1 release...

examples/distributed_evaluation.py
SkafteNicki: but then user would need to install those just to run the example

Borda (approved):

src/torchmetrics/utilities/distributed.py
```diff
@@ -146,3 +147,65 @@ def gather_all_tensors(result: Tensor, group: Optional[Any] = None) -> List[Tens
...
+        shuffle: bool = True,
+        seed: int = 0,
+        drop_last: bool = False,
+    ) -> None:
+        # From:
+        # https://github.com/pytorch/pytorch/issues/25162#issuecomment-1227647626
+        super().__init__(dataset=dataset, num_replicas=num_replicas, rank=rank, shuffle=shuffle, seed=seed)
+
+        len_dataset = len(self.dataset)  # type: ignore[arg-type]
+        if not self.drop_last and len_dataset % self.num_replicas != 0:
```
justusschock, line 207: the issue with this that it wouldn't necessarily work with validation, since not all ranks would reach the same distributed function calls and therefore time out which would kill the entire process. Also this would never work with FSDP, since some ranks have a batch more and for fsdp, not all processes would reach the forward syncing points also resulting in timeouts.

justusschock (commented):
can we please add tests for validation and training as well? And maybe an fsdp test? Also some notes on caveats might be good to add to the sampler docs

justusschock:
calling @awaelchli for distributed review :)

SkafteNicki:
> can we please add tests for validation and training as well? And maybe an fsdp test? Also some notes on caveats might be good to add to the sampler docs

You are right that we need to test this feature better to clearly state the limitations.
I am going to remove it from the 1.1 milestone to future because it is not important to get done right now.

SkafteNicki:
Converted to draft until better tested.

src/torchmetrics/utilities/distributed.py
awaelchli: I agree that in the context of Lightning this wouldn't work well, as it does not support Join (https://github.com/Lightning-AI/lightning/issues/3325)
FSDP also doesn't support join afaik (https://github.com/pytorch/pytorch/issues/64683)

But outside Lightning, and taking FSDP out of the equation, I agree this can work and is a good utility to have IMO. It also suits the metric design well, since synchronization is only necessary when all processes have finished collecting their statistics and `.compute()` can be called.

Borda:
@SkafteNicki, what is missing here to make it land? :chipmunk:

Borda:
@SkafteNicki this seems to be pending for a while, shall we continue? :rabbit:

SkafteNicki:
Closing as while this would be a nice to have it is probably too limited in scope to provide enough value + it is not a core feature of torchmetrics.

PR #2093, opened 2023-09-20
Fix RuntimeError: CUDA out of memory. when using lpips

### LINKED ISSUE #2078: Some metrics save all targets and predictions in the buffer. For large datasets, this may lead to a large memory footprint
## 🐛 Bug

Hi everyone! I am currently working with your libraries to manage my deep-learning projects. However, I have run into some metrics-related challenges, especially when dealing with large data sets. The problem arises because the used metrics are storing the entire data set in memory during training, validation, or testing, which can cause memory overflow issues. Specifically, I've been experiencing this problem with the SpectralAngleMapper metric.

### To Reproduce

Just add the metric:

`self.sam = torchmetrics.image.SpectralAngleMapper()`

in your model `lightning.LightningModule` and use it in `training_step`, `validation_step`, or `test_step` functions.

The exact warning I obtain in console is this:

`/home/myusername/anaconda3/envs/myenv/lib/python3.10/site-packages/torchmetrics/utilities/prints.py:43: UserWarning: Metric `SpectralAngleMapper` will save all targets and predictions in the buffer. For large datasets, this may lead to a large memory footprint.`

### Environment

- TorchMetrics 1.1.0
- Python 3.10.12
- torch 2.0.1+cu118

Viola-Siemens:
## What does this PR do?

Fixes #2078 in lpips

The origin code stores all the compute graph when computing lpips, which usually runs out of memory. After detaching, the full compute graph will be garbage-collected and save the memory.

Check File Changes for detail.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
:books: Documentation preview :books:: https://torchmetrics--2093.org.readthedocs.build/en/2093/


Borda (approved):

SkafteNicki (approved):

src/torchmetrics/image/lpip.py
```diff
@@ -137,7 +137,7 @@ def __init__(
     def update(self, img1: Tensor, img2: Tensor) -> None:
         """Update internal states with lpips score."""
         loss, total = _lpips_update(img1, img2, net=self.net, normalize=self.normalize)
-        self.sum_scores += loss.sum()
+        self.sum_scores += loss.sum().detach()
```
stancld, line 140: Btw, we say the metric is differentiable. If we detach the tensor, we cannot call `backward` on that tensor. Is that something what we want?
cc: @Borda @SkafteNicki

Viola-Siemens: Maybe a detached version is needed if users do not care about the backward? If so, add a parameter of update to determine detach or not.

SkafteNicki (changes requested):
@stancld you are right, I was too quick to approve. The tensors are already detached. If we run
```python
from torchmetrics.image import LearnedPerceptualImagePatchSimilarity
import torch

m = LearnedPerceptualImagePatchSimilarity()

img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
img2 = (torch.rand(10, 3, 100, 100) * 2) - 1

m.update(img1, img2)
print(m.sum_scores.requires_grad)

m(img1, img2)
print(m.sum_scores.requires_grad)
```
it returns `False` and `False` in both cases.
@Viola-Siemens it must be something else in your code that is creating this problem.

src/torchmetrics/image/lpip.py
```diff
@@ -134,10 +134,13 @@ def __init__(
...
 
-    def update(self, img1: Tensor, img2: Tensor) -> None:
+    def update(self, img1: Tensor, img2: Tensor, detach: bool = False) -> None:
         """Update internal states with lpips score."""
         loss, total = _lpips_update(img1, img2, net=self.net, normalize=self.normalize)
-        self.sum_scores += loss.sum()
+        if detach:
+            self.sum_scores += loss.sum().detach()
+        else:
+            self.sum_scores += loss.sum()
```
Borda, line 143:
```suggestion
        self.sum_scores += loss.sum().detach() if detach else loss.sum()
```

lucadiliello (approved):

src/torchmetrics/image/lpip.py
```diff
@@ -134,10 +134,10 @@ def __init__(
         self.add_state("sum_scores", torch.tensor(0.0), dist_reduce_fx="sum")
         self.add_state("total", torch.tensor(0.0), dist_reduce_fx="sum")
 
-    def update(self, img1: Tensor, img2: Tensor) -> None:
+    def update(self, img1: Tensor, img2: Tensor, detach: bool = False) -> None:
```
Borda, line 137: also add docstring as this argument may not be intuistive

stancld (approved):

Viola-Siemens:
> @stancld you are right, I was too quick to approve. The tensors are already detached. If we run
> 
> ```python
> from torchmetrics.image import LearnedPerceptualImagePatchSimilarity
> import torch
> 
> m = LearnedPerceptualImagePatchSimilarity()
> 
> img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
> img2 = (torch.rand(10, 3, 100, 100) * 2) - 1
> 
> m.update(img1, img2)
> print(m.sum_scores.requires_grad)
> 
> m(img1, img2)
> print(m.sum_scores.requires_grad)
> ```
> 
> it returns `False` and `False` in both cases. @Viola-Siemens it must be something else in your code that is creating this problem.

Fixed. Now add a parameter for indicating detach or not.

SkafteNicki:
@Viola-Siemens could you please give an example of the issue. As shown in my example the tensors should already have been detached, and this additional detach operation should not be needed.
Another simple example does also show constant memory usage by the metric:
```python
from torchmetrics.image import LearnedPerceptualImagePatchSimilarity
import torch

# using forward method
lpips = LearnedPerceptualImagePatchSimilarity().cuda()
for _ in range(100):
    img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
    img2 = (torch.rand(10, 3, 100, 100) * 2) - 1
    val = lpips(img1.cuda(), img2.cuda())
    print(torch.cuda.memory_allocated())

print("===")

# using update method
lpips = LearnedPerceptualImagePatchSimilarity().cuda()
for _ in range(100):
    img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
    img2 = (torch.rand(10, 3, 100, 100) * 2) - 1
    lpips.update(img1.cuda(), img2.cuda())
    print(torch.cuda.memory_allocated())
```

SkafteNicki (changes requested):
Going to block until we have figured out why the detach operation are needed. We are already running the update method in `no_grad` mode:
https://github.com/Lightning-AI/torchmetrics/blob/0d7d6c7ccfedbb4941dfc55e65bbfca88506b2dc/src/torchmetrics/metric.py#L455

Viola-Siemens:
> @Viola-Siemens could you please give an example of the issue. As shown in my example the tensors should already have been detached, and this additional detach operation should not be needed. Another simple example does also show constant memory usage by the metric:
> 
> ```python
> from torchmetrics.image import LearnedPerceptualImagePatchSimilarity
> import torch
> 
> # using forward method
> lpips = LearnedPerceptualImagePatchSimilarity().cuda()
> for _ in range(100):
>     img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
>     img2 = (torch.rand(10, 3, 100, 100) * 2) - 1
>     val = lpips(img1.cuda(), img2.cuda())
>     print(torch.cuda.memory_allocated())
> 
> print("===")
> 
> # using update method
> lpips = LearnedPerceptualImagePatchSimilarity().cuda()
> for _ in range(100):
>     img1 = (torch.rand(10, 3, 100, 100) * 2) - 1
>     img2 = (torch.rand(10, 3, 100, 100) * 2) - 1
>     lpips.update(img1.cuda(), img2.cuda())
>     print(torch.cuda.memory_allocated())
> ```

I was running lpips on Cityscapes dataset with 2975 images(2048×1024). Try this:
```
lpips = LearnedPerceptualImagePatchSimilarity().cuda()
for _ in range(2975):
  img1 = (torch.rand(1, 3, 2048, 1024) * 2) - 1
  img2 = (torch.rand(1, 3, 2048, 1024) * 2) - 1
  lpips.update(img1.cuda(), img2.cuda())
  print(torch.cuda.memory_allocated())
```

But using detach method, it works and no error occurs.

SkafteNicki:
I am still getting that the memory consumption is the exact same for every step regardless of the image sizes increasing and the number of steps increasing

Viola-Siemens:
> @Viola-Siemens how is it going here? 🐰

I finally realized that `with torch.no_grad()` can solve this problem, maybe I should close this issue.

PR #2129, opened 2023-10-02
ci: proper setting oldest/latest for integrations & `PL>=1.9`

Borda:
## What does this PR do?

Integrations on GPU were strangely set always oldest req.

Dropped compatibility with:
- `1.6` -  no special reason, but the next two versions wont work so skip also this old one
- `1.7` - Pl was looking for removed protected function, `ERROR integrations/test_lightning.py - ImportError: cannot import name '_compare_version' from 'torchmetrics.utilities.imports'`
- `1.8` - not aligned dependencies:
	```
	pytorch-lightning 1.8.0 depends on lightning-utilities==0.3.*
    torchmetrics 1.3.0.dev0 depends on lightning-utilities<0.10.0 and >=0.8.0
	```

<details>
  <summary>Before submitting</summary>

- [x] Was this **discussed/agreed** via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/torchmetrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to **update the docs**?
- [x] Did you write any new **necessary tests**?

</details>

<details>
  <summary>PR review</summary>

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

</details>

## Did you have fun?

Make sure you had fun coding 🙃

----
:books: Documentation preview :books:: https://torchmetrics--2129.org.readthedocs.build/en/2129/

PR #2141, opened 2023-10-04
deprecate `_update_called`

matsumotosan:
## What does this PR do?

Removes deprecated `Metric._update_called` in favor of `Metric.update_called`.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
:books: Documentation preview :books:: https://torchmetrics--2141.org.readthedocs.build/en/2141/


SkafteNicki:
Here are the lines that should be updated on lightning side:
https://github.com/Lightning-AI/lightning/blob/09a0fb26d2bb819c6cbabf87c695b657ac5d305f/src/lightning/pytorch/trainer/connectors/logger_connector/result.py#L268-L273

awaelchli:
Thank you for taking care of this. With #18714 Lightning 2.1 will call the public method. Sorry that it was calling the protected API 😅 

What is torchmetrics' policy around supporting different lightning versions? Could one use a newer torchmetrics version with an older lightning version? If yes, probably the change here would have to wait otherwise it would break. Anyway, great work on sorting it out for the future! Thanks

SkafteNicki:
> What is torchmetrics' policy around supporting different lightning versions? Could one use a newer torchmetrics version with an older lightning version? If yes, probably the change here would have to wait otherwise it would break. Anyway, great work on sorting it out for the future! Thanks

@awaelchli I do not think we have an official policy regarding lightning, only torch where we aim for new releases of torchmetrics to be compatible with up to 2 year old releases of torch.
@Borda opinion on what we should do here?

Borda:
> I do not think we have an official policy regarding lightning

As we are the same group, we try to support as large as we can without too much overhead, but said so, we do not need to be more generous than lightning itself meaning we are actively maintaining 1.9.x and 2.0+ even here we have set range:
https://github.com/Lightning-AI/torchmetrics/blob/d8962b4fcaad100ea4370430240a5879ef0e2826/requirements/integrate.txt#L1
also addressed in #2129 

So, to be clear, this would be a breaking change such that users with TM 1.3+ can't use lighting bellow 2.1, correct?
cc: @lantiga

src/torchmetrics/metric.py
```diff
@@ -168,8 +168,8 @@ def __init__(
 
     @property
     def _update_called(self) -> bool:
-        # TODO: this is needed for internal lightning, remove after v0.12 and update on lightning side
-        return self._update_count > 0
+        # NOTE: this is needed for internal lightning; remove after v0.12 and update on lightning side
+        return self.update_called()
```
Borda, line 172: let's add a deprecation warning to be removed in 2.0?

Borda (approved):

justusschock (approved):

PR #2198, opened 2023-11-02
Add `zero_division` option to the precision, recall, f1, fbeta.

i-aki-y:
## What does this PR do?

I want to add zero_division option to the precision, recall, f1, fbeta metrics as well as the sklearn counterparts.
cf, https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html#sklearn-metrics-precision-score

The zero_division is important when we use samplewise metrics (`multidim_average="samplewise"`) where some samples have no positive targets.

The following example shows that the `preds1` and `preds2` have the same f1-scores (0.3333). 
But the `pred2` perfectly matches the target while the `pred1` does not.
This means that we could not distinguish the two models: The model can correctly predict no positive sample from the model that returns many false positives.
We can fix this by setting the zero_division=1.
For this example, the `preds2` would become (f1=1.0) while the `preds1` is still (f1=0.3333).

```py
import torch
from torchmetrics.functional.classification import f1_score

targets = torch.tensor([
    [0, 0, 0, 0],  # sample1
    [0, 0, 0, 0],  # sample2
    [0, 0, 1, 1],  # sample3
])

preds1 = torch.tensor([
    [0, 1, 1, 1],
    [1, 0, 1, 1],
    [0, 0, 1, 1],
])

preds2 = torch.tensor([
    [0, 0, 0, 0],
    [0, 0, 0, 0],
    [0, 0, 1, 1],
])

## default behavior
scores1 = f1_score(preds1, targets, task="binary", multidim_average="samplewise")
print(scores1, scores1.mean())
#=> tensor([0., 0., 1.]) tensor(0.3333)

scores2 = f1_score(preds2, targets, task="binary", multidim_average="samplewise")
print(scores2, scores2.mean())
#=> tensor([0., 0., 1.]) tensor(0.3333)

## If zero_division = 1
scores1 = f1_score(preds1, targets, task="binary", multidim_average="samplewise", zero_division=1)
print(scores1, scores1.mean())
#=> tensor([0., 0., 1.]) tensor(0.3333)

scores2 = f1_score(preds2, targets, task="binary", multidim_average="samplewise", zero_division=1)
print(scores2, scores2.mean())
#=> tensor([1., 1., 1.]) tensor(1.)
```

Note:
The latest sklearn (ver 1.3) has a bug in f1_score when zero_division=1.
https://github.com/scikit-learn/scikit-learn/pull/27577

So, some test cases that compare the results with the sklearn will fail until the bug is fixed.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
:books: Documentation preview :books:: https://torchmetrics--2198.org.readthedocs.build/en/2198/


SkafteNicki:
Hi @i-aki-y,
I am okay with this change but would I think it would make sense base `StatScores` class and then use Precision, Recall, FScore etc.

i-aki-y:
@SkafteNicki  Thank you for your comment.

I tried refactoring StatScores to have zero_division.
And removed `__init__` and `*_precision_recall_score_arg_validation` methods, which I added, from the Precision and Recall.

I avoid declaring `zero_division` argument in StatScores' constructor because StatScores (and some other sub-classes) does not use zero_division.
So, the `zero_division` argument is passed through the kwargs and is poped before the `supre().__init__(**kwargs)` code.

ex.
```py
class BinaryStatScores(_AbstractStatScores):
        ...
    def __init__(
        self,
        threshold: float = 0.5,
        multidim_average: Literal["global", "samplewise"] = "global",
        ignore_index: Optional[int] = None,
        validate_args: bool = True,
        **kwargs: Any,
    ) -> None:
        zero_division = kwargs.pop("zero_division", 0)
        super(_AbstractStatScores, self).__init__(**kwargs)
        if validate_args:
            _binary_stat_scores_arg_validation(threshold, multidim_average, ignore_index, zero_division)

```

Borda:
@i-aki-y seems that the results are different then expected...

i-aki-y:
@Borda Thanks

I found some bugs and fixed them. 

Since the PR (https://github.com/scikit-learn/scikit-learn/pull/27577) was merged, I confirmed that the fix passes related test cases (ex. `pytest tests/unittests/classification`) using the dev version of sklearn==1.4.dev0 and pytorch==2.0.1 in my local machine.

Borda:
@i-aki-y could you pls have a look at all the failing cases, some with the wrong value...
turning it to draft till the tests are resolved, pls make it ready when test are mostly green :)

i-aki-y:
@Borda  OK, I put the [WIP] in this PR title.

As mentioned above, the current sklearn (1.3.2) has a bug that mishandles zero_division.
I think it will cause the CI errors.
Fortunately, the bugfix has been merged recently, so I expect the next sklearn's version (1.3.3 or 1.4?) will fix these problems.

Borda:
> Fortunately, the bugfix has been merged recently, so I expect the next sklearn's version (1.3.3 or 1.4?) will fix these problems.

Nice, 1.4 is out so let's finish it =)

i-aki-y:
@Borda  Thank you for your support.

Now I removed [WIP], since most checks have passed.

i-aki-y:
It appears that the problem of sklearn's bug is only in the sklearn == 1.3 and does not occur in 1.2, 1.4.
So, the workaround that separates requirements versions should solve the issue.

```
scikit-learn >=1.1.1, <1.3.0; python_version < "3.9"
scikit-learn >=1.4.0, <1.5.0; python_version >= "3.9"
```


The errors that are still observed in the 3.8 unit tests seem to be a different problem.

Actually, I found that there are discrepancies between sklearn and torchmetrics in multiclass scores when the average="weighted" and empty inputs are set.

ex.

```py
from torch import tensor
from torchmetrics.functional.classification import multiclass_f1_score
from sklearn.metrics import f1_score as sk_f1_score
target = tensor([]).reshape(1, 0)
preds = tensor([]).reshape(1, 0)

# torchmetrics
print(multiclass_f1_score(preds, target, num_classes=5, average="weighted", zero_division=1,).item())
# ==> 0
# sklearn
print(sk_f1_score(target.numpy().flatten(), preds.numpy().flatten(), labels=[0, 1, 2, 3, 4], average="weighted", zero_division=1,))
# ==> 1

```

Note that empty inputs can happen when we  use ignore_index as follows:

```py
target = tensor([0, 0, 0]).reshape(1, 0)
preds = tensor([4, 0, 1]).reshape(1, 0)
ignore_index = 0

# The target, press are eventually empty.
multiclass_f1_score(preds, target, num_classes=5, ignore_index=ignore_index, average="weighted", zero_division=1)
```

I am investigating why this happens.

requirements/_tests.txt
```diff
@@ -15,4 +15,5 @@ requests <=2.31.0
 fire <=0.5.0
 
 cloudpickle >1.3, <=3.0.0
-scikit-learn >=1.1.1, <1.4.0
+scikit-learn >=1.1.1, <1.3.0; python_version < "3.9"
+scikit-learn >=1.4.0, <1.5.0; python_version >= "3.9"
```
mathematicalmichael, line 19: shouldn't this be
```suggestion
scikit-learn >=1.3.0, <1.5.0; python_version >= "3.9"
```
otherwise versions in the interval `[1.3.0, 1.4.0)` are omitted.
assuming 1.3.0 breaks python 3.8 functionality. 

if 1.4.0 breaks it, then the change would be

```
scikit-learn >=1.1.1, <1.4.0; python_version < "3.9"
scikit-learn >=1.4.0, <1.5.0; python_version >= "3.9"
```

mathematicalmichael, line 19: if 1.4.0 breaks it, then the change would be
```suggestion
scikit-learn >=1.1.1, <1.4.0; python_version < "3.9"
scikit-learn >=1.4.0, <1.5.0; python_version >= "3.9"
```

i-aki-y: @mathematicalmichael  Thank you for your comment.

I intentionally skipped version 1.3 because the sklearn 1.3 has a bug breaking `zero_division` test.
As of Python 3.8, this is a summary:
- sklearn 1.4 does not support py3.8 --> NG
- sklearn 1.3 support py3.8 but has a bug breaking tests --> NG
- sklearn 1.2 support py3.8 and does not have the bug --> OK
I think sklearn 1.2 is the latest version that can pass the tests in this PR.

i-aki-y:
Now, I have modified the test code to avoid the errors that occur in some special cases.
The problem is caused by the inconsistency of some skclearn's function results for multiple versions.
For example, the result of f1_score([], [], labels=None, average="weighted", zero_division=zero_division) 
varies depending on the sklearn version:

- 1.2 -> the value of zero_division
- 1.3 -> nan
- 1.4 -> nan
(The `nan` value is replaced with 0.0 in the test code later.)

torchmetrics functions look to expect the fixed value 0.0 for this condition independent of the sklearn's version and zero_division values.
So I hard-coded the result of sklearn's function to be 0.0 when the above condition (average=weighted and empty inputs) was satisfied.

requirements/_tests.txt
i-aki-y: @mathematicalmichael Sorry for the confusion. I want to follow up on my explanation.

There are two problems now:

1. sklearn 1.3.x has a bug related to zero_division that will break this PR test.
2. sklearn >= 1.4.x that fixed the bug does not support python <= 3.8

Your suggestion `scikit-learn >=1.1.1, <1.4.0; python_version < "3.9"` would break CI test for python3.8  because the condition allows installing sklearn==1.3.x that has the zero_division bug.

Fortunately, the latest sklearn==1.4 fixed the bug and passed all other torchmetrics test cases. 

If I understand correctly, the original condition `scikit-learn >=1.1.1, <1.4.0` means CI should pass one of the versions from 1.1.1 to 1.3.x and does not mean that CI passes all of them. 

So, I think we can update the condition to `scikit-learn >=1.4.0, <1.5.0` for the `python_version >= 3.9` if there is no dependency problem.

As for `python <= 3.8`, I had to set `scikit-learn < 1.3`, but it is a subset of the original condition and should not conflict with it.

lantiga:
@SkafteNicki let's revive this

i-aki-y:
@robmarkcole  The jaccard index seems to be implemented by using `_safe_devide` in the `_jaccard_index_reduce`, so I think a similar fix is possible.

requirements/_tests.txt
Borda: The same as https://github.com/Lightning-AI/torchmetrics/pull/2198/files#r1508408419

SkafteNicki:
@robmarkcole and @i-aki-y I added support in jaccard index now.
I make sure to get the remaining test passing so we can finally land this PR either today or tomorrow and then have a release afterwards.

SkafteNicki:
The PR has been failing for the `sensitivity_at_specificity` metric for python 3.8 for some time now. Reason being that a bugfix only present in `scikit-learn>=1.3.0` for their implementation was needed to match our implementation but this PR makes the required version of `scikit-learn<1.3` for python 3.8, so that was the reason those test were failing.
They are now being skipped for old sklearn versions because that is the only solution where we can land this PR.

SkafteNicki (approved):

Borda (approved):

PR #2335, opened 2024-01-30
ci: cache reference metrics [wip]

### LINKED ISSUE #949: Tests: Prepare reference scores once for all env. configs
## 🚀 Feature

If we use the reference metric only for generating static results, these results will be loaded and used in all testing environments?

The proposed flow is semi-online; still, the references are generated during CI test time but only once and shared to remaining tests... so consider using some JSON or YAML universal format.

_TIP1: The references can be shared among jobs as artifact:rabbit:_ 

_TIP2: To speed up we can use some caching with limited time as we do for pip/package caching_

For example, the case with all Condas:
1. create all testing references with frozen env that is supported by all reference metrics
2. run full matric of all configurations PT v{1.4, ...1.12}

### Motivation

this may help to unblock various configurations and also:
- guarantee that the results are identical regardless of env. and if needed, they can be inspected...
- testing could be faster as references are not computed for each config...
- we have not locked with testing for env that the reference packages support...

### Pitch

Make tests more predictable and scalable.

### Alternatives

Still generating on-the-fly results, BUT let's discuss this option a bit... :)

### Additional context

part of #678
pls comment @PyTorchLightning/core-metrics

Borda:
## What does this PR do?

Resolves #949

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?


Borda:
it seems tha hashing is not supported by its nature for `list` or `np.ndarray`, ref: https://github.com/python-cachier/cachier/issues/136

shaypal5:
> it seems that hashing is not supported by its nature for `list` or `np.ndarray`, see:
> 
> * [Numpy array default hasher python-cachier/cachier#43](https://github.com/python-cachier/cachier/issues/43)
> * [workaround for `TypeError: unhashable type: 'list'` python-cachier/cachier#136](https://github.com/python-cachier/cachier/issues/136)

@Borda, please note you can easily work around this using custom hashing logic, which `cachier` supports:
https://github.com/python-cachier/cachier?tab=readme-ov-file#working-with-unhashable-arguments

Borda:
> note you can easily work around this using custom hashing logic, which `cachier` supports

yes, I saw it but this I believe is quite often the case so it would be beneficial to cover it out of the box (I mean `list` not `numpy.andarray`)

shaypal5:
> > note you can easily work around this using custom hashing logic, which `cachier` supports
> 
> @shaypal5, yes, I saw it but this I believe is quite often the case so it would be beneficial to cover it out of the box (I mean `list` not `numpy.andarray`)

Cool. As long as you are aware this shouldn't block your use case, I'd love to get contributions by you into `cachier`!

Borda:
An idea bout increasing the testing range, we could move importing the dependency package to the `def _reference_XXX(...)` function with eventual skip if not imported... that would mean once the reference computed on with feasible configuration (mostly means package and python version compatibility issue) it will be used as part of a cache for remaining configurations.
This would fully cover the suggested improvement mentioned in #949 with minimal engineering overhead :rocket: 

cc: @SkafteNicki @justusschock @lantiga

SkafteNicki (approved):

PR #2358, opened 2024-02-06
Improve confusion matrix plotting

JonasVerbickas:
## What does this PR do?

Current confusion matrix plotting tends to produce unusable results.
When I tried to create a normalized multilabel confusion matrix plot this cluttered mess was produced:
image

1. When normalizing converting tensor to float using `val.item()` produces numbers with too many decimal places.
2. Redundant "True class" and "Predicted class" use up a lot of space and make the graph unreadable
3. Longer labels overlap

### Round floats to avoid floating point errors leading to UI overflow.

Rounding to two decimal places seems reasonable since it's difficult to fit more digits into multilabel confusion matrices.
```python
ax.text(jj, ii, str(round(val.item(), 2)), ha="center", va="center", fontsize=15)
```
> Rounding the `val` tensor itself does not work here. The issue is with the `item()` method converting incorrectly.

Changing this line converts this plot:
image

> It might be worth considering using less aggressive rounding for simpler binary confusion matrices since they have room to display more digits without any overlap.

### Removing redundant "True class" and "Predicted class"
Code reducing the number of times x and y labels are shown:
```python
if i // cols == rows-1:  # bottom row only
    ax.set_xlabel("Predicted class", fontsize=15)
if i % cols == 0:  # leftmost column only
    ax.set_ylabel("True class", fontsize=15)
```

Produces a much cleaner plot without sacrificing any information:
image

### Reduce overlap
By utilizing `constrained_layout=True`:
```python
fig, axs = plt.subplots(nrows=rows, ncols=cols, constrained_layout=True) if ax is None else (ax.get_figure(), ax)
```
Reduces overlap between labels horizontally and vertically from {0, 1} ticks:
image
> Even though this does not fix the top row, the longest names in this example are extreme. 
Ensuring proper middle-row separation is good enough for now.

<details>

- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/torchmetrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--2358.org.readthedocs.build/en/2358/


Borda:
yes, the readability for binary cases is bad, just thinking about making the long labels as multi-line or smaller font? :thinking:

JonasVerbickas:
If I have the time, I will look into multiline labels this week.
That said, I think this commit is a good first step towards confusion matrix plot clarity.

SkafteNicki (approved):

Borda (approved):

PR #2381, opened 2024-02-14
Add support for SQ & RQ as well as per-class metrics

### LINKED ISSUE #2380: Add Support for SQ and RQ in Panoptic Quality
## 🚀 Feature

First of all thanks for implementing the Panoptic Quality (PQ) metric. However, the current PQ implementation only computes the PQ. The Segmentation Quality (SQ) and Recognition Quality (RQ) are not computed. It would be amazing to have an option to also compute these metrics within the PQ implementation.

### Motivation

Almost all papers (_e.g._, [U2Seg](https://arxiv.org/pdf/2312.17243.pdf)) report the PQ alongside SQ and RQ so extending the PQ implementation to also compute SQ and RQ would be amazing.

### Alternatives

We could also implement both SQ and RQ as separate metrics, however, since I'm not aware of any work reporting SQ or RQ w/o PQ this would be somewhat inconvenient. This would also lead to an additional compute overhead.

ChristophReich1996:
## What does this PR do?

Fixes #2380

This pull request adds support for also computing the Segmentation and Recognition Quality alongside the Panoptic Quality. I also added a flag to return the per-class metrics. Additional tests probably need to be added, right?

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--2381.org.readthedocs.build/en/2381/


SkafteNicki:
@ChristophReich1996 thanks for sending the pull request!
Overall I think it looks good. I have a suggestions:
* Should we rename `per_class` to `return_per_class`?
* I would prefer if we try to keep one output type. So what I suggest is
  - If `return_sq_and_rq=True` return a tensor of length 3 e.g. `torch.tensor([pq, sq, rq])`
  - If `return_per_class=True` return a tensor of length classes but along the first dimension such that it is a [1, C] tensor
  - If both are `True` return a tensor with shape [3, C]

What do you think?

ChristophReich1996:
@SkafteNicki Yeah having the same return time is useful and renaming the `per_class` flag is as well. I updated the code. Let me know what you think!

Borda:
@SkafteNicki how is it going here, anything missing?

SkafteNicki (approved):

Borda (approved):

Issue #2481, opened 2024-03-29
Retrieval Metrics GPU Memory Leak


astirn:
## 🐛 Bug

Updating a `RetrievalMetric` in training and/or validation results in runaway GPU memory usage.

### To Reproduce

In a `LightningModule`'s `def __init__(...)`, declare

```py
metrics = tm.MetricCollection({
   'mrr': tm.RetrievalMRR()
})
self.train_retrieval_metrics = metrics.clone(prefix='train_')
self.val_retrieval_metrics = metrics.clone(prefix='val_')
```

Then in `training_step(self, batch, batch_idx)` and `validation_step(self, batch, batch_idx)`, respectively call

```py
self.train_retrieval_metrics(logits, targets, indexes)
```
and 

```py
self.val_retrieval_metrics(logits, targets, indexes)
```

where `logits` is a batch size x batch size matrix of logits. You can assume `targets` and `indexes` are correct since I am getting good MRR measurements.

I believe I tracked down the problem. In `retrieval/base.py`, the `RetrievalMetric` class initializes

```py
self.add_state("indexes", default=[], dist_reduce_fx=None)
self.add_state("preds", default=[], dist_reduce_fx=None)
self.add_state("target", default=[], dist_reduce_fx=None)
```

using python lists. Because these are lists, the `Metric` class's `reset(self)` in `metric.py`

```py
for attr, default in self._defaults.items():
    current_val = getattr(self, attr)
    if isinstance(default, Tensor):
        setattr(self, attr, default.detach().clone().to(current_val.device))
    else:
        setattr(self, attr, [])
```

the `setattr(self, attr, default.detach().clone().to(current_val.device))` is never reached.

### Expected behavior

GPU memory does not increase overtime.

### Environment

- TorchMetrics version: 1.3.2
- Python & PyTorch Version: 3.11.8 & 2.2.1

lucadiliello:
Hello @astirn. I think the behaviour is correct, as Retrieval Metrics are designed only to be globally valid, and to be computed on the whole dataset, thus they will accumulate results till the end of the epoch.

Please try the following in your `LightningModule`:
```python
def training_step(self, batch, batch_idx):
    ...
    self.train_retrieval_metrics.update(logits, targets, indexes)

def on_train_epoch_end(self):
    metrics = self.train_retrieval_metrics.compute()
```

[...]

PR #2493, opened 2024-04-06
Clear `list` states (i.e. delete their contents), not reassign the default `[]`

### LINKED ISSUE #2492: `list` states leak (`Tensor`) memory
## 🐛 Bug

Hello!

I've recently tracked a CPU (possibly GPU also) memory leak to `list[Tensor]` states. For example:
```
import torch
from torchmetrics import Metric

class DummyListMetric(Metric):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.add_state("x", default=[])

    def update(self, x=None):
        x = torch.tensor(1) if x is None else x
        self.x.append(x)
```

When (the parent) `Metric.reset()` is called, `self.x` is simply overwritten with an empty list [1]. Unfortunately, this doesn't guarantee that the contents of `self.x` are deleted, meaning `Tensor` elements are not always correctly freed

After some investigation, I found my custom metrics (subclasses of`Metric`) didn't cause my (work) system to run out of memory if I added the following overload:
```
def reset(self):
    for attr, default in self._defaults.items():
        if isinstance(default, list):
            getattr(self, attr).clear()

    return super().reset()
```

The same fix can be applied directly inside `torchmetrics` with a single line change, modifying [2] to:
```
getattr(self, attr).clear()
```

Looking at other open issues, this issue might be related to #2481 (which also references `list` states). I'd be very happy to open a PR if the above sounds reasonable. Many thanks!

- [1] https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/metric.py#L673
- [2] https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/metric.py#L684

### Environment

- TorchMetrics version: 1.3.2
- Python & PyTorch Version: 3.11.8 & 2.2.1

dominicgkerr:
Rather than overwriting `list` states (with `[]`), call `.clear()` to correctly (more robustly?) delete/free `Tensor` elements. Previous behaviour results in CPU (possibly GPU also) memory leak

## What does this PR do?

Fixes #2492 (maybe #2481 also)

<details>

- [X] Was this **discussed** via a Github issue? (no need for typos and docs improvements)
- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/torchmetrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] ~Did you make sure to **update the docs**?~

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--2493.org.readthedocs.build/en/2493/

### DIFF
--- a/.github/workflows/publish-pkg.yml
+++ b/.github/workflows/publish-pkg.yml
@@ -67,7 +67,7 @@ jobs:
       - run: ls -lh dist/
       # We do this, since failures on test.pypi aren't that bad
       - name: Publish to Test PyPI
-        uses: pypa/gh-action-pypi-publish@v1.8.12
+        uses: pypa/gh-action-pypi-publish@v1.8.14
         with:
           user: __token__
           password: ${{ secrets.test_pypi_password }}
@@ -94,7 +94,7 @@ jobs:
           path: dist
       - run: ls -lh dist/
       - name: Publish distribution 📦 to PyPI
-        uses: pypa/gh-action-pypi-publish@v1.8.12
+        uses: pypa/gh-action-pypi-publish@v1.8.14
         with:
           user: __token__
           password: ${{ secrets.pypi_password }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -83,6 +83,7 @@ repos:
     rev: v3.1.0
     hooks:
       - id: prettier
+        files: \.(json|yml|yaml|toml)
         # https://prettier.io/docs/en/options.html#print-width
         args: ["--print-width=120"]
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ ci:
 
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v4.6.0
     hooks:
       - id: end-of-file-fixer
       - id: trailing-whitespace
@@ -46,7 +46,7 @@ repos:
         exclude: pyproject.toml
 
   - repo: https://github.com/crate-ci/typos
-    rev: v1.16.26
+    rev: v1.20.7
     hooks:
       - id: typos
         # empty to do not write fixes
@@ -112,7 +112,7 @@ repos:
       - id: text-unicode-replacement-char
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.2
+    rev: v0.3.5
     hooks:
       # try to fix what is possible
       - id: ruff
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,0 @@
-# Ignore all MD files:
-**/*.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed memory leak in metrics using list states ([#2492](https://github.com/Lightning-AI/torchmetrics/pull/2492))
 
 
+- Fixed bug in computation of `ERGAS` metric ([#2498](https://github.com/Lightning-AI/torchmetrics/pull/2498))
+
+
 - Fixed `BootStrapper` wrapper not working with `kwargs` provided argument ([#2503](https://github.com/Lightning-AI/torchmetrics/pull/2503))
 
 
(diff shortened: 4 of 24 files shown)


Borda (approved):
nice, can we also have a test to cover this edge-case?

dominicgkerr:
> nice, can we also have a test to cover this edge-case?

Absolutely! Added 1fa7077 to help illustrate the issue (and how the fix helps)

justusschock (commented):
Thanks for your PR and that's certainly a great find. 
That being said, I am actually not sure what's the desired behavior here.

I can see good reasons for both ways.

I.e. It might be very confusing for people if they keep a reference to things why this one suddenly empty, whereas with just instantiating a new list we're doing everything correctly on our side and still give the user the possibility to retain a reference to the old state if they wish to. So this would only lead to a memory leak in user code, not on our side.

dominicgkerr:
Interesting - I hadn't thought of that scenario being desirable...

I would argue though, the exiting behaviour is certainly unexpected (according to the current documentation). My reading of `.add_state` is that users (intentionally) defer the memory/device management entirely to `Metric`, and wouldn't use it if they needed something more complex (like persistent references.) Similarly, I naturally expect `.reset` to return the `Metric` to its original states, AND to tidy-up any intermediate values calculated by `.update` (which are currently leaked.) 

In my specific use-case, I'm fitting a `LighntningModule` to a very large dataset. I have a setup very similar to [1], except I'm using a custom `Metric` with a `list` state - over the course of a large number of training steps, I consistently see my CPU memory usage increase (as a result of #2492) until the python process crashes.

If you're keen to support the ability to retain references to `Metric` states, would introducing a (say) `free_on_reset: bool` keyword argument to `.add_state` (that conditionally deletes/clears states before reassigning the default value) be preferable? Setting `free_on_reset=True` by could enable the new behaviour by default, while still allowing users to revert to the existing behaviour whn they require something more complex?

- [1] https://lightning.ai/docs/torchmetrics/stable/pages/lightning.html#logging-torchmetrics

justusschock:
I think you have some valid arguments there. Could you just add that to the documentation and refer people to use copy/deepcopy if the want to retain the states?

dominicgkerr:
Brilliant, thanks! I've updated the docstring/documentation - let me know if it needs rewording/expanding/etc... 

Unfortunately I couldn't get `make docs` to build locally (as I couldn't `pip install lai-sphinx-theme`), so just waiting on the CI to double check my formatting is correct

Thanks again

Borda:
@SkafteNicki thoughts?

justusschock (approved):

SkafteNicki (approved):

Borda:
@dominicgkerr there are too many failing tests, could you pls have look...

dominicgkerr:
@SkafteNicki Nice (5758977) - was just looking at the `.forward` logic!

SkafteNicki:
@dominicgkerr you are welcome :)
Hopefully all tests should pass now

dominicgkerr:
I think I've got things working (better?) now... @SkafteNicki / [5758977](https://github.com/Lightning-AI/torchmetrics/pull/2493/commits/57589773c28cebd6de20ec5bc76a3c4a0099b1c0) was a huge help (thanks!), as I hadn't spotted the caching behaviour inside `.forward` / `.update` before.

Rather than simply using `deepcopy` (which worked in all but one edge case), I added a `._copy_state_dict` helper method to `.detach().clone()` `Tensor` / `list[Tensor]` states inserted into the cache. Unfortunately this caused `tests/unittests/bases/test_metric.py:test_constant_memory_on_repeat_init` to fail, but I found (in [ef27215](https://github.com/Lightning-AI/torchmetrics/pull/2493/commits/ef2721575cfe5993ceba94c455903f3bf1356831)) that removing the `.clone()` calls, fixed it. But, as the resulting cache might now contain references to `Tensor` values, I was suspicious that it somehow leak memory...

After some digging, I came to the conclusion that `test_constant_memory_on_repeat_init` wasn't actually testing memory leakage during initialisation, but actually ensure repeated `.forward` calls didn't (significantly) increase the metric's memory. As an illustration, moving `x = torch.randn(10000).cuda()` inside the test's `for` loop (where a constant `Tensor` couldn't be automatically referenced by python/pytorch in subsequent calls), actually caused it to fail...

I claim, memory should be allowed to increase here (as users might legitimately want to collect lots of observations during fitting, and combine them inside `.compute`), as long as the allocated memory is eventually freed by `.reset`. I've added a new test (`test_freed_memory_on_reset`) check this. 

I appreciate changing (failing) tests to pass the CI is pretty suspect - hopefully the above explains why I did (let me know if not!)

stancld (approved):

janEbert:
Should the same be applied here as well?
https://github.com/Lightning-AI/torchmetrics/blob/d528131c8e7f130c65ba62a14e12f15906d488c1/src/torchmetrics/metric.py#L530

PR #2571, opened 2024-05-28
Fix incorrect caching of MetricCollection

### LINKED ISSUE #2570: Incorrect caching (`m._compute`) of metrics inside a `MetricCollection` if `compute_groups` are used and `.compute` is called twice
## 🐛 Bug

Hello Team! I have found a bug which I already tracked down to is related to the caching attribute `_compute` of a metric in a metric collection. If the `.compute()` method is called twice then the second call returns different (and false) results if compute_groups are used. That is because the first `.compute()` call will store result of the first metric in the `_compute` cache for all other metrics of that compute group, which is then returned in the second call.

### To Reproduce

<details>

```py
import torch
from torchmetrics import AUROC, ROC, Recall, F1Score, MetricCollection

metrics = MetricCollection({
    "auroc": AUROC(task='binary'),
    "roc": ROC(task='binary'),
    "recall": Recall(task='binary'),
    "f1": F1Score(task='binary')
})

y_true = torch.tensor([1, 0, 0, 1])
y_pred = torch.tensor([0.6, 0.2, 0.4, 0.2])

for batch in range(10):
    metrics.update(y_pred, y_true)

print("First compute call:")
print(metrics.compute())
print("Second (and all consecutive) compute call:")
print(metrics.compute())
print("Compute Groups:")
print(metrics.compute_groups)
```

This results in:

```sh
First compute call:
{
    'auroc': tensor(0.6250),
    'f1': tensor(0.6667),
    'recall': tensor(0.5000),
    'roc': (
        tensor([0.0000, 0.0000, 0.5000, 1.0000]),
        tensor([0.0000, 0.5000, 0.5000, 1.0000]),
        tensor([1.0000, 0.6000, 0.4000, 0.2000])
    )
}
Second (and all consecutive) compute call:
{'auroc': tensor(0.6250), 'f1': tensor(0.6667), 'recall': tensor(0.6667), 'roc': tensor(0.6250)}
Compute Groups:
{0: ['auroc', 'roc'], 1: ['f1', 'recall']}
```

</details>

### Expected behavior

<details>
  <summary>I expect the output of above code to be</summary>

```sh
First compute call:
{
    'auroc': tensor(0.6250),
    'f1': tensor(0.6667),
    'recall': tensor(0.5000),
    'roc': (
        tensor([0.0000, 0.0000, 0.5000, 1.0000]),
        tensor([0.0000, 0.5000, 0.5000, 1.0000]),
        tensor([1.0000, 0.6000, 0.4000, 0.2000])
    )
}
Second (and all consecutive) compute call:
{
    'auroc': tensor(0.6250),
    'f1': tensor(0.6667),
    'recall': tensor(0.5000),
    'roc': (
        tensor([0.0000, 0.0000, 0.5000, 1.0000]),
        tensor([0.0000, 0.5000, 0.5000, 1.0000]),
        tensor([1.0000, 0.6000, 0.4000, 0.2000])
    )
}
Compute Groups:
{0: ['auroc', 'roc'], 1: ['f1', 'recall']}
```
</details>

### Environment

- TorchMetrics version (and how you installed TM, e.g. `conda`, `pip`, build from source):  1.4.0.post0 installed via `rye add`
- Python & PyTorch Version (e.g., 1.0): Python 3.10.14 & PyTorch 2.2.0
- Any other relevant information such as OS (e.g., Linux): Ubuntu 22 on WSL

### Additional context

I already tracked down the respective row:

https://github.com/Lightning-AI/torchmetrics/blob/8d9d2ae5f6fea65e861026b90e365ec34b5c57b1/src/torchmetrics/collections.py#L307

It was introduced by the PR #2211 . Since they also fixed something with that, simply reverting the change would not help. I currently try to find an other solution for this.

_But Tobi, why do you want to have two compute calls in the first place?_

The reason why I want to use two seperatly `.compute()` calls is that in my current code structure I want to log my metrics in two different functions which have access to the same metrics object. The use case is actually pretty close to the example I provided above. Of course I could use a workaround by just storing the result by myself and handle it in my respective functions, but this would make my code less readable. Therefore a fix for this Issue would be a nice to have for me. **However, other people could run into the same problem without noticing, since for some e.g. scalar metrics the code would run as usual, just with the wrong outputs.**

relativityhd:
## What does this PR do?

Fixes  #2570. Also provides a better test for #2211. A somewhat larger change is that for this fix (especially the testcases) the compute method needs to return deepcopied values.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--2571.org.readthedocs.build/en/2571/

### DIFF
--- a/src/torchmetrics/metric.py
+++ b/src/torchmetrics/metric.py
@@ -619,7 +619,7 @@ class Metric(Module, ABC):
 
             # return cached value
             if self._computed is not None:
-                return self._computed
+                return deepcopy(self._computed)
 
             # compute relies on the sync context manager to gather the states across processes and apply reduction
             # if synchronization happened, the current rank accumulated states will be restored to keep
@@ -634,7 +634,8 @@ class Metric(Module, ABC):
             if self.compute_with_cache:
                 self._computed = value
 
-            return value
+            # Return a deep copy to avoid side effects for non-scalar values, e.g. ConfusionMatrix
+            return deepcopy(value)
 
         return wrapped_func
 
--- a/src/torchmetrics/metric.py
+++ b/src/torchmetrics/metric.py
@@ -619,7 +619,7 @@ class Metric(Module, ABC):
 
             # return cached value
             if self._computed is not None:
-                return deepcopy(self._computed)
+                return self._computed
 
             # compute relies on the sync context manager to gather the states across processes and apply reduction
             # if synchronization happened, the current rank accumulated states will be restored to keep
@@ -630,12 +630,13 @@ class Metric(Module, ABC):
                 should_unsync=self._should_unsync,
             ):
                 value = _squeeze_if_scalar(compute(*args, **kwargs))
+                # clone tensor to avoid in-place operations after compute, altering already computed results
+                value = apply_to_collection(value, Tensor, lambda x: x.clone())
 
             if self.compute_with_cache:
                 self._computed = value
 
-            # Return a deep copy to avoid side effects for non-scalar values, e.g. ConfusionMatrix
-            return deepcopy(value)
+            return value
 
         return wrapped_func
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
--
+- Fixed bug in `MetricCollection` when using compute groups and `compute` is called more than once ([#2571](https://github.com/Lightning-AI/torchmetrics/pull/2571))
 
 
 
--- a/src/torchmetrics/collections.py
+++ b/src/torchmetrics/collections.py
@@ -187,6 +187,11 @@ class MetricCollection(ModuleDict):
 
         self.add_metrics(metrics, *additional_metrics)
 
+    @property
+    def metric_state(self) -> Dict[str, Dict[str, Any]]:
+        """Get the current state of the metric."""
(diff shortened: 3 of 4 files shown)


Borda (approved):
Let's add chlog entry

SkafteNicki (approved):

Borda:
@SkafteNicki this is everywhere: `RuntimeError: Only Tensors created explicitly by the user (graph leaves) support the deepcopy protocol at the moment`

src/torchmetrics/metric.py
```diff
@@ -619,7 +619,7 @@ def wrapped_func(*args: Any, **kwargs: Any) -> Any:
 
             # return cached value
             if self._computed is not None:
-                return self._computed
+                return deepcopy(self._computed)
```
SkafteNicki, line 622: this is definitely messing with all other code

src/torchmetrics/metric.py
```diff
@@ -634,7 +634,8 @@ def wrapped_func(*args: Any, **kwargs: Any) -> Any:
             if self.compute_with_cache:
                 self._computed = value
 
-            return value
+            # Return a deep copy to avoid side effects for non-scalar values, e.g. ConfusionMatrix
+            return deepcopy(value)
```
SkafteNicki, line 638: this is definitely messing with all other code

SkafteNicki (changes requested):
we need to make this work without deepcopy of the values

SkafteNicki (approved):
Okay, changing the `deepcopy` to `clone` works out. Thanks for the contribution!

PR #2626, opened 2024-07-15
Bertscore warnings

baskrahmer:
## What does this PR do?

Fixes a number of warnings when using `BertScore` with the default settings:
```
Some weights of RobertaModel were not initialized from the model checkpoint at roberta-base and are newly initialized: ['roberta.pooler.dense.bias', 'roberta.pooler.dense.weight']
You should probably TRAIN this model on a down-stream task to be able to use it for predictions and inference.
```

and

```
`resume_download` is deprecated and will be removed in version 1.0.0. Downloads always resume when possible. If you want to force a new download, use `force_download=True`.
```

which I think shouldn't be raised when using the metric correctly.

----
📚 Documentation preview 📚: https://torchmetrics--2626.org.readthedocs.build/en/2626/

### DIFF
--- a/src/torchmetrics/functional/text/bert.py
+++ b/src/torchmetrics/functional/text/bert.py
@@ -15,7 +15,7 @@ import csv
 import logging
 import urllib
 from contextlib import contextmanager
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tuple, Union
 
 import torch
 from torch import Tensor
@@ -37,7 +37,7 @@ from torchmetrics.utilities.imports import _TQDM_AVAILABLE, _TRANSFORMERS_GREATE
 
 
 @contextmanager
-def _ignore_log_warning():  # noqa: ANN202
+def _ignore_log_warning() -> Iterator[None]:
     """Ignore irrelevant fine-tuning warning from transformers when loading the model for BertScore."""
     logger = logging.getLogger("transformers.modeling_utils")
     original_level = logger.getEffectiveLevel()
--- a/src/torchmetrics/functional/text/bert.py
+++ b/src/torchmetrics/functional/text/bert.py
@@ -37,8 +37,8 @@ from torchmetrics.utilities.imports import _TQDM_AVAILABLE, _TRANSFORMERS_GREATE
 
 
 @contextmanager
-def _ignore_log_warning():
-    # Ignore irrelevant fine-tuning warning from transformers when loading the model for BertScore
+def _ignore_log_warning():  # noqa: ANN202
+    """Ignore irrelevant fine-tuning warning from transformers when loading the model for BertScore."""
     logger = logging.getLogger("transformers.modeling_utils")
     original_level = logger.getEffectiveLevel()
     try:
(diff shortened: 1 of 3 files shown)


src/torchmetrics/functional/text/bert.py
```diff
@@ -33,6 +35,19 @@
 from torchmetrics.utilities.checks import _SKIP_SLOW_DOCTEST, _try_proceed_with_timeout
 from torchmetrics.utilities.imports import _TQDM_AVAILABLE, _TRANSFORMERS_GREATER_EQUAL_4_4
 
+
+@contextmanager
+def ignore_log_warning():
```
Borda, line 40:
```suggestion
def _ignore_log_warning():
```

src/torchmetrics/functional/text/bert.py
```diff
@@ -41,8 +56,9 @@
 
     def _download_model_for_bert_score() -> None:
         """Download intensive operations."""
-        AutoTokenizer.from_pretrained(_DEFAULT_MODEL, resume_download=True)
-        AutoModel.from_pretrained(_DEFAULT_MODEL, resume_download=True)
+        with ignore_log_warning():
```
Borda, line 59:
```suggestion
        with _ignore_log_warning():
```

src/torchmetrics/functional/text/bert.py
```diff
@@ -356,8 +372,9 @@ def bert_score(
                 " `transformers` model are used."
                 f"It is, therefore, used the default recommended model - {_DEFAULT_MODEL}."
             )
-        tokenizer = AutoTokenizer.from_pretrained(model_name_or_path or _DEFAULT_MODEL)
-        model = AutoModel.from_pretrained(model_name_or_path or _DEFAULT_MODEL)
+        with ignore_log_warning():
```
Borda, line 375:
```suggestion
        with _ignore_log_warning():
```

Borda (approved):

src/torchmetrics/functional/text/bert.py
```diff
@@ -33,6 +35,19 @@
 from torchmetrics.utilities.checks import _SKIP_SLOW_DOCTEST, _try_proceed_with_timeout
 from torchmetrics.utilities.imports import _TQDM_AVAILABLE, _TRANSFORMERS_GREATER_EQUAL_4_4
 
+
+@contextmanager
+def _ignore_log_warning():  # noqa: ANN202
```
Borda, line 40: `mypy` really expects here return type

stancld: Maybe `contextlib.AbstractContextManager`

stancld (approved):

src/torchmetrics/functional/text/bert.py
```diff
@@ -12,7 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import csv
+import logging
 import urllib
+from contextlib import contextmanager
 from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union
```
Borda, line 18:
```suggestion
from typing import Any, Callable, Dict, Iterator, List, Optional, Sequence, Tuple, Union
```

src/torchmetrics/functional/text/bert.py
Borda:
```suggestion
def _ignore_log_warning() -> Iterator[None]:
```
ref: https://stackoverflow.com/a/49736916/4521646

PR #2674, opened 2024-08-03
improve: update `InfoLM` class to dynamically set `higher_is_better`

### LINKED ISSUE #2659: `higher_is_better` for `InfoLM` should change depending on `information_measure`
Not sure if there are other metrics this will apply to but for InfoLM, if the `information_measure` is l2, then the distance will always positive, but `infolm.higher_is_better` shows True. I am pretty sure that is no longer the case compared to the base measure which is KL in which case trending towards 0 from vals < 0 is improvement

grahamannett:
…ed on information_measure

## What does this PR do?

Fixes #2659 

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

I had problems with the devcontainer and the environment.  

Might try and add tests for this but it seems the devcontainer uses the torchmetrics in `/usr/local/lib/python3.9/site-packages` (or similar depending on python) and its not obvious to me how you are supposed to setup things to easily/quickly add tests?  I ended up just doing editable install. 

I manually checked all the values but if I add a test ill try and include some info/example for each of these that shows the metric output on strs.

----
📚 Documentation preview 📚: https://torchmetrics--2674.org.readthedocs.build/en/2674/

### DIFF
--- a/src/torchmetrics/text/infolm.py
+++ b/src/torchmetrics/text/infolm.py
@@ -38,6 +38,21 @@ if not _TRANSFORMERS_GREATER_EQUAL_4_4:
     __doctest_skip__ = ["InfoLM", "InfoLM.plot"]
 
 
+_Information_Measure_Higher_Is_Better = {
+    # following values are <0
+    "kl_divergence": True,
+    "alpha_divergence": True,
+    # following values are >0
+    "beta_divergence": False,
+    "ab_divergence": False,
+    "renyi_divergence": False,
+    "l1_distance": False,
+    "l2_distance": False,
+    "l_infinity_distanc": False,
+    "fisher_rao_distanc": False,
+}
+
+
 class InfoLM(Metric):
     """Calculate `InfoLM`_.
 
@@ -111,7 +126,6 @@ class InfoLM(Metric):
     """
 
     is_differentiable = False
-    higher_is_better = True
     preds_input_ids: List[Tensor]
     preds_attention_mask: List[Tensor]
     target_input_ids: List[Tensor]
@@ -148,6 +162,7 @@ class InfoLM(Metric):
 
         self.tokenizer, self.model = _load_tokenizer_and_model(model_name_or_path, device)
         self.information_measure_cls = _InformationMeasure(information_measure, alpha, beta)
+        self.higher_is_better = _Information_Measure_Higher_Is_Better[information_measure]
         self.max_length = max_length or self.model.config.max_length
         self.special_tokens_map = _get_special_tokens_map(self.tokenizer)
 
--- a/src/torchmetrics/text/infolm.py
+++ b/src/torchmetrics/text/infolm.py
@@ -162,7 +162,6 @@ class InfoLM(Metric):
 
         self.tokenizer, self.model = _load_tokenizer_and_model(model_name_or_path, device)
         self.information_measure_cls = _InformationMeasure(information_measure, alpha, beta)
-        self.higher_is_better = _Information_Measure_Higher_Is_Better[information_measure]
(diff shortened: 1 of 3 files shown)


src/torchmetrics/text/infolm.py
```diff
@@ -38,6 +38,21 @@
     __doctest_skip__ = ["InfoLM", "InfoLM.plot"]
 
 
+_Information_Measure_Higher_Is_Better = {
```
Borda, line 41: can we rather have somehow dynamic, with import the class and ask for its attribute?
my point is not to duplicate this and also prevent forgetting update this part when new metric is added...

Borda: something like:
```py
def _information_measure_higher_is_better(name):
    cls = impotlib.load(f'torchmetrics.retrival.{name}')
    return getatr(cls, 'higher_is_better')
```

src/torchmetrics/text/infolm.py
```diff
@@ -148,6 +162,7 @@ def __init__(
 
         self.tokenizer, self.model = _load_tokenizer_and_model(model_name_or_path, device)
         self.information_measure_cls = _InformationMeasure(information_measure, alpha, beta)
+        self.higher_is_better = _Information_Measure_Higher_Is_Better[information_measure]
```
Borda, line 165: also this need to be a property overwriting the constant attribute

grahamannett: I looked at what @property was being used for and seemed like it would more coincide with setting it like this since its tied to the measure (so wont need to be updated/changed), but sure, will make it a property

src/torchmetrics/text/infolm.py
grahamannett: not sure I understand why you would want it like that?  

was thinking it might make slightly more sense to have it be part of something in `functional/text/infolm.py` but didn't see anything similar so figured I should make the PR/diff as small as possible since it didnt seem to apply to other metrics

Borda: my main point it so load it dynamically without creating another hard-ded Look-up-table

grahamannett: still not sure I follow.  the value will be hardcoded just in another file? and why `torchmetrics.retrival.{name}` instead of off of something in `src/torchmetrics/functional/text/infolm.py`?

Borda: The `higher_is_better` is already an attribute for each class, right?
so setting here again `false/true` is duplicate...

SkafteNicki: @borda I think maybe you are missing the point here that the attribute depends on the input arguments in the class (usually our metrics do not). So I do think the implementation makes sense here.

Borda: My point is not to map to bools but if needed to class names and source the bools directly...

grahamannett:
Im still not sure where to put the info that was requested for the higher is better but incorporated the changes asked for the `@property`related bit.  if you can outline/direct what you would like I will incorporate the changes but seems likely the ways I would decide to implement it/where to store the info isnt what @Borda wants

SkafteNicki:
@borda can you give more specific instructions on what you want changed here?

SkafteNicki:
@Borda should be good to go now

src/torchmetrics/text/infolm.py
SkafteNicki: After discussion with @Borda offline we have come to the conclusion that the current implementation was fine (except that I have moved the dict to be a private attribute of the metric).

Borda (approved):

PR #2698, opened 2024-08-21
Fix `segmentation.MeanIoU`

### LINKED ISSUE #2558: segmentation.MeanIoU is wrong
## 🐛 Bug
MeanIoU scored 56(!) over validation dataset
image

Let's look at the source code:
https://github.com/Lightning-AI/torchmetrics/blob/596ed09b18c3d6786cb094c87da97838228461f3/src/torchmetrics/segmentation/mean_iou.py#L109

```py
def update(self, preds: Tensor, target: Tensor) -> None:
    """Update the state with the new data."""
    intersection, union = _mean_iou_update(preds, target, self.num_classes, self.include_background)
    score = _mean_iou_compute(intersection, union, per_class=self.per_class)
    self.score += score.mean(0) if self.per_class else score.mean()

def compute(self) -> Tensor:
    """Update the state with the new data."""
    return self.score  # / self.num_batches
```

There are several issues there:
- `self.score` is accumulated with each call of `update` method.
- `compute` method just returns the accumulated score
- description of `compute ` method is copypasted from `update` method
- division by `self.num_batches` is commented
- `num_batches` is definded on class level and not used anywhere else

Obviously, that code was neither reviewed nor tested, but somehow was released.

### To Reproduce

Call `metric.update(y_hat, y)` in `validation_step`
Log the metric in `on_validation_epoch_end`

### Expected behavior

MeanIoU computes correct value in [0, 1] range.

### Environment

- TorchMetrics 1.4.0.post0 from pip
- Python 3.11.9, PyTorch 2.3.0+cu121
- pytorch-lightning 2.2.4
- Windows 11

vkinakh:
## What does this PR do?

Fixes #2558 
Fixes a few typos

<details>

</details>

This PR fixes the `segmentation.MeanIoU` metric. 

# Issues
- when the metric is updated via calling `self.update`, the `self.score` is accumulated with each call. Then when `self.compute` is called, accumulated metric is returned
- when the metric is updated via calling `self.forward`, metric works correct

It is happens because when `MeanIoU` is updated via `self.forward`, it calls `self._reduce_states` method, which updates score using the following formula:

`reduced = ((self._update_count - 1) * global_state + local_state).float() / self._update_count`

where `global_state` is the score accumulated over previous steps and `local_state` is the score on current batch.

# Solution

To solve this issue:
- use sum reduce function for `self.score`
- add state `num_batches` to keep number of processed batches
- add increment of `num_batches` in every `self.update` call
- in `self.compute` return sum of scores divided by number of processed batches

----
📚 Documentation preview 📚: https://torchmetrics--2698.org.readthedocs.build/en/2698/

### DIFF
--- a/src/torchmetrics/segmentation/mean_iou.py
+++ b/src/torchmetrics/segmentation/mean_iou.py
@@ -123,7 +123,7 @@ class MeanIoU(Metric):
         self.num_batches += 1
 
     def compute(self) -> Tensor:
-        """Update the state with the new data."""
+        """Compute the final Mean Intersection over Union (mIoU)."""
         return self.score / self.num_batches
 
     def plot(self, val: Union[Tensor, Sequence[Tensor], None] = None, ax: Optional[_AX_TYPE] = None) -> _PLOT_OUT_TYPE:
--- a/src/torchmetrics/segmentation/mean_iou.py
+++ b/src/torchmetrics/segmentation/mean_iou.py
@@ -110,7 +110,8 @@ class MeanIoU(Metric):
         self.input_format = input_format
 
         num_classes = num_classes - 1 if not include_background else num_classes
-        self.add_state("score", default=torch.zeros(num_classes if per_class else 1), dist_reduce_fx="mean")
+        self.add_state("score", default=torch.zeros(num_classes if per_class else 1), dist_reduce_fx="sum")
+        self.add_state("num_batches", default=torch.tensor(0), dist_reduce_fx="sum")
 
     def update(self, preds: Tensor, target: Tensor) -> None:
         """Update the state with the new data."""
@@ -119,10 +120,11 @@ class MeanIoU(Metric):
         )
         score = _mean_iou_compute(intersection, union, per_class=self.per_class)
         self.score += score.mean(0) if self.per_class else score.mean()
+        self.num_batches += 1
 
     def compute(self) -> Tensor:
         """Update the state with the new data."""
-        return self.score  # / self.num_batches
+        return self.score / self.num_batches
 
     def plot(self, val: Union[Tensor, Sequence[Tensor], None] = None, ax: Optional[_AX_TYPE] = None) -> _PLOT_OUT_TYPE:
         """Plot a single or multiple values from the metric.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed wrong aggregation in `segmentation.MeanIoU` ([#2698](https://github.com/Lightning-AI/torchmetrics/pull/2698))
+
+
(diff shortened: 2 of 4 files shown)


vkinakh:
Here is an example code to reproduce the issue

```python
import
from torchmetrics.segmentation import MeanIoU

```

vkinakh:
Here

src/torchmetrics/segmentation/mean_iou.py
```diff
@@ -120,10 +121,11 @@ def update(self, preds: Tensor, target: Tensor) -> None:
         )
         score = _mean_iou_compute(intersection, union, per_class=self.per_class)
         self.score += score.mean(0) if self.per_class else score.mean()
+        self.num_batches += 1
 
     def compute(self) -> Tensor:
-        """Update the state with the new data."""
-        return self.score  # / self.num_batches
```
Borda, line 126: @SkafteNicki any idea why this was commented?

Borda (commented):
great work, can we please add a test to prevent this issue in the future?

src/torchmetrics/segmentation/mean_iou.py
vkinakh: I think the best would be to add a testcase to `MetricTester`, that checks if the return of `self.compute` is the same, when class is updated using `self.forward` and `self.update`

vkinakh:
I have also noticed that names and descriptions in test for `GeneralizedDiceScore` are just copy-paste of `MeanIoU` tests. Tests are correct, but names and descriptions are incorrect and misleading

Should I create a new issue?

https://github.com/Lightning-AI/torchmetrics/blob/24e99c5d80674f4145c0b21027eb9fe579e36bb6/tests/unittests/segmentation/test_generalized_dice_score.py#L66-L108

Borda:
> I have also noticed that names and descriptions in test for `GeneralizedDiceScore` are just copy-paste of `MeanIoU` tests.

that was a typo, fixing it in #2709

SkafteNicki:
Added generalized testing that `forward/update` works as expected (comment https://github.com/Lightning-AI/torchmetrics/pull/2698#discussion_r1727149727) in commit https://github.com/Lightning-AI/torchmetrics/pull/2698/commits/3d3d7b5a6924119bb16293d9dc22e7fdd54eb9dd. Hopefully, this check should pass for all other metrics, else lets move this to another PR.

SkafteNicki (approved):

vkinakh:
> Added generalized testing that `forward/update` works as expected (comment [#2698 (comment)](https://github.com/Lightning-AI/torchmetrics/pull/2698#discussion_r1727149727)) in commit [3d3d7b5](https://github.com/Lightning-AI/torchmetrics/commit/3d3d7b5a6924119bb16293d9dc22e7fdd54eb9dd). Hopefully, this check should pass for all other metrics, else lets move this to another PR.

I suggest adding a test case to verify the metric aggregation behavior: the aggregated metric of a batch with 2*N predictions should match the aggregated metrics from two separate updates, each with N predictions. Are there any metrics that are not expected to behave this way?

Borda (approved):

justusschock (approved):

PR #2840, opened 2024-11-22
Delete Device2Host caused by comm with device and host

zhaozheng09:
## Delete some D2H ops

`if not` will be bring a sync within host and device, so replace it from `if not` to `torch.where`

<details>

</details>

before:
image
after:
image
Host will be block until device done, this D2H maybe block pipeline of host and device, so I use `where op` to replace `if not`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
--
+- Delete `Device2Host` caused by comm with device and host ([#2840](https://github.com/PyTorchLightning/metrics/pull/2840))
 
 
 ---
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Delete `Device2Host` caused by comm with device and host ([#2840](https://github.com/PyTorchLightning/metrics/pull/2840))
+
+
 - Fixed issue with shared state in metric collection when using dice score ([#2848](https://github.com/PyTorchLightning/metrics/pull/2848))
 
 
--- a/src/torchmetrics/utilities/compute.py
+++ b/src/torchmetrics/utilities/compute.py
@@ -15,6 +15,7 @@ from typing import Optional
 
 import torch
 from torch import Tensor
+from typing_extensions import Literal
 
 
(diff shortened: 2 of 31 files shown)


Borda (commented):
can we pls add tests for this fix?

Borda:
> this code will raise in old torch version, I don't know how to cover, please give me an example ? thx very much

ok, so you say that this PR does not have any impact on functional metrics results but improves performances, correct?

Borda (approved):

SkafteNicki (approved):

SkafteNicki:
did a couple of changes to the PR:
* move the logic `torch.where` logic to a separate function such that it could be used multiple places
* made the testing of cpu+sigmoid+half less restrictive e.g. instead of always being skipped it is now only skipped for older pytorch builds

src/torchmetrics/utilities/compute.py
```diff
@@ -184,3 +185,45 @@ def interp(x: Tensor, xp: Tensor, fp: Tensor) -> Tensor:
...
+        tensor([[0.0900, 0.2447, 0.6652],
+                [0.6652, 0.2447, 0.0900]])
+        >>> tensor = torch.tensor([0.0, 0.5, 1.0])
+        >>> normalize_logits_if_needed(tensor, normalization="sigmoid")
+        tensor([0.0000, 0.5000, 1.0000])
+
+    """
+    # decrease sigmoid on cpu .
+    if tensor.device == torch.device("cpu"):
+        if not torch.all((preds >= 0) * (preds <= 1)):
```
Borda, line 219: `preds` is not defined...

justusschock (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Delete `Device2Host` caused by comm with device and host ([#2840](https://github.com/PyTorchLightning/metrics/pull/2840))
```
Borda, line 30:
```suggestion
- Delete `Device2Host` caused by comm with device and host ([#2840](https://github.com/PyTorchLightning/metrics/pull/2840))


```

lantiga (approved):
Looks good, thanks!

In general we will have to become sensitive to data-dependent control flow, to make torchmetrics more compiler-friendly in the future. This applies to both before and after this PR of course, having the condition in a separate function like it is now allows us to side-step the problem, so the final state is better from this point of view.

PR #2861, opened 2024-12-06
Fix `ignore_index`

baskrahmer:
Opening a draft PR since I cannot push to the other branch directly

----
📚 Documentation preview 📚: https://torchmetrics--2861.org.readthedocs.build/en/2861/


baskrahmer:
The remaining failing tests have to do with [this line](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/utilities/compute.py#L79-L80):
```python
        if not multilabel:
            weights[tp + fp + fn == 0 if top_k == 1 else tp + fn == 0] = 0.0
```

Which is different in the reference metric `_reference_sklearn_precision_recall_multiclass`, which does include classes even if there are no labels of that class present (`tp + fn == 0`).

If we follow `sklearn`, then I believe some of the expected values in the tests will need to be updated in order to make them pass.

baskrahmer:
Closing since changes are now in the other PR

PR #2979, opened 2025-02-27
Fix corner case in manually specifying `compute_groups` in `MetricCollection`

### LINKED ISSUE #2897: Extend support for manual specification of MetricCollection `compute_groups`
## 🚀 Feature

Support incomplete specification of MetricCollection `compute_groups` + inherit compute groups from child metric collections

### Motivation

Currently, new instances of the `MetricCollection` class allow users to manually specify the `compute_groups` used to simplify `update` steps. However, if a user doesn't explicitly specify any metric in the compute groups lists, it simply won't be updated. It seems like the expected behavior would be for those additional metrics to simply be updated normally (or possibly automatically inherited by some of the manually specified groups).

Moreover, if a uses the supported behavior of a single level "recursion" of MetricCollection's (i.e. creating a MetricCollection of MetricCollection's), the `compute_groups` arguments of the child collections are simply ignored.

### Pitch

1) It might be desirable to have any metrics which aren't included in the list of `compute_groups` to be automatically detected and put into their own single-item compute groups. Possibly we could also attempt to merge them into existing groups, but this might be overly complex and unnecessary.

2) The specified compute groups from child collections should be inherited by the parent (with the prefix's/postfix's updated appropriately). Admittedly, this introduces some complexities which may be more simply handled by just automatically creating the groups (as is currently done). If so, it should at least be documented that the child `compute_groups` arguments will be ignored.

### Alternatives

As noted above, if the current behavior is intended, it should at least be documented what will happen in these "corner" cases.

SkafteNicki:
## What does this PR do?

Fixes #2897

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--2979.org.readthedocs.build/en/2979/

### DIFF
--- a/src/torchmetrics/collections.py
+++ b/src/torchmetrics/collections.py
@@ -24,7 +24,7 @@ from typing_extensions import Literal
 
 from torchmetrics.metric import Metric
 from torchmetrics.utilities import rank_zero_warn
-from torchmetrics.utilities.data import _flatten_dict, allclose
+from torchmetrics.utilities.data import _flatten, _flatten_dict, allclose
 from torchmetrics.utilities.imports import _MATPLOTLIB_AVAILABLE
 from torchmetrics.utilities.plot import _AX_TYPE, _PLOT_OUT_TYPE, plot_single_or_multi_val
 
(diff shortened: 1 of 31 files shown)


Borda (approved):

src/torchmetrics/collections.py
```diff
@@ -210,7 +211,7 @@ def __init__(
         self._enable_compute_groups = compute_groups
         self._groups_checked: bool = False
         self._state_is_copy: bool = False
-
+        self._groups: Dict[int, List[str]] = {}
```
Borda, line 214:
```suggestion
        self._groups: Dict[int, Sequence[str]] = {}
```

src/torchmetrics/collections.py
```diff
@@ -210,7 +211,7 @@ def __init__(
         self._enable_compute_groups = compute_groups
         self._groups_checked: bool = False
         self._state_is_copy: bool = False
-
+        self._groups: Dict[int, Sequence[str]] = {}
```
Borda, line 214:
```suggestion
        self._groups: Dict[int, list[str]] = {}
```

PR #3008, opened 2025-03-14
Fix edge case in integration between classwise wrapper and metric tracker

### LINKED ISSUE #3007: `IndexError` when calling `MetricTracker.best_metric()` for `ClasswiseWrapper` nested in `MetricCollection`
## 🐛 Bug

Following the improved support of `ClasswiseWrapper` in `MetricTracker` to fix [this issue](https://github.com/Lightning-AI/torchmetrics/issues/2994), there is still an edge-case that fails. (Thanks a lot for the very quick response and fix to the issue BTW 😉 )

When `ClasswiseWrapper` is nested in a `MetricCollection` and `maximize=None`, `MetricTracker` builds a `maximize` list based on the `higher_is_better` attribute of its base metrics. In the introduced fix, `higher_is_better` is only one boolean value for `ClasswiseWrapper`. However, `best_metric` expects the `maximize` list to have the same length as the resulting scores, i.e. the number of classes. Thus, calling `best_metric` fails on an `IndexError: list index out of range`.

I have tried various other combinations of tracker params and collections, and no others raised an error.  E.g., when `maximize` is set in `MetricTracker`'s init, the `maximize` list is created dynamically based on the length of the resulting scores, which works out-of-the-box. And `MetricTracker` also works when applied directly on the `ClasswiseWrapper`, without the nesting inside a `MetricCollection`.

### To Reproduce
The following minimal example reproduces the bug:

<details>

```python
from torch import tensor
from torchmetrics import ClasswiseWrapper, MetricTracker
from torchmetrics.classification import MulticlassAccuracy

target = tensor([0, 1, 2, 3])
preds = tensor([0, 2, 1, 3])

classwise_metric = ClasswiseWrapper(MulticlassAccuracy(num_classes=4, average=None))
collection = MetricCollection([classwise_metric])

collection_tracker = MetricTracker(collection, maximize=None)
collection_tracker.increment()
collection_tracker.update(target, preds)
collection_tracker.best_metric()
```

</details>

<details>
  <summary>Error traceback of the above example</summary>
The last line fails with following error:

```console
        else:  # this is a metric collection
            maximize = self.maximize if isinstance(self.maximize, list) else len(res) * [self.maximize]
            value, idx = {}, {}  # type: ignore[assignment]
            for i, (k, v) in enumerate(res.items()):
                try:
>                   fn = torch.max if maximize[i] else torch.min
E                   IndexError: list index out of range
```

</details>

### Expected behavior

The above example should return the following result:
```python
>>> collection_tracker.best_metric()
{'multiclassaccuracy_0': 1.0, 'multiclassaccuracy_1': 0.0, 'multiclassaccuracy_2': 0.0, 'multiclassaccuracy_3': 1.0}
```

### Environment

- TorchMetrics version (if build from source, add commit SHA): recent master branch, commit SHA: 0423a2c6d04c8cb0f4afc74bf2d6be7db6884b74

nathanpainchaud:
## What does this PR do?

Fixes #3007

An alternative fix for the issue could be to change the definition of the `higher_is_better` property for the `ClasswiseWrapper` to broadcast the base metric's property to a list of length `num_classes`. This approach would work out-of-the-box, w/o the change to the `MetricTracker` code and test proposed here.

However, I found this alternative solution potentially more confusing for the end user, who might not expect a list as the `higher_is_better` property of the `ClasswiseWrapper`. Although I don't know if `higher_is_better` is meant to be user-facing in the first place...

Therefore, feel free to update my proposed fix with the alternative solution if you think the simplified implementation and maintenance cost is worth it :slightly_smiling_face: 

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--3008.org.readthedocs.build/en/3008/


Borda (approved):

src/torchmetrics/wrappers/tracker.py
```diff
@@ -134,7 +134,10 @@ def __init__(
                             f"The metric '{name}' in the MetricCollection does not have a 'higher_is_better' attribute."
                             " Please provide the `maximize` argument explicitly."
                         )
-                    self.maximize.append(m.higher_is_better)  # type: ignore[arg-type]  # this is false alarm
+                    m_higher_is_better = [m.higher_is_better]
+                    if isinstance(m, ClasswiseWrapper):
+                        m_higher_is_better *= m.metric.num_classes
```
Borda, line 139:
```suggestion
                        m_higher_is_better = m_higher_is_better * m.metric.num_classes
```

SkafteNicki (approved):

PR #3018, opened 2025-03-18
Add `aggregation_level` option to `DiceScore` and exclude samples with zero denominator from score calculation

### LINKED ISSUE #3014: Resolve confusion between `classification.Dice` and `segmentation.DiceScore` and improve the implementation
## 🚀 Feature

Resolve confusion around dice score implementations and improve the current implementation.

### Motivation

Currently there are two different implementations for dice in `torchmetrics`:
- `torchmetrics.classification.Dice`
- `torchmetrics.segmentation.DiceScore`

It seems that originally the idea was to deprecate the old `Dice` implementation, as mentioned in https://github.com/Lightning-AI/torchmetrics/pull/2725.

Not only is the coexistence of those two implementations confusing, there are also significant differences in the implementation of the two metrics, which can lead to significantly different scores in downstream tasks, especially in multi-class settings. For instance in `macro` mode, `DiceScore` by default calculates the pixel overlaps/unions first separately per sample and then when doing the division `numerator/denominator` uses a fallback value of `zero_division` (https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/functional/segmentation/dice.py#L94). `Dice` on the other hand doesn't calculate per-sample (see `mdmc_average="global"`) and excludes entries where the denominator is 0 (https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/functional/classification/dice.py#L49).

Most of those differences stem from different possible strategies of dealing with 0 values in the denominator of the dice metric. A few possible choices here are:
1. Ignore samples/entries where denominator is 0 (this is what `classification.Dice` does in `macro` mode)
2. Ignore samples where gnd truth is empty (this is what `monai.metrics.DiceMetric` does by default, which can be controlled through the `ignore_empty` init argument)
3. Ignore samples where both the prediction and the ground truth is empty (i.e. the denominator for that sample is all 0).
4. Use `1.0` or `0.0` fallback values for `0` divisions (`segmentation.DiceScore` uses a `1.0` default fallback for `torchmetrics` versions `<=1.6.1` and a `0.0` fallback value for version `1.6.2`)
5. Always use `nan` as fallback value combined with `torch.nanmean` and `torch.nansum` aggregation functions
6. Before doing the `numerator/denominator` divison, do a `torch.sum(numerator, dim=0)`, `torch.sum(denominator, dim=0)`, summing up the pixel overlap & union counts accross all samples. You then shouldn't have any 0 values in the denominator anymore, assuming that each class occurs at least once in one sample in the ground truth.

While each choice has its pros/cons, I do think that option 4. which `segmentation.DiceScore` currently implements is particularly problematic. Especially for rare classes, this pulls the aggregated score towards the fallback value, and as a consequence, the resulting score cannot be interpreted visually as "overlap" anymore, as it leads to either **overestimating** or **underestimating** the real overlap / score. To visualize this, here a real example, where a model yields a much higher aggregated macro dice score due to the fallback value being set to 1.0 (this experiment used `torchmetrics==1.6.1`):
Image
Note that the orange model performs significantly worse than blue when looking at the classwise dice metrics. However, the aggregated metric is much higher, because the score here is driven towards the fallback value, which is `1.0` for `torchmetrics==1.6.1`.

### Pitch

- Deprecate `torchmetrics.classification.Dice` to avoid confusion between the two metrics, which also calculate different things.
- Improve the implementation of `torchmetrics.segmentation.Dice`. I recommend moving away from `0.0` / `1.0` fallback values. To me, option 3. makes the most sense (note that option 2. which MONAI implements and which is similar to 3. has the issue of not measuring false positives on samples where a class is not present at all). Also, implementing 6. can be added as an additional option by introducing a `aggregation_level: Literal["sample", "global"]` init argument. This can be easily implemented by adding `torch.sum` at the beginning of `_dice_score_compute`:
```python
    if aggregation_level == "global":
        numerator = torch.sum(numerator, dim=0)
        denominator = torch.sum(denominator, dim=0)
        support = torch.sum(support, dim=0) if support is not None else None
```

Given that dice is the main metric used for semantic segmentation tasks nowadays, I think improving this implementation would have a big impact both on research & industry usecases.

nkaenzig:
## What does this PR do?

Fixes #3014

- For `macro` & `weighted`, calculates the score while excluding samples where no ground truth nor prediction is present (i.e. `denominator == 0`)
- Adds `aggregation_level` option, which when set to `global` will sum up the overlaps & unions across all samples, before calculating the final dice score.

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--3018.org.readthedocs.build/en/3018/


Borda (approved):

Borda:
could we pls add tests for it :)

nkaenzig:
> could we pls add tests for it :)

I updated now the existing tests, and added a new one which tests the case where not all samples contain all classes.

nkaenzig:
@Borda, @SkafteNicki do you think this can be merged soon?

src/torchmetrics/functional/segmentation/dice.py
```diff
@@ -108,6 +113,8 @@ def dice_score(
     include_background: bool = True,
     average: Optional[Literal["micro", "macro", "weighted", "none"]] = "micro",
     input_format: Literal["one-hot", "index"] = "one-hot",
+    aggregation_level: Optional[Literal["samplewise", "global"]] = "samplewise",
+    reduce: bool = False,
```
SkafteNicki, line 117:
```suggestion
```

src/torchmetrics/functional/segmentation/dice.py
```diff
@@ -120,6 +127,10 @@ def dice_score(
           or ``None``. This determines how to average the dice score across different classes.
         input_format: What kind of input the function receives. Choose between ``"one-hot"`` for one-hot encoded tensors
           or ``"index"`` for index tensors
+        aggregation_level: The level at which to aggregate the dice score. Options are ``"samplewise"`` or ``"global"``.
+            For ``"samplewise"`` the dice score is computed for each sample and then averaged. For ``"global"`` the dice
+            score is computed globally over all samples.
+        reduce: Whether to reduce the output to a single number. Default is ``False``.
```
SkafteNicki, line 133:
```suggestion
```

src/torchmetrics/functional/segmentation/dice.py
```diff
@@ -164,4 +175,5 @@ def dice_score(
         )
     _dice_score_validate_args(num_classes, include_background, average, input_format)
     numerator, denominator, support = _dice_score_update(preds, target, num_classes, include_background, input_format)
-    return _dice_score_compute(numerator, denominator, average, support=support)
+    dice = _dice_score_compute(numerator, denominator, average, aggregation_level=aggregation_level, support=support)
+    return torch.nanmean(dice, dim=0) if reduce else dice
```
SkafteNicki, line 179:
```suggestion
    return _dice_score_compute(numerator, denominator, average, aggregation_level=aggregation_level, support=support)
```

src/torchmetrics/functional/segmentation/dice.py
```diff
@@ -120,6 +127,10 @@ def dice_score(
           or ``None``. This determines how to average the dice score across different classes.
         input_format: What kind of input the function receives. Choose between ``"one-hot"`` for one-hot encoded tensors
           or ``"index"`` for index tensors
+        aggregation_level: The level at which to aggregate the dice score. Options are ``"samplewise"`` or ``"global"``.
```
SkafteNicki, line 130: please add an example in the docstring for the two settings

src/torchmetrics/segmentation/dice.py
```diff
@@ -112,8 +115,8 @@ def __init__(
         num_classes: int,
         include_background: bool = True,
         average: Optional[Literal["micro", "macro", "weighted", "none"]] = "micro",
+        aggregation_level: Optional[Literal["samplewise", "global"]] = "samplewise",
         input_format: Literal["one-hot", "index"] = "one-hot",
-        zero_division: Union[float, Literal["warn", "nan"]] = 0.0,
```
SkafteNicki, line 116: why are we removing the `zero_division` argument?

src/torchmetrics/functional/segmentation/dice.py
```diff
@@ -28,7 +28,6 @@ def _dice_score_validate_args(
     include_background: bool,
     average: Optional[Literal["micro", "macro", "weighted", "none"]] = "micro",
     input_format: Literal["one-hot", "index"] = "one-hot",
-    zero_divide: Union[float, Literal["warn", "nan"]] = 1.0,
 ) -> None:
```
SkafteNicki, line 31:
```suggestion
    aggregation_level: Optional[Literal["samplewise", "global"]] = "samplewise"
) -> None:
```

src/torchmetrics/functional/segmentation/dice.py
```diff
@@ -40,10 +39,6 @@ def _dice_score_validate_args(
         raise ValueError(f"Expected argument `average` to be one of {allowed_average} or None, but got {average}.")
     if input_format not in ["one-hot", "index"]:
         raise ValueError(f"Expected argument `input_format` to be one of 'one-hot', 'index', but got {input_format}.")
```
SkafteNicki, line 41:
```suggestion
        raise ValueError(f"Expected argument `input_format` to be one of 'one-hot', 'index', but got {input_format}.")
    if aggregation_level not in ("samplewise", "global"):
        raise ValueError(f"Expected argument `aggregation_level` to be one of `samplewise`, `global`, but got {aggregation_level}")
```

src/torchmetrics/functional/segmentation/dice.py
nkaenzig: Added an example now.

src/torchmetrics/segmentation/dice.py
nkaenzig: Good question, in the issue description I explain why using fallback values for the zero divisions is problematic. In the new implementation, whenever a zero division occurs, it will use "nan" instead, which then in the aggregation calls will be excluded from the calculation.

image

nkaenzig:
@Borda @SkafteNicki Thanks for your review. I think I have addressed all your comments, are there any other blockers to merge this?

SkafteNicki (approved):

PR #3094, opened 2025-05-14
fix: blocking variable instantiation to prevent race condition

### LINKED ISSUE #3095: Non-blocking call in _safe_divide leads to race condition
## 🐛 Bug

There's an apparent race condition here:

https://github.com/Lightning-AI/torchmetrics/blob/e79093558ae48f52cb5882685b07545649171230/src/torchmetrics/utilities/compute.py#L77-L78

When moving the tensor to the target device (MPS in my case), I get sometimes the correct default (`0.0`) but sometimes uninitialized numbers, screwing up everything.

<details>

- TorchMetrics version (if build from source, add commit SHA): `1.7.1`
- Python & PyTorch Version (e.g., 1.0): Python 3.12.10, PyTorch 2.7.0
- Any other relevant information such as OS (e.g., Linux): macOS, Darwin Kernel Version 24.4.0

</details>

wsascha:
## What does this PR do?

Fixes #3095 

<details>

</details>

<details>
  <summary>PR review</summary>

This PR fixes a race condition from a non-blocking call.

</details>

## Did you have fun?

Not so much fun debugging this, but a lot of fun to help the community! 😄 

----
📚 Documentation preview 📚: https://torchmetrics--3094.org.readthedocs.build/en/3094/

### DIFF
--- a/src/torchmetrics/utilities/compute.py
+++ b/src/torchmetrics/utilities/compute.py
@@ -74,7 +74,7 @@ def _safe_divide(
         if zero_division == "warn" and torch.any(denom == 0):
             rank_zero_warn("Detected zero division in _safe_divide. Setting 0/0 to 0.0")
         zero_division = 0.0 if zero_division == "warn" else zero_division
-        zero_division_tensor = torch.tensor(zero_division, dtype=num.dtype).to(num.device, non_blocking=True)
+        zero_division_tensor = torch.tensor(zero_division, dtype=num.dtype, device=num.device)
         return torch.where(denom != 0, num / denom, zero_division_tensor)
     return torch.true_divide(num, denom)
 


Borda:
There are so many failing tests on GPU and mainly with `RuntimeError: called a synchronizing CUDA operation`

wsascha:
I think the magic happens here:

https://github.com/Lightning-AI/torchmetrics/blob/4c42cf2c29f48f0024bbceb612672c20006a88ec/tests/unittests/classification/test_accuracy.py#L365-L373

However, I’m not sure how to handle this, assuming everything works fine for CUDA-capable devices.

Borda:
> However, I’m not sure how to handle this, assuming everything works fine for CUDA-capable devices.

I think this would be the right way #3101

wsascha:
> > However, I’m not sure how to handle this, assuming everything works fine for CUDA-capable devices.
> 
> I think this would be the right way #3101

Good for me if it’s just the MPS device that makes problems. On the other hand, I’m also wondering if it’s really an advantage to construct a tensor on CPU and transfer it to GPU, compared to just creating it on GPU (at least that’s what seems to be happening). Yes, there may be a sync, bit I’m curious what the actual performance benefit is in that case.

Anyway, I’ll close this PR then. Thanks for your time!

PR #3104, opened 2025-05-25
Fix: Discard zero and negative predictions in retrieval metrics

AymenKallala:
## What does this PR do?

Retrieval metrics are currently considering all input predictions as positive predictions regardless of their probability values. This was leading to unexpected numbers.

For instance, an input prediction tensor ([0.0,0.0,0.0]) is equivalent to no prediction and should always return 0 (for all retrieval metrics). A simple way to fix this problem is to use a mask of positive predictions to calculate the metrics.

This PR fixes this bug for the following metrics:
- Precision
- Mean Average Precision
- Recall
- MRR (reciprocal_rank)
- NDCG

### DIFF
--- a/src/torchmetrics/functional/retrieval/average_precision.py
+++ b/src/torchmetrics/functional/retrieval/average_precision.py
@@ -50,11 +50,17 @@ def retrieval_average_precision(preds: Tensor, target: Tensor, top_k: Optional[i
 
     top_k = top_k or preds.shape[-1]
     if not isinstance(top_k, int) and top_k <= 0:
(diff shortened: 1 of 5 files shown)


Borda (commented):
could we pls add test for this edge-case?

Borda:
@rittik9 could you pls help here? 🐰

Isalia20:
I can take a look at this

Borda:
@Isalia20 have you have a chance to look at it? :)

Borda:
or @rittik9 could help to land this PR? 🦩

PR #3251, opened 2025-09-04
Bugfix for `GeneralizedDiceScore` to yield `NaN` for missing classes

### LINKED ISSUE #2846: `GeneralizedDiceScore` yields 0 scores when using `per_class=True` for samples where class is not present
## 🐛 Bug

The current implementation of `GeneralizedDiceScore` yields scores of `0.0` for samples that don't contain a particular class when calculating class-wise metrics via `per_class=True`.

This leads to very low dice scores, particularly for rare classes and therefore makes the dice scores between classes incomparable.  

### To Reproduce

The following code sample should print `tensor([0.2500, 0.2500, 0.0000])`, even though all the predictions match the targets:

<details>

```python
import torch
from torchmetrics.segmentation import GeneralizedDiceScore
from torchmetrics.segmentation import DiceScore

N_SAMPLES = 4
N_CLASSES = 3

target = torch.full((N_SAMPLES, N_CLASSES, 128, 128), 0, dtype=torch.int8)
preds = torch.full((N_SAMPLES, N_CLASSES, 128, 128), 0, dtype=torch.int8)

target[0, 0] = 1
target[2, 1] = 1
preds[0, 0] = 1
preds[2, 1] = 1

generalized_dice = GeneralizedDiceScore(num_classes=3, per_class=True, include_background=True)
print(generalized_dice(preds, target))
```

</details>

### Expected behavior

I'd expect the above code sample to either return `[1.0, 1.0, 1.0]` or `[1.0, 1.0, nan]` for the class-wise cores (it could be argued if `nan` should be returned for the third class, given that this class is not present in any of the samples, therefore returning a 1.0 score might also be misleading).

### Environment

- TorchMetrics version (if build from source, add commit SHA): `1.6.0`
- Python & PyTorch Version (e.g., 1.0): `3.11.10`
- Any other relevant information such as OS (e.g., Linux): macOS 15.1.1 (24B91)

### Additional context

The new `DiceScore` implementation which was released in `1.6.0` ([PR](https://github.com/Lightning-AI/torchmetrics/pull/2725)) seems to have solved this issue - in the above example it returns `[1.0, 1.0, 1.0]`.

VijayVignesh1:
## What does this PR do?
Fixes GeneralizedDiceScore to yield NaN for missing classes when per_class=True.
Added testcases to verify the same.

Fixes #2846

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--3251.org.readthedocs.build/en/3251/


Borda (dismissed):

src/torchmetrics/segmentation/generalized_dice.py
```diff
@@ -100,15 +100,16 @@ class GeneralizedDiceScore(Metric):
         tensor(0.4992)
         >>> gds = GeneralizedDiceScore(num_classes=3, per_class=True)
         >>> gds(preds, target)
-        tensor([0.5001, 0.4993, 0.4982])
+        tensor([5.0008, 4.9930, 4.9825])
```
rittik9: ?? dice score range 0 to 1

src/torchmetrics/segmentation/generalized_dice.py
```diff
@@ -100,15 +100,16 @@ class GeneralizedDiceScore(Metric):
...
         >>> gds = GeneralizedDiceScore(num_classes=3, per_class=True, include_background=False)
         >>> gds(preds, target)
-        tensor([0.4993, 0.4982])
+        tensor([4.9930, 4.9825])
```
rittik9: ??

Borda (approved):

tests/unittests/segmentation/test_generalized_dice_score.py
```diff
@@ -82,6 +82,8 @@ def _reference_generalized_dice(
 class TestGeneralizedDiceScore(MetricTester):
     """Test class for `GeneralizedDiceScore` metric."""
 
+    atol = 2e-2
```
Borda:
```suggestion
    atol = 2e-3
```

src/torchmetrics/segmentation/generalized_dice.py
```diff
@@ -133,20 +134,29 @@ def __init__(
...
+        if self.per_class:
+            class_mask = target.sum(dim=(0, *range(2, target.ndim))) > 0
+            self.class_present += class_mask[1:] if not self.include_background else class_mask
+            self.numerator = torch.sum(self.numerator, dim=0, keepdim=True)
+            self.denominator = torch.sum(self.denominator, dim=0, keepdim=True)
 
     def compute(self) -> Tensor:
         """Compute the final generalized dice score."""
-        return self.score / self.samples
+        score = _generalized_dice_compute(self.numerator, self.denominator, self.per_class)
```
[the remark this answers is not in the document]
VijayVignesh1: The comment is wrong since the summing in the update function is performing the intended global accumulation across batches. Instead of storing a list of batch-wise numerators/denominators, we are collapsing them into global per-class totals immediately. Deferring this to compute would only add overhead without changing correctness.

src/torchmetrics/segmentation/generalized_dice.py
```diff
@@ -133,20 +134,29 @@ def __init__(
...
         numerator, denominator = _generalized_dice_update(
             preds, target, self.num_classes, self.include_background, self.weight_type, self.input_format
         )
-        self.score += _generalized_dice_compute(numerator, denominator, self.per_class).sum(dim=0)
-        self.samples += preds.shape[0]
+        self.numerator = torch.cat([self.numerator, numerator], dim=0)
+        self.denominator = torch.cat([self.denominator, denominator], dim=0)
+        if self.per_class:
+            class_mask = target.sum(dim=(0, *range(2, target.ndim))) > 0
+            self.class_present += class_mask[1:] if not self.include_background else class_mask
```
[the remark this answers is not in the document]
VijayVignesh1: This is not true. We need to track class_present state only for per_class=True scenario. For per_class=False, the final mean calculation would take care of the NaN scenario.

Borda:
@VijayVignesh1 dont worry, Copilot is not always right but it may spot some overlooked issues

tests/unittests/segmentation/test_generalized_dice_score.py
VijayVignesh1: The testcases seem to failing for ubuntu and mac for atol = 2e-3. It passes when it's 2e-2. Can we make it 2e-2?

src/torchmetrics/segmentation/generalized_dice.py
```diff
@@ -133,20 +135,32 @@ def __init__(
         self.input_format = input_format
 
         num_classes = num_classes - 1 if not include_background else num_classes
-        self.add_state("score", default=torch.zeros(num_classes if per_class else 1), dist_reduce_fx="sum")
-        self.add_state("samples", default=torch.zeros(1), dist_reduce_fx="sum")
+        self.add_state("class_present", default=torch.zeros(num_classes, dtype=torch.int), dist_reduce_fx="sum")
+        self.add_state("numerator", default=[], dist_reduce_fx="cat")
+        self.add_state("denominator", default=[], dist_reduce_fx="cat")
```
justusschock: why are these changes required? this changes the memory profile of this metric quite a bit.

src/torchmetrics/functional/segmentation/generalized_dice.py
```diff
@@ -90,6 +90,7 @@ def _generalized_dice_compute(numerator: Tensor, denominator: Tensor, per_class:
     if not per_class:
         numerator = torch.sum(numerator, 1)
         denominator = torch.sum(denominator, 1)
+        return _safe_divide(numerator, denominator, "nan")
```
justusschock: can we not just add "nan" at the divide below?

src/torchmetrics/segmentation/generalized_dice.py
VijayVignesh1: Yes, I just realized it. I've removed them and made the changes :)

src/torchmetrics/functional/segmentation/generalized_dice.py
VijayVignesh1: Done!

src/torchmetrics/functional/segmentation/generalized_dice.py
```diff
@@ -126,10 +129,7 @@ def generalized_dice_score(
         >>> generalized_dice_score(preds, target, num_classes=5)
         tensor([0.4830, 0.4935, 0.5044, 0.4880])
         >>> generalized_dice_score(preds, target, num_classes=5, per_class=True)
-        tensor([[0.4724, 0.5185, 0.4710, 0.5062, 0.4500],
-                [0.4571, 0.4980, 0.5191, 0.4380, 0.5649],
-                [0.5428, 0.4904, 0.5358, 0.4830, 0.4724],
-                [0.4715, 0.4925, 0.4797, 0.5267, 0.4788]])
+        tensor([[0.4845, 0.4997, 0.4993, 0.4864, 0.4912]])
```
justusschock: the tensorshape shouldn't change here!

src/torchmetrics/functional/segmentation/generalized_dice.py
```diff
@@ -139,10 +139,7 @@ def generalized_dice_score(
         >>> generalized_dice_score(preds, target, num_classes=5, input_format="index")
         tensor([0.1991, 0.1971, 0.2350, 0.2216])
         >>> generalized_dice_score(preds, target, num_classes=5, per_class=True, input_format="index")
-        tensor([[0.1714, 0.2500, 0.1304, 0.2524, 0.2069],
-                [0.1837, 0.2162, 0.0962, 0.2692, 0.1895],
-                [0.3866, 0.1348, 0.2526, 0.2301, 0.2083],
-                [0.1978, 0.2804, 0.1714, 0.1915, 0.2783]])
+        tensor([[0.2234, 0.2170, 0.1597, 0.2399, 0.2204]])
```
justusschock: also here the shape should not change.

src/torchmetrics/functional/segmentation/generalized_dice.py
VijayVignesh1: I've made the changes.

src/torchmetrics/functional/segmentation/generalized_dice.py
VijayVignesh1: Done

src/torchmetrics/segmentation/generalized_dice.py
```diff
@@ -133,20 +135,26 @@ def __init__(
...
 
     def update(self, preds: Tensor, target: Tensor) -> None:
         """Update the state with new data."""
         numerator, denominator = _generalized_dice_update(
             preds, target, self.num_classes, self.include_background, self.weight_type, self.input_format
         )
-        self.score += _generalized_dice_compute(numerator, denominator, self.per_class).sum(dim=0)
-        self.samples += preds.shape[0]
+        self.numerator.append(numerator)
```
justusschock, line 146: this still changes the memory format if you add to a list rather than adding to a tensor

VijayVignesh1: Can you please elaborate this a bit more?

RecreationalMath:
Hi, I was investigating #2846 and took a look at why this PR stalled. I'd like to help move it forward.

On the memory-format concern @justusschock raised: I checked this out locally and measured state memory directly. Master holds ~44 bytes of state regardless of how many `update()` calls you make. This PR's `List[Tensor]` grows linearly:

| updates | PR state size |
| ------: | ------------: |
|   1,000 |        1.2 MB |
|  10,000 |         12 MB |
|  62,500 |         76 MB |

That's fine for a single val sweep, but the regression is **unbounded in the number of `update()` calls** (vs master's O(1)), plus 125k+ individual tensor objects at the high end, and `compute()` transiently doubles that footprint via `dim_zero_cat`. For a metric that logically only needs running sums, that's concerning.

To preserve master's memory profile: The per-class numerator and denominator are sum-reducible across batches (so they become two fixed-size tensor sums). Master's `per_class=False` path averages per-sample ratios rather than doing a global sum/sum, so to preserve that semantic we can track two scalars, a running sum of valid per-sample scores and a count of valid samples. All four states are fixed-size, O(C) total:

```python
self.add_state("numerator_sum",   default=torch.zeros(n), dist_reduce_fx="sum")
self.add_state("denominator_sum", default=torch.zeros(n), dist_reduce_fx="sum")
# per_class=False: sample-wise mean, excluding degenerate (all-zero target) samples
self.add_state("score_sum",     default=torch.zeros(1), dist_reduce_fx="sum")
self.add_state("valid_samples", default=torch.zeros(1), dist_reduce_fx="sum")

def update(self, preds, target):
    num, den = _generalized_dice_update(...)  # [N, C]
    self.numerator_sum   += num.sum(dim=0)
    self.denominator_sum += den.sum(dim=0)
    per_sample = _safe_divide(num.sum(dim=1), den.sum(dim=1), "nan")
    self.score_sum     += torch.nansum(per_sample)
    self.valid_samples += (~torch.isnan(per_sample)).sum()

def compute(self):
    if self.per_class:
        return _safe_divide(self.numerator_sum, self.denominator_sum, "nan")
    return _safe_divide(self.score_sum, self.valid_samples, "nan").squeeze()
```

I ran this against the #2846 repro on current master:

```
  per_class=True : tensor([1., 1., nan])   # matches the expected output
  per_class=False: tensor(1.)              # two valid samples, both 1.0
  multi-update   : tensor([1., 1., nan])   # same under split update() calls
  all-empty      : nan / [nan, nan, nan]   # degenerate case still signals nan
```

For `per_class=True` this converges to the same "global GDS" aggregation the current PR produces (sum num / sum den), it just does it incrementally instead of buffering per-sample tensors.

One thing worth flagging in the current PR regardless of which structure ends up merged: the `per_class=False` path does `score.mean(dim=0)` over per-sample ratios post-`_safe_divide(..., "nan")`, so a single empty-target sample mixed with real ones turns the whole result into NaN (`mean([1.0, nan, 1.0, nan]) = nan`). Existing tests only cover the all-empty case, so this slips through. Either `.nanmean()` or a `nansum`/valid-count pattern would avoid it.

I'm happy to drive this to merge. Two options depending on PR author's availability:
- If @VijayVignesh1 is still active on this, I'm glad to contribute the changes directly to PR #3251.
- If not, I can open a follow-up PR that builds on #3251 and credits @VijayVignesh1 as co-author, whichever closes out #2846 fastest.

@Borda @justusschock, does this direction work? Happy to start once I have a thumbs up.

VijayVignesh1:
> Hi, I was investigating #2846 and took a look at why this PR stalled. I'd like to help move it forward.
> 
> On the memory-format concern @justusschock raised: I checked this out locally and measured state memory directly. Master holds ~44 bytes of state regardless of how many `update()` calls you make. This PR's `List[Tensor]` grows linearly:
> updates 	PR state size
> 1,000 	1.2 MB
> 10,000 	12 MB
> 62,500 	76 MB
> 
> That's fine for a single val sweep, but the regression is **unbounded in the number of `update()` calls** (vs master's O(1)), plus 125k+ individual tensor objects at the high end, and `compute()` transiently doubles that footprint via `dim_zero_cat`. For a metric that logically only needs running sums, that's concerning.
> 
> To preserve master's memory profile: The per-class numerator and denominator are sum-reducible across batches (so they become two fixed-size tensor sums). Master's `per_class=False` path averages per-sample ratios rather than doing a global sum/sum, so to preserve that semantic we can track two scalars, a running sum of valid per-sample scores and a count of valid samples. All four states are fixed-size, O(C) total:
> 
> ```python
> self.add_state("numerator_sum",   default=torch.zeros(n), dist_reduce_fx="sum")
> self.add_state("denominator_sum", default=torch.zeros(n), dist_reduce_fx="sum")
> # per_class=False: sample-wise mean, excluding degenerate (all-zero target) samples
> self.add_state("score_sum",     default=torch.zeros(1), dist_reduce_fx="sum")
> self.add_state("valid_samples", default=torch.zeros(1), dist_reduce_fx="sum")
> 
> def update(self, preds, target):
>     num, den = _generalized_dice_update(...)  # [N, C]
>     self.numerator_sum   += num.sum(dim=0)
>     self.denominator_sum += den.sum(dim=0)
>     per_sample = _safe_divide(num.sum(dim=1), den.sum(dim=1), "nan")
>     self.score_sum     += torch.nansum(per_sample)
>     self.valid_samples += (~torch.isnan(per_sample)).sum()
> 
> def compute(self):
>     if self.per_class:
>         return _safe_divide(self.numerator_sum, self.denominator_sum, "nan")
>     return _safe_divide(self.score_sum, self.valid_samples, "nan").squeeze()
> ```
> 
> I ran this against the #2846 repro on current master:
> 
> ```
>   per_class=True : tensor([1., 1., nan])   # matches the expected output
>   per_class=False: tensor(1.)              # two valid samples, both 1.0
>   multi-update   : tensor([1., 1., nan])   # same under split update() calls
>   all-empty      : nan / [nan, nan, nan]   # degenerate case still signals nan
> ```
> 
> For `per_class=True` this converges to the same "global GDS" aggregation the current PR produces (sum num / sum den), it just does it incrementally instead of buffering per-sample tensors.
> 
> One thing worth flagging in the current PR regardless of which structure ends up merged: the `per_class=False` path does `score.mean(dim=0)` over per-sample ratios post-`_safe_divide(..., "nan")`, so a single empty-target sample mixed with real ones turns the whole result into NaN (`mean([1.0, nan, 1.0, nan]) = nan`). Existing tests only cover the all-empty case, so this slips through. Either `.nanmean()` or a `nansum`/valid-count pattern would avoid it.
> 
> I'm happy to drive this to merge. Two options depending on PR author's availability:
> 
>     * If @VijayVignesh1 is still active on this, I'm glad to contribute the changes directly to PR [Bugfix for `GeneralizedDiceScore` to yield `NaN` for missing classes #3251](https://github.com/Lightning-AI/torchmetrics/pull/3251).
> 
>     * If not, I can open a follow-up PR that builds on [Bugfix for `GeneralizedDiceScore` to yield `NaN` for missing classes #3251](https://github.com/Lightning-AI/torchmetrics/pull/3251) and credits @VijayVignesh1 as co-author, whichever closes out [`GeneralizedDiceScore` yields 0 scores when using `per_class=True` for samples where class is not present #2846](https://github.com/Lightning-AI/torchmetrics/issues/2846) fastest.
> 
> 
> @Borda @justusschock, does this direction work? Happy to start once I have a thumbs up.

Sure, we can work on the same PR!

PR #3270, opened 2025-09-22
Feature/brier score

### LINKED ISSUE #2196: Implement the Brier score and it's decomposition into resolution, reliability and uncertainty.
## 🚀 Feature

I would like to contribute to torchmetrics, by implementing the Brier score and its associated decomposition. 

### Motivation

The Brier score is widely used when measuring the calibration of machine learning methods see:

https://arxiv.org/abs/2302.04019  
https://arxiv.org/abs/2002.06470

It is also a proper scoring rule as opposed to the Expected Calibration Error (ECE) and Thresholded Calibration Error (TACE). This means that the ECE and the TACE have trivial minima where the classifier has **zero** test accuracy **while being perfectly calibrated** (https://arxiv.org/abs/1906.02530). The Brier score being a proper scoring rule doesn't have this pathological behaviour.

The Brier score coincides with the mean squared error for common use cases. However, its decomposition into resolution, reliability and uncertainty see https://en.wikipedia.org/wiki/Brier_score is a unique and useful feature. Roughly speaking `resolution` captures a notion of accuracy and `reliability` a notion of calibration. Thus both have to be optimized for the Brier score to be low. 

Finally, no standard implementation in common packages exists to the best of my knowledge. 

### Pitch

I plan to follow the original paper describing the decomposition of the Brier score into resolution, reliability and uncertainty

https://journals.ametsoc.org/view/journals/apme/12/4/1520-0450_1973_012_0595_anvpot_2_0_co_2.xml

and specifically the implementation found in 

https://github.com/google-research/google-research/blob/master/uq_benchmark_2019/metrics_lib.py
and the paper
https://arxiv.org/abs/1906.02530

The decomposition into uncertainty, resolution and reliability was originally formulated for predictions which take a finite set of values. This is in contrast with the output vectors of most deep neural network classifiers which output a vector of probabilities per class, which take continuous values. Thus we need to create bins for our output vectors. The specific bins in this implementation are with respect to the top most probable class for each input signal. Thus we create C bins where C is the number of classes. Then two prediction vectors [0 , 0.9, 0.1] and [0.2, 0.6, 0.2] fall in the same bin, the bin of class 2. The derivation of resolution, reliability and uncertainty is then relatively straighforward.

konstantinos-p:
## What does this PR do?

Fixes #2196

Implement Brier score and it's decomposition

- The ECE and other similar calibration metrics such as the TACE, are not proper scoring rules. In particular, an untrained model with no predictive accuracy will minimize the ECE. This pathological behaviour can be fixed by using a proper scoring rule such as the Brier score.
- The Brier score can be decomposed into Uncertainty, Reliability and Resolution. Reliability measures the model's calibration why Resolution correlates with model accuracy. Uncertainty, is the inherent difficulty of the problem.
- The Brier score is used in many papers that evaluate the calibration of classification models. ex https://arxiv.org/abs/2302.04019 https://arxiv.org/abs/2002.06470

I followed the original paper describing the decomposition of the Brier score into resolution, reliability and uncertainty

https://journals.ametsoc.org/view/journals/apme/12/4/1520-0450_1973_012_0595_anvpot_2_0_co_2.xml

and specifically the implementation found in 

https://github.com/google-research/google-research/blob/master/uq_benchmark_2019/metrics_lib.py
and the paper
https://arxiv.org/abs/1906.02530

State of the PR
I added some rudimentary "tests" and the code seems to work for the Binary and Multiclass settings.

@SkafteNicki 

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--3270.org.readthedocs.build/en/3270/

### DIFF
--- a/src/torchmetrics/classification/brier.py
+++ b/src/torchmetrics/classification/brier.py
@@ -18,14 +18,17 @@ from torch import Tensor
 from typing_extensions import Literal
 
 from torchmetrics.classification.base import _ClassificationTaskWrapper
+from torchmetrics.functional.classification.brier import (
+    _binary_brier_format,
+    _mean_brier_score_and_decomposition,
+    _multiclass_brier_format,
+)
 from torchmetrics.functional.classification.confusion_matrix import (
     _binary_confusion_matrix_arg_validation,
-    _binary_confusion_matrix_compute,
     _binary_confusion_matrix_format,
     _binary_confusion_matrix_tensor_validation,
     _binary_confusion_matrix_update,
     _multiclass_confusion_matrix_arg_validation,
-    _multiclass_confusion_matrix_compute,
     _multiclass_confusion_matrix_format,
     _multiclass_confusion_matrix_tensor_validation,
     _multiclass_confusion_matrix_update,
@@ -35,16 +38,11 @@ from torchmetrics.functional.classification.confusion_matrix import (
     _multilabel_confusion_matrix_tensor_validation,
     _multilabel_confusion_matrix_update,
 )
-from torchmetrics.functional.classification.brier import (
-    _mean_brier_score_and_decomposition,
-    _binary_brier_format,
-    _multiclass_brier_format,
-)
 from torchmetrics.metric import Metric
+from torchmetrics.utilities.data import dim_zero_cat
 from torchmetrics.utilities.enums import ClassificationTask
 from torchmetrics.utilities.imports import _MATPLOTLIB_AVAILABLE
 from torchmetrics.utilities.plot import _AX_TYPE, _CMAP_TYPE, _PLOT_OUT_TYPE, plot_confusion_matrix
-from torchmetrics.utilities.data import dim_zero_cat
 
 if not _MATPLOTLIB_AVAILABLE:
     __doctest_skip__ = [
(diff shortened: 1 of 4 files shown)


src/torchmetrics/classification/brier.py
```diff
@@ -0,0 +1,565 @@
...
+if not _MATPLOTLIB_AVAILABLE:
+    __doctest_skip__ = [
+        "BinaryBrier.plot",
+        "MulticlassBrier.plot",
+        "MultilabelBrier.plot",
+    ]
+
+
+class BinaryBrier(Metric):
+    r"""Compute the `confusion matrix`_ for binary tasks.
```
Borda, line 56: sounds strange...

konstantinos-p: To be clear this is still early WIP. I could turn it into a draft(?)

justusschock:
hey @konstantinos-p , thanks for your work so far. There's definitely interest in new metrics. Especially after all the work you've put in so far.

What's the relation of the brier score to https://scikit-learn.org/stable/modules/generated/sklearn.metrics.brier_score_loss.html ? is it the inverse/1-score or is it the same with different naming?

Otherwise there'd be https://scikit-survival.readthedocs.io/en/stable/api/generated/sksurv.metrics.brier_score.html as well

If we have no reference implementation, we'd generally be fine comparng against hand-computed values as well but it seems like this wouldn't be necessary here.

PR #3309, opened 2025-12-02
Add `metric_state` property to `CompositionalMetric` for advance logging in lightning

SkafteNicki:
## What does this PR do?

Fixes https://github.com/Lightning-AI/pytorch-lightning/issues/20153
PR does the following:
* Adds a `metric_state` property to `CompositionalMetric`
* By utilizing this property in `MetricCollection`, we can now make sure that `CompositonalMetric` created from indexing the same underlying metric will end up in the same compute group and therefore only be calculated once even though they are now "separate" metrics
* This will allow for a more advance logging in lightning of metrics that are non-scalar by doing something like this

```python
# initialize a nons-scalar returning metric, in this case we use MeanAveragePrecision
# this metric returns a dict, each being a scalar tensor, assume we are only interested in a few of them
map = torchmetrics.detection.MeanAveragePrecision()

# index into the metric the values we want to log and repack into a MetricCollection
# MetricCollection compute group feature will make sure that calculations are only done once
self.metrics = torchmetrics.MetricCollection(
    {
        "map_50": map["map_50"],
        "map_75": map["map_75"],
        "map_small": map["map_small"],
    },
    prefix="val_",
)
# then all can be logged using log_dict
self.log_dict(self.metrics)
```

This has the intended purpose of getting users to just log the metric object instead of calling `compute` and `forward` then logging the values themself, which can lead to problems (see linked issue).

<details>

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--3309.org.readthedocs.build/en/3309/

### DIFF
--- a/.lightning/workflows/uinttests.yml
+++ b/.lightning/workflows/uinttests.yml
@@ -21,6 +21,7 @@ env:
   TEST_DIRS: "unittests"
   FREEZE_REQUIREMENTS: 1
   SKIP_SLOW_DOCTEST: 1
+  LIGHTNING_CI: "1"
 
 run: |
   whereis nvidia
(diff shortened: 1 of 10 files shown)


src/torchmetrics/metric.py
```diff
@@ -1224,6 +1224,16 @@ def __init__(
         else:
             self.metric_b = metric_b
 
+    @property
+    def metric_state(self) -> dict[str, Union[List[Tensor], Tensor]]:
```
justusschock, line 1228: should we just call this `state`?

Since it's a metric, it's naturally a `metric_state` :D

PR #3317, opened 2026-01-20
fix(lit ci/gpu): stabilize GPU tests, fix multiprocessing + DDP race conditions

bhimrazy:
## What does this PR do?

This PR stabilizes GPU test execution on the master branch by addressing multiprocessing failures, DDP-related race conditions, and excessive CI runtimes. 

 image
 image
</p>

> The goal is to unblock CI, reduce flakiness, and make GPU test behavior consistent across Python and PyTorch versions.

Also closes #3298 #3315
closed #3307 #3308

### Summary of Changes

1. Optimize parallel test execution
    - Set `numprocesses=3` for GPU test runs.
    - Empirically chosen based on non-DDP GPU test benchmarks:
        - **2 or 4 processes** → ~45 minutes
        - **3 processes** → ~36 minutes (best tradeoff)
        - **>4 processes** → exceeds 1 hour / unstable
    - This avoids GPU oversubscription and reduces contention during multiprocessing-heavy tests.

2. Fix multiprocessing daemon assertion failures
     -  `AssertionError: daemonic processes are not allowed to have children`
     - This was triggered when tests attempted to spawn child processes from daemonized workers, leading to cascading failures and leaked semaphores in audio-related tests (e.g. `TestDNSMOS`).
     - ref: [ci](https://lightning.ai/lightning-ai/ci/jobs/ci-run-lightning-ai-torchmetrics-84ddc31-uinttests-yml-lit-job-pytorchlightning-torchmetrics-ubuntu24-04-cuda12-6-3-py3-12-torch2-8-ab5da0aa?app_id=jobs&job_detail_tab=logs)

3. Correct DDP test classification and execution
    The following tests were incorrectly executed as non-DDP tests, causing port conflicts and race conditions:
    
    - `test_ms_ssim_reduction_none_ddp`
    - `test_ssim_reduction_none_ddp`
    - `test_score_ddp`
    - ref: [ci](https://lightning.ai/lightning-ai/ci/jobs/ci-run-lightning-ai-torchmetrics-ef13e2d-uinttests-yml-lit-job-pytorchlightning-torchmetrics-ubuntu22-04-cuda12-1-1-py3-9-torch2-0-cc2575dc?app_id=jobs&job_detail_tab=logs)

    These are now correctly marked and executed as **DDP-only** tests and run with the appropriate process pool configuration.

### Skipped tests (Lightning CI only)
To unblock GPU CI and allow other PRs to progress, the following test is temporarily skipped only in Lightning CI:
  - `tests/unittests/audio/test_sdr.py::TestSDR::test_sdr`
  Once this test is skipped, all remaining GPU tests complete successfully and reliably.
  
 This test will be investigated and re-enabled in a follow-up PR once the root cause of the DDP timeout / blocking behavior is identified.

<details><summary>Details</summary>
<p>

- ref ci: [3.9](https://lightning.ai/lightning-ai/ci/jobs/ci-run-lightning-ai-torchmetrics-ad37967-uinttests-yml-lit-job-pytorchlightning-torchmetrics-ubuntu22-04-cuda12-1-1-py3-9-torch2-0-b2069b10?app_id=jobs&job_detail_tab=logs)
> image

- ref ci: [3.12](https://lightning.ai/lightning-ai/ci/jobs/ci-run-lightning-ai-torchmetrics-ad37967-uinttests-yml-lit-job-pytorchlightning-torchmetrics-ubuntu24-04-cuda12-6-3-py3-12-torch2-8-ee11a69e?app_id=jobs&job_detail_tab=logs)
> image

</p>
</details> 

## Did you have fun?

----
📚 Documentation preview 📚: https://torchmetrics--3317.org.readthedocs.build/en/3317/

### DIFF
--- a/.lightning/workflows/uinttests.yml
+++ b/.lightning/workflows/uinttests.yml
@@ -5,7 +5,7 @@ trigger:
     branches: ["master"]
 
 timeout: "300" # minutes
-machine: "L4_X_4"
+machine: "L4_X_2"
 parametrize:
   matrix:
     image: # note that this is setting also all oldest requirements which is linked to Torch == 2.0
@@ -68,7 +68,7 @@ run: |
 
   # DocTesting
   cd src/
-  timeout 1800 python -m pytest torchmetrics --cov=torchmetrics \
+  python -m pytest torchmetrics --cov=torchmetrics \
     --timeout=240 --durations=50 \
     --reruns 2 --reruns-delay 1
   cd ..
@@ -81,14 +81,14 @@ run: |
   #du -h --max-depth=1 .
   # UnitTesting common
   #if [ "$testing" == "common" ]; then
-  timeout 7000 python -m pytest ${TEST_DIRS} -v \
-    -m "not DDP" --numprocesses=4 --dist=loadfile \
+  timeout 5000 python -m pytest ${TEST_DIRS} -v \
+    -m "not DDP" --numprocesses=3 --dist=loadfile \
     --cov=torchmetrics --timeout=360 --durations=100 \
     --reruns 3 --reruns-delay 1
   # UnitTesting DDP
   #elif [ "$testing" == "distributed" ]; then
   export USE_PYTEST_POOL="1"
-  timeout 7000 python -m pytest ${TEST_DIRS} -v \
+  timeout 5000 python -m pytest ${TEST_DIRS} -v \
     --cov=torchmetrics -m "DDP" \
     --timeout=360 --durations=100
   #fi
--- a/.lightning/workflows/uinttests.yml
+++ b/.lightning/workflows/uinttests.yml
@@ -5,7 +5,7 @@ trigger:
     branches: ["master"]
 
 timeout: "300" # minutes
-machine: "L4_X_2"
+machine: "L4_X_4"
 parametrize:
   matrix:
     image: # note that this is setting also all oldest requirements which is linked to Torch == 2.0
@@ -82,7 +82,7 @@ run: |
   # UnitTesting common
   #if [ "$testing" == "common" ]; then
   timeout 7000 python -m pytest ${TEST_DIRS} -v \
-    -m "not DDP" --numprocesses=2 --dist=loadfile \
+    -m "not DDP" --numprocesses=4 --dist=loadfile \
     --cov=torchmetrics --timeout=360 --durations=100 \
     --reruns 3 --reruns-delay 1
   # UnitTesting DDP
--- a/.lightning/workflows/uinttests.yml
+++ b/.lightning/workflows/uinttests.yml
@@ -81,14 +81,14 @@ run: |
   #du -h --max-depth=1 .
   # UnitTesting common
   #if [ "$testing" == "common" ]; then
-  timeout 4500 python -m pytest ${TEST_DIRS} -v \
+  timeout 7000 python -m pytest ${TEST_DIRS} -v \
     -m "not DDP" --numprocesses=3 --dist=loadfile \
     --cov=torchmetrics --timeout=360 --durations=100 \
     --reruns 3 --reruns-delay 1
   # UnitTesting DDP
   #elif [ "$testing" == "distributed" ]; then
   export USE_PYTEST_POOL="1"
-  timeout 4500 python -m pytest ${TEST_DIRS} -v \
+  timeout 7000 python -m pytest ${TEST_DIRS} -v \
     --cov=torchmetrics -m "DDP" \
     --timeout=360 --durations=100
   #fi
--- a/.lightning/workflows/uinttests.yml
+++ b/.lightning/workflows/uinttests.yml
@@ -80,14 +80,14 @@ run: |
   #du -h --max-depth=1 .
   # UnitTesting common
   #if [ "$testing" == "common" ]; then
-  # timeout 3600 python -m pytest ${TEST_DIRS} -v \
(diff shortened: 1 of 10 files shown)


tests/unittests/_helpers/__init__.py
```diff
@@ -38,3 +39,4 @@ def seed_all(seed):
 _TORCH_LESS_THAN_2_1 = RequirementCache("torch<2.1.0")
 _TRANSFORMERS_RANGE_GE_4_50_LT_4_54 = RequirementCache("transformers>=4.50.0,<4.54.0")
 _TRANSFORMERS_GREATER_EQUAL_4_54 = RequirementCache("transformers>=4.54.0")
+_IS_LIGHTNING_CI = os.environ.get("LIGHTNING_CI", "0") == "1"
```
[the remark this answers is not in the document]
bhimrazy: this can be skipped.

tests/unittests/image/test_ssim.py
```diff
@@ -365,33 +362,27 @@ def test_ssim_for_correct_padding():
...
+    for _ in range(3):
+        x, y = torch.rand(4, 3, 224, 224).to(device).chunk(2)
+        metric.update(x, y)
 
-        result = metric.compute()
-        assert isinstance(result, torch.Tensor), "Expected compute result to be a tensor"
-    finally:
-        cleanup_ddp()
+    result = metric.compute()
+    assert isinstance(result, torch.Tensor), "Expected compute result to be a tensor"
```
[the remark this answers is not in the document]
bhimrazy: nice catch, updated. 
Thanks

tests/unittests/image/test_ms_ssim.py
```diff
@@ -110,33 +107,27 @@ def test_ms_ssim_contrast_sensitivity():
...
+    for _ in range(3):
+        x, y = torch.rand(4, 3, 224, 224).to(device).chunk(2)
+        metric.update(x, y)
 
-        result = metric.compute()
-        assert isinstance(result, torch.Tensor), "Expected compute result to be a tensor"
-    finally:
-        cleanup_ddp()
+    result = metric.compute()
+    assert isinstance(result, torch.Tensor), "Expected compute result to be a tensor"
```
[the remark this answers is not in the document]
bhimrazy: nice catch, updated. 
Thanks

rittik9 (approved):

tchaton (approved):
Nice work !

.lightning/workflows/uinttests.yml
```diff
@@ -81,7 +82,7 @@ run: |
   # UnitTesting common
   #if [ "$testing" == "common" ]; then
   python -m pytest ${TEST_DIRS} -v \
-    -m "not DDP" --numprocesses=6 --dist=loadfile \
+    -m "not DDP" --numprocesses=3 --dist=loadfile \
```
justusschock, line 85: How was this number determined? Seems kind of arbitrary

tests/unittests/audio/test_sdr.py
```diff
@@ -69,7 +69,14 @@ class TestSDR(MetricTester):
...
 
-    @pytest.mark.parametrize("ddp", [pytest.param(True, marks=pytest.mark.DDP), False])
+    @pytest.mark.parametrize(
+        "ddp",
+        [
+            pytest.param(True, marks=[pytest.mark.DDP]),
+            False,
+        ],
+    )
+    @pytest.mark.skipif(_IS_LIGHTNING_CI, reason="test too slow and unreliable on Lightning CI")
```
justusschock, line 79: @bhimrazy could you check why exactly it is slow and unreliable there? We control all sides there so maybe we can do something about this. Not a blocker for this PR though

justusschock (approved):

.lightning/workflows/uinttests.yml
bhimrazy: Sure @justusschock, I’ve already added the details in the description.

> image

The time taken for the non-DDP tests varied when changing the number of processes. It also seemed to depend on the number of GPUs available.

On Azure, the machine had 8 GPUs, so it didn’t create much pressure. But here, with only 2 GPUs available, it looked like the tests were running into race conditions and stepping on each other. This caused some tests to take more than 20 minutes for a single run, which then had a cascading impact on the overall runtime.

After testing with a few different configurations, the values shown in the screenshot were finalized.

We’ll probably need to experiment a bit more to bring the runtime closer to the Azure setup, but for now this at least unblocks us.

tests/unittests/audio/test_sdr.py
bhimrazy: Sure @justusschock, I did some initial investigation on this as well, which is mentioned in the description above, though it may not be very clearly explained there.

What I found is that when this test is enabled, for some reason this single test takes more than 20 minutes to run, which then delays the other tests as well but when it's disabled, the overall runtime is normal.. This behavior only seems to happen for the image with Python 3.12.

image

My assumption is that this could be related to the mir_eval library, which also contains deprecated codes. I briefly looked into whether it could be replaced with [fast_bss_eval](https://github.com/fakufaku/fast_bss_eval) (as suggested upstream), but that will likely require a deeper investigation.

> Also, I came to this conclusion by comparing it with the Python 3.9 image, where the issue does not occur. The 3.9 image uses an older version of the mir_eval package (and generally older dependencies overall), so the slowdown could also be influenced by other package differences.

PR #3337, opened 2026-03-17
  Fix MetricCollection state comparison for nested sequence states

omkar-334:
## What does this PR do?

Fixes #3335

<details>

</details>

<details>
  <summary>PR review</summary>
  
 Fixes a bug where `MetricCollection` crashes while auto-merging compute groups for metrics with nested sequence state, such as
  `MeanAveragePrecision`.
  
  Previously, `_equal_metric_states` assumed that list-valued state always contained tensors and accessed `.shape` directly on each element. This
  breaks for metrics whose state contains tuples or other nested structures, leading to:

  ```python
  AttributeError: 'tuple' object has no attribute 'shape'
  ```
  
  
  Fix -  i replaced the tensor-only list comparison with a recursive state comparision function
  
  

----
📚 Documentation preview 📚: https://torchmetrics--3337.org.readthedocs.build/en/3337/

### DIFF
--- a/src/torchmetrics/collections.py
+++ b/src/torchmetrics/collections.py
@@ -56,6 +56,23 @@ def _remove_suffix(string: str, suffix: str) -> str:
     return string[: -len(suffix)] if string.endswith(suffix) else string
 
 
+def _equal_state_value(state1: Any, state2: Any) -> bool:
+    """Recursively compare metric state values while preserving structure checks."""
+    if type(state1) is not type(state2):
+        return False
+
+    if isinstance(state1, Tensor):
+        return state1.shape == state2.shape and allclose(state1, state2)
+
+    if isinstance(state1, Mapping):
+        return state1.keys() == state2.keys() and all(_equal_state_value(state1[k], state2[k]) for k in state1)
+
+    if isinstance(state1, Sequence) and not isinstance(state1, str):
+        return len(state1) == len(state2) and all(_equal_state_value(s1, s2) for s1, s2 in zip(state1, state2))
+
+    return state1 == state2
+
+
 class MetricCollection(ModuleDict):
     """MetricCollection class can be used to chain metrics that have the same call pattern into one single class.
 
@@ -316,21 +333,7 @@ class MetricCollection(ModuleDict):
             state1 = getattr(metric1, key)
             state2 = getattr(metric2, key)
 
-            if type(state1) != type(state2):  # noqa: E721
-                return False
-
-            if (
-                isinstance(state1, Tensor)
-                and isinstance(state2, Tensor)
-                and not (state1.shape == state2.shape and allclose(state1, state2))
-            ):
-                return False
-
-            if (
-                isinstance(state1, list)
-                and isinstance(state2, list)
-                and not (all(s1.shape == s2.shape and allclose(s1, s2) for s1, s2 in zip(state1, state2)))
-            ):
+            if not _equal_state_value(state1, state2):
                 return False
 
         return True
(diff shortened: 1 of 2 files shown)


m-matthias:
Is there any update on this? It would be great if metric collections containing e.g. MeanAveragePrecision can be used with compute_groups enabled.

src/torchmetrics/collections.py
```diff
@@ -56,6 +56,23 @@ def _remove_suffix(string: str, suffix: str) -> str:
     return string[: -len(suffix)] if string.endswith(suffix) else string
 
 
+def _equal_state_value(state1: Any, state2: Any) -> bool:
```
Borda, line 59: The _equal_state_value docstring is one line. Consider listing supported types for callers: e.g., "Recursively compare metric state values. Supports: Tensor (shape+value), Mapping (key+recursive value), Sequence/str-excluded (length+recursive element), and primitives (direct ==)."

src/torchmetrics/collections.py
```diff
@@ -56,6 +56,23 @@ def _remove_suffix(string: str, suffix: str) -> str:
...
+    """Recursively compare metric state values while preserving structure checks."""
+    if type(state1) is not type(state2):
```
Borda, line 61: `type(state1) is not type(state2)` is intentionally strict (exact type match), which is the right call here -- but worth a comment since most Python code uses != or isinstance. Consider # noqa: E721 or a short note.

src/torchmetrics/collections.py
```diff
@@ -56,6 +56,23 @@ def _remove_suffix(string: str, suffix: str) -> str:
...
+        return False
+
+    if isinstance(state1, Tensor):
+        return state1.shape == state2.shape and allclose(state1, state2)
+
+    if isinstance(state1, Mapping):
+        return state1.keys() == state2.keys() and all(_equal_state_value(state1[k], state2[k]) for k in state1)
+
+    if isinstance(state1, Sequence) and not isinstance(state1, str):
+        return len(state1) == len(state2) and all(_equal_state_value(s1, s2) for s1, s2 in zip(state1, state2))
```
Borda, line 71: The Mapping and Sequence branches use generator expressions with all(). For very deep nesting (e.g., list of list of list of ...) this could hit Python recursion limits. In practice metric states are rarely >3 levels deep, so this is theoretical -- but note the limit if it matters for your use case.

Borda (commented):
Thanks for tracking down this issue and the PR is well-targeted. A few things stand out:

The root cause was clear: _equal_metric_states used isinstance(state1, list) which misses tuple states that MeanAveragePrecision stores. The fix generalizes this to all Sequence types via a recursive helper.

The new _equal_state_value function is clean -- it handles Tensor, Mapping, Sequence (with string exclusion), and falls back to direct == for primitives. No new imports needed, no API changes.

The test case is well-constructed: DummyNestedListMetric faithfully reproduces the MeanAveragePrecision state shape (list[tuple[tensor, tensor]]), and the assertions verify both the compute-group merging and the state comparison.

Thematic areas to consider:

The helper function is internal (not exported in __init__.py) which is correct, but the docstring could benefit from mentioning what types it supports.
For future robustness, the MeanAveragePrecision reproduction in the linked issue would be a valuable addition (or a comment noting the fix covers it).
Consider whether dict states with tensor values are explicitly tested somewhere -- the recursive Mapping branch is new and worth confirming.
I've left a couple of minor observations below.

m-matthias:
Is it possible to merge this? Or does somebody need to address the comments first?

PR #781, opened 2022-01-19
Fix Matthews correlation coefficient when the denominator is 0

SkafteNicki:
## What does this PR do?

Fixes #779

## Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [x] Did you write any new necessary tests?

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #3371, opened 2026-05-04
fix: return 0 for MCC when denominator is zero (align with sklearn)

jashshah999:
## What does this PR do?

When any row or column of the confusion matrix sums to zero, the MCC denominator is zero and the metric is undefined. The previous code had complex special-casing with eps perturbations that produced inconsistent results (e.g. 1.0 when all predictions and targets are 0, while sklearn returns 0.0).

Replace with the sklearn convention: return 0 when denominator is zero.

## Before (inconsistent with sklearn)

```python
# All negatives: TorchMetrics=1.0, sklearn=0.0
preds = torch.zeros(10)
targets = torch.zeros(10)
BinaryMatthewsCorrCoef()(preds, targets)  # 1.0 (wrong)
```

## After (matches sklearn)

```python
BinaryMatthewsCorrCoef()(preds, targets)  # 0.0 (correct)
```

The fix removes ~25 lines of complex special-casing and replaces with a 1-line return.

Fixes #3355.

----
📚 Documentation preview 📚: https://torchmetrics--3371.org.readthedocs.build/en/3371/


mateenali66 (commented):
ran this against the MCC tests on current master. baseline is 12 passed, with the patch applied it is 9 failed and 3 passed.

```
FAILED test_corner_cases[binary_matthews_corrcoef-preds0-target0-1.0]  - assert tensor(0.) == 1.0
FAILED test_corner_cases[binary_matthews_corrcoef-preds1-target1-1.0]  - assert tensor(0.) == 1.0
FAILED test_corner_cases[binary_matthews_corrcoef-preds4-target4--1.0] - assert tensor(0.) == -1.0
FAILED test_corner_cases[binary_matthews_corrcoef-preds5-target5--1.0] - assert tensor(0.) == -1.0
FAILED test_corner_cases[metric_fn6-preds6-target6-1.0]                - assert tensor(0.) == 1.0
FAILED test_corner_cases[metric_fn7-preds7-target7-1.0]                - assert tensor(0.) == 1.0
FAILED test_corner_cases[metric_fn8-preds8-target8--1.0]               - assert tensor(0.) == -1.0
FAILED test_corner_cases[metric_fn9-preds9-target9--1.0]               - assert tensor(0.) == -1.0
FAILED test_matthews_corrcoef_reduce                                   - assert tensor(0.) != 0
```

worth separating those, because two different things are breaking and only one of them is the design question.

**the eight corner cases are the real disagreement.** `test_corner_cases` encodes the current convention deliberately: one class only and predicted perfectly is 1.0, one class only and predicted perfectly wrong is -1.0. sklearn calls the same situation undefined and returns 0.0. both readings are defensible, and whichever wins, those eight parameters have to move with the code. at the moment the PR changes the behaviour and leaves the test asserting the old values, so it cannot go green as it stands.

**`test_matthews_corrcoef_reduce` is a separate problem.** it fails on the first matrix only:

```python
confmat_tp_zero = torch.tensor([[19392673, 1], [76216, 0]]).to(torch.bfloat16)
```

that one is not mathematically undefined. its true MCC in float64 is -0.0000142080. the denominator reaches zero only because `s**2` underflows in bfloat16, so `denom == 0` there is a precision artefact rather than a degenerate confusion matrix, and the eps branch being removed is what was covering it. the other two matrices in that test have nonzero denominators and are unaffected.

returning 0.0 instead of -1.42e-5 is a harmless numerical difference in practice, so this one is fixable by relaxing the assert rather than by keeping the eps machinery. it just should not happen as a silent side effect, since #781 and #1863 added that branch for exactly this case.

**on the design question.** torchmetrics already has a `zero_division` argument on 34 other classification metrics, including precision_recall, f_beta, jaccard and stat_scores. MCC not having one is the odd case out, and an opt-in avoids a breaking default.

one caveat if you go that way. a scalar `zero_division` cannot reproduce today's behaviour on its own, because the current code is sign aware, +1.0 for perfectly predicted single class and -1.0 for perfectly wrong. so the default would have to mean "keep the current convention" rather than a number, with `zero_division=0.0` selecting sklearn's.

the branch is also BEHIND master, so it needs a rebase before CI will say anything useful.

PR #3422, opened 2026-06-25
fix(image): correct PSNRB `block_size` input validation (and → or)

### LINKED ISSUE #3364: PSNRB: `block_size` input validation is broken
## 🐛 Bug

Input validation for `block_size` in `PeakSignalNoiseRatioWithBlockedEffect` is broken: invalid values either slip through silently or raise `TypeError`, instead of the intended `ValueError`.

[`src/torchmetrics/image/psnrb.py:82`](https://github.com/Lightning-AI/torchmetrics/blob/master/src/torchmetrics/image/psnrb.py#L82) uses `and` where `or` is intended:

```python
if not isinstance(block_size, int) and block_size < 1:
    raise ValueError("Argument ``block_size`` should be a positive integer")
```

With `and`, the first clause (`not isinstance(..., int)`) is False for every int, which short-circuits the expression and skips the `< 1` check, so invalid ints like `0` and `-5` slip through. For non-ints, the second comparison runs and either returns False (e.g. `1.5 < 1`) or raises `TypeError` (e.g. `"foo" < 1`). Neither path reaches the intended `ValueError`.

| Input | Observed | Expected |
|---|---|---|
| `0` | no error | `ValueError` |
| `-5` | no error | `ValueError` |
| `1.5` | no error | `ValueError` |
| `"foo"` | `TypeError: '<' not supported between instances of 'str' and 'int'` | `ValueError` |

### To Reproduce

<details>

```python
from torchmetrics.image import PeakSignalNoiseRatioWithBlockedEffect

# All four should raise ValueError per the existing error message.

# Silently accepted (no error):
PeakSignalNoiseRatioWithBlockedEffect(data_range=1.0, block_size=0)
PeakSignalNoiseRatioWithBlockedEffect(data_range=1.0, block_size=-5)
PeakSignalNoiseRatioWithBlockedEffect(data_range=1.0, block_size=1.5)

# Raises TypeError instead of ValueError:
PeakSignalNoiseRatioWithBlockedEffect(data_range=1.0, block_size="foo")
```

</details>

<details>

- TorchMetrics version: 1.9.0
- Python version: 3.12.13
- PyTorch version: 2.11.0
- OS: macOS 26.4.1

</details>

Sriniketh24:
## Summary

Fixes #3364

The validation guard on `block_size` in `PeakSignalNoiseRatioWithBlockedEffect.__init__` uses `and` instead of `or`:

```python
# Before (buggy):
if not isinstance(block_size, int) and block_size < 1:
```

Because `isinstance(-1, int)` is `True`, the first clause is `False` and Python short-circuits the `and`, so the `< 1` check is never evaluated. Invalid integers like `0`, `-1`, `-5` pass through silently with no error.

Non-integers (e.g. `1.5`, `"foo"`) do raise an error, but a `TypeError` (from the `< 1` comparison) rather than the documented `ValueError`.

## Fix

```python
# After:
if not isinstance(block_size, int) or block_size < 1:
    raise ValueError(f"Argument `block_size` should be a positive integer, got {block_size!r}")
```

With `or`, both conditions are checked independently. The error message now includes the invalid value for easier debugging.

## Tests added

```python
@pytest.mark.parametrize("block_size", [0, -1, -5, 1.5, "foo"])
def test_error_on_invalid_block_size(block_size):
    with pytest.raises(ValueError, match="block_size"):
        PeakSignalNoiseRatioWithBlockedEffect(data_range=1.0, block_size=block_size)
```

## Related

Same `and`/`or` validation pattern was fixed in retrieval metrics: #3379, #3416, #3418. This is the image-metric counterpart.

---

AI assistance: Developed with Claude. I have reviewed every changed line.


Borda (approved):

Sriniketh24:
Thanks for the approval @Borda! Also pushed a small follow-up to align the error message wording with the convention used elsewhere in the codebase (e.g. `rase.py`, `vif.py`: "is expected to be ..., but got ...") per Copilot's review nit.

Rebased onto the latest `master` merge already in this branch — no conflicts. Let me know if there's anything else needed to get this merged.

PR #3447, opened 2026-08-09
Fix functional nominal metrics for sparse category labels

### LINKED ISSUE #3446: Functional nominal metrics fail for non-zero-based category labels
## Bug description

The functional nominal association metrics infer the number of categories from the count of unique values, but pass the original category IDs directly to the confusion-matrix implementation. As a result, valid categorical labels that are not zero-based and contiguous can raise a reshape error.

Nominal association should be invariant to renaming categories.

## Reproduction

```python
import torch

from torchmetrics.functional.nominal import (
    cramers_v,
    pearsons_contingency_coefficient,
    theils_u,
    tschuprows_t,
)

base = torch.tensor([0, 0, 1, 1])
shifted = base + 1

metrics = [
    (cramers_v, {"bias_correction": False}),
    (pearsons_contingency_coefficient, {}),
    (theils_u, {}),
    (tschuprows_t, {"bias_correction": False}),
]

for metric, kwargs in metrics:
    print(metric(base, base, **kwargs))
    print(metric(shifted, shifted, **kwargs))
```

The first call succeeds for each metric. The relabeled input raises:

```text
RuntimeError: shape '[2, 2]' is invalid for input of size 7
```

The corresponding matrix APIs are affected as well when columns contain sparse or independently labeled categories.

## Expected behavior

The shifted input describes the same two categories and should produce the same association value as the zero-based input. For example, Cramer's V should be `tensor(1.)` in both cases.

## Root cause

The functional paths derive `num_classes` using the number of unique values, while the classification confusion-matrix updater uses the raw label values as bin indices. With labels `1` and `2`, the inferred class count is two, but label `2` creates a bin outside the expected `2 x 2` range.

Normalizing the observed categories to contiguous IDs after logits and NaN preprocessing avoids that mismatch. Stateful metrics that receive an explicit `num_classes` should keep their existing stable encoding behavior across updates.

## Environment

- TorchMetrics: 1.9.0 / current `master` at `37a80809`
- PyTorch: 2.13.0+cpu
- Python: 3.11.9
- OS: Windows

AI disclosure: I used Codex to help identify and reproduce the behavior, search for existing reports, and draft this issue. I reviewed the reproduction and diagnosis.

aswanth-07:
## What does this PR do?

Fixes #3446.

Functional nominal association metrics inferred `num_classes` from the number of unique values but passed raw category IDs to the classification confusion-matrix updater. Sparse or non-zero-based labels could therefore produce bin indices outside the inferred matrix and fail during reshape.

This change centralizes nominal confusion-matrix preprocessing. Functional APIs jointly encode observed categories as contiguous IDs after logits and NaN handling, while stateful metrics retain their explicit `num_classes` behavior so their encoding remains stable across updates.

Regression coverage checks category-renaming invariance for the scalar and matrix forms of:

- Cramer's V
- Pearson's contingency coefficient
- Theil's U
- Tschuprow's T

## Validation

- `python -m pytest tests/unittests/nominal -q`: 112 passed
- `python -m ruff check <six changed files>`
- `python -m ruff format --check <six changed files>`
- `git diff --check`
- Clean-master reproduction confirmed all four scalar functions fail before this patch and return label-invariant values afterward

## Before submitting

- [x] Bug documented and reproduced in #3446
- [x] Contributor guidelines reviewed
- [x] Regression tests added
- [x] Documentation not changed because this restores already-documented categorical input behavior

AI disclosure: I used Codex to help identify and reproduce the bug, search for existing reports, implement and validate the fix, and draft this PR. I reviewed the code, tests, and validation output and will personally handle review feedback.

### DIFF
--- a/src/torchmetrics/functional/nominal/cramers.py
+++ b/src/torchmetrics/functional/nominal/cramers.py
@@ -18,12 +18,11 @@ import torch
 from torch import Tensor
 from typing_extensions import Literal
 
-from torchmetrics.functional.classification.confusion_matrix import _multiclass_confusion_matrix_update
 from torchmetrics.functional.nominal.utils import (
     _compute_bias_corrected_values,
     _compute_chi_squared,
     _drop_empty_rows_and_cols,
-    _handle_nan_in_data,
+    _nominal_confusion_matrix_update,
     _nominal_input_validation,
     _unable_to_use_bias_correction_warning,
 )
@@ -32,7 +31,7 @@ from torchmetrics.functional.nominal.utils import (
 def _cramers_v_update(
     preds: Tensor,
     target: Tensor,
-    num_classes: int,
+    num_classes: Optional[int],
     nan_strategy: Literal["replace", "drop"] = "replace",
     nan_replace_value: Optional[float] = 0.0,
 ) -> Tensor:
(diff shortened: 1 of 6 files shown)


aswanth-07:
Closing this PR as a duplicate of #3461, which is the active implementation for #3460 and covers the same functional nominal-metric category encoding fix. Keeping one upstream implementation avoids splitting maintainer review.

PR #3452, opened 2026-08-11
fix: account for len(thresholds) when dispatching the vectorized PR-curve update

### LINKED ISSUE #3299: MulticlassPrecisionRecallCurve with thresholds=[integer] does not use constant memory on MPS
## 🐛 Bug

According to the [MulticlassPrecisionRecallCurve documentation](https://lightning.ai/docs/torchmetrics/stable/classification/precision_recall_curve.html#multiclassprecisionrecallcurve), setting the `thresholds` argument should cause MulticlassPrecisionRecallCurve to use a constant amount of memory.
When the metric is on MPS, this is not the case.

### To Reproduce

Sample code and steps to reproduce the behavior with expected result...

<details>

```python
# Correct behavior
metric = torchmetrics.classification.MulticlassPrecisionRecallCurve(num_classes=3, ignore_index=-1, thresholds=200).to('cpu')
pred = torch.randn(60000, 3).to('cpu')
target = torch.randint(0, 3, (60000,)).to('cpu')
metric = metric(pred, target)
print(metric)


# Crashes with OOM error
metric = torchmetrics.classification.MulticlassPrecisionRecallCurve(num_classes=3, ignore_index=-1, thresholds=200).to('mps')
pred = torch.randn(60000, 3).to('mps')
target = torch.randint(0, 3, (60000,)).to('mps')
metric = metric(pred, target)
print(metric)
```

</details>

<details>

- TorchMetrics version (if build from source, add commit SHA): 1.8.1
- Python & PyTorch Version (e.g., 1.0): 3.11.7
- Any other relevant information such as OS (e.g., Linux): MacOS Tahoe 26.0.1 on Macbook Pro M4 Max

</details>

### Additional context

Here's the error I get when I run the code on MPS. I can see that it does not use constant memory, because it attempts to allocate different amounts of memory when I vary the input size.

<details>

<summary>Traceback</summary>

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[15], [line 5](vscode-notebook-cell:?execution_count=15&line=5)
      3 target = torch.randint(0, 3, (60000,)).to('mps')
      4 print( pred.shape, target.shape)
----> [5](vscode-notebook-cell:?execution_count=15&line=5) metric = metric(pred, target)
      6 metric

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py:1773, in Module._wrapped_call_impl(self, *args, **kwargs)
   1771     return self._compiled_call_impl(*args, **kwargs)  # type: ignore[misc]
   1772 else:
-> [1773](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py:1773)     return self._call_impl(*args, **kwargs)

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py:1784, in Module._call_impl(self, *args, **kwargs)
   1779 # If we don't have any hooks, we want to skip the rest of the logic in
   1780 # this function, and just call forward.
   1781 if not (self._backward_hooks or self._backward_pre_hooks or self._forward_hooks or self._forward_pre_hooks
   1782         or _global_backward_pre_hooks or _global_backward_hooks
   1783         or _global_forward_hooks or _global_forward_pre_hooks):
-> [1784](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torch/nn/modules/module.py:1784)     return forward_call(*args, **kwargs)
   1786 result = None
   1787 called_always_called_hooks = set()

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:315, in Metric.forward(self, *args, **kwargs)
    313     self._forward_cache = self._forward_full_state_update(*args, **kwargs)
    314 else:
--> [315](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:315)     self._forward_cache = self._forward_reduce_state_update(*args, **kwargs)
    317 return self._forward_cache

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:384, in Metric._forward_reduce_state_update(self, *args, **kwargs)
    381 self._enable_grad = True  # allow grads for batch computation
    383 # calculate batch state and compute batch value
--> [384](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:384) self.update(*args, **kwargs)
    385 batch_val = self.compute()
    387 # reduce batch and global state

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:559, in Metric._wrap_update.<locals>.wrapped_func(*args, **kwargs)
    551         if "Expected all tensors to be on" in str(err):
    552             raise RuntimeError(
    553                 "Encountered different devices in metric calculation (see stacktrace for details)."
    554                 " This could be due to the metric class not being on the same device as input."
   (...)    557                 " device corresponds to the device of the input."
    558             ) from err
--> [559](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:559)         raise err
    561 if self.compute_on_cpu:
    562     self._move_list_states_to_cpu()

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:549, in Metric._wrap_update.<locals>.wrapped_func(*args, **kwargs)
    547 with torch.set_grad_enabled(self._enable_grad):
    548     try:
--> [549](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/metric.py:549)         update(*args, **kwargs)
    550     except RuntimeError as err:
    551         if "Expected all tensors to be on" in str(err):

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/classification/precision_recall_curve.py:379, in MulticlassPrecisionRecallCurve.update(self, preds, target)
    375     _multiclass_precision_recall_curve_tensor_validation(preds, target, self.num_classes, self.ignore_index)
    376 preds, target, _ = _multiclass_precision_recall_curve_format(
    377     preds, target, self.num_classes, self.thresholds, self.ignore_index, self.average
    378 )
--> [379](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/classification/precision_recall_curve.py:379) state = _multiclass_precision_recall_curve_update(
    380     preds, target, self.num_classes, self.thresholds, self.average
    381 )
    382 if isinstance(state, Tensor):
    383     self.confmat += state

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/functional/classification/precision_recall_curve.py:486, in _multiclass_precision_recall_curve_update(preds, target, num_classes, thresholds, average)
    484 else:
    485     update_fn = _multiclass_precision_recall_curve_update_loop
--> [486](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/functional/classification/precision_recall_curve.py:486) return update_fn(preds, target, num_classes, thresholds)

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/functional/classification/precision_recall_curve.py:507, in _multiclass_precision_recall_curve_update_vectorized(preds, target, num_classes, thresholds)
    505 unique_mapping += 4 * torch.arange(num_classes, device=preds.device).unsqueeze(0).unsqueeze(-1)
    506 unique_mapping += 4 * num_classes * torch.arange(len_t, device=preds.device)
--> [507](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/functional/classification/precision_recall_curve.py:507) bins = _bincount(unique_mapping.flatten(), minlength=4 * num_classes * len_t)
    508 return bins.reshape(len_t, num_classes, 2, 2)

File ~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/utilities/data.py:203, in _bincount(x, minlength)
    200     minlength = len(torch.unique(x))
    202 if torch.are_deterministic_algorithms_enabled() or _XLA_AVAILABLE or x.is_mps:
--> [203](https://file+.vscode-resource.vscode-cdn.net/Users/pgq596/Programs/marcos/notebooks/~/Programs/marcos/.venv/lib/python3.11/site-packages/torchmetrics/utilities/data.py:203)     mesh = torch.arange(minlength, device=x.device).repeat(len(x), 1)
    204     return torch.eq(x.reshape(-1, 1), mesh).sum(dim=0)
    206 return torch.bincount(x, minlength=minlength)

RuntimeError: Invalid buffer size: 643.73 GiB
``` 

</details>

Kayvan-Zahiri:
Closes #3299.

### The bug

`thresholds` is documented to bound memory, but the dispatch that chooses between
the vectorized and loop update paths never looks at how many thresholds there are.

```python
if preds.numel() * num_classes <= 1_000_000:
    update_fn = _multiclass_precision_recall_curve_update_vectorized
```

The vectorized path then allocates `preds.numel() * len(thresholds)` elements at

```python
preds_t = (preds.unsqueeze(-1) >= thresholds.unsqueeze(0).unsqueeze(0)).long()
```

and again for `unique_mapping`. For the reported case, `preds` of shape
`(60000, 3)` with `thresholds=200`, the gate computes `540,000 <= 1,000,000` and
picks the vectorized path, which then asks for **288 MB** of int64, twice over:

| thresholds | gate says | intermediate |
| --- | --- | --- |
| 5 | vectorized | 7.2 MB |
| 50 | vectorized | 72.0 MB |
| 200 | vectorized | **288.0 MB** |
| 1000 | vectorized | **1440.0 MB** |

MPS surfaces this as an OOM because its ceiling is lower, but the memory model is
wrong on every device. CPU just absorbs it.

### The fix

Bound the intermediate as well as the input. The binary dispatch has the same
defect (`preds.numel() <= 50_000`, also blind to `len_t`), so both are fixed.

This is a pure narrowing: a case that previously went vectorized either still does,
or falls back to the loop. Nothing that used the loop before is now vectorized, so
no workload gets slower unless it was also allocating hundreds of megabytes.

### Why this is safe

The two paths return identical state. I verified across thresholds of 3, 7, 20 and
200 for multiclass and 3, 20, 200 for binary, on both the vectorized and loop
implementations, with `torch.equal` in every case. Falling back only trades speed
for memory, never correctness.

### Tests

- `test_multiclass_update_memory_does_not_scale_with_thresholds`, parametrized over
  5 and 200 thresholds, asserts which path the dispatch actually picks. The 200 case
  fails on `main` and passes here; the 5 case passes both ways, which pins that small
  threshold counts keep their current fast path.
- `test_multiclass_update_paths_agree` and `test_binary_update_paths_agree` pin the
  invariant the fix depends on.

Full file: **214 passed, 24 skipped, 4 xfailed**, no regressions.

### One question for a reviewer

`_MAX_VECTORIZED_ELEMENTS = 1_000_000` (8 MB of int64, doubled by `unique_mapping`)
is my choice, not something derived from an existing constant. It keeps the reported
case out of the vectorized path with room to spare, but the right budget is really a
maintainer call and I would happily change it. If you would rather express it as a
byte budget, or scale it from the existing `50_000` / `1_000_000` gates, say which
and I will update.

### Not covered here

`_multilabel_precision_recall_curve_update` has the same `preds.numel() * len_t`
allocation and **no** loop fallback to dispatch to, so it is unconditionally exposed.
Fixing it means writing a multilabel loop implementation, which felt like a separate
change from the reported bug. Happy to add it to this PR or open a follow-up,
whichever you prefer.

### DIFF
--- a/src/torchmetrics/functional/classification/precision_recall_curve.py
+++ b/src/torchmetrics/functional/classification/precision_recall_curve.py
@@ -189,6 +189,14 @@ def _binary_precision_recall_curve_format(
     return preds, target, thresholds
 
 
+# The vectorized update paths materialise an intermediate holding
+# ``preds.numel() * len(thresholds)`` elements, so their peak memory grows with the
+# number of thresholds. The input-size checks below cannot see that on their own, so a
+# large ``thresholds`` silently defeats the constant-memory behaviour that argument is
+# documented to provide. Bound the intermediate as well as the input.
+_MAX_VECTORIZED_ELEMENTS = 1_000_000
+
+
 def _binary_precision_recall_curve_update(
     preds: Tensor,
     target: Tensor,
@@ -202,7 +210,7 @@ def _binary_precision_recall_curve_update(
     """
     if thresholds is None:
         return preds, target
-    if preds.numel() <= 50_000:
+    if preds.numel() <= 50_000 and preds.numel() * len(thresholds) <= _MAX_VECTORIZED_ELEMENTS:
         update_fn = _binary_precision_recall_curve_update_vectorized
     else:
         update_fn = _binary_precision_recall_curve_update_loop
@@ -479,7 +487,10 @@ def _multiclass_precision_recall_curve_update(
         return preds, target
     if average == "micro":
         return _binary_precision_recall_curve_update(preds, target, thresholds)
-    if preds.numel() * num_classes <= 1_000_000:
+    if (
+        preds.numel() * num_classes <= 1_000_000
+        and preds.numel() * len(thresholds) <= _MAX_VECTORIZED_ELEMENTS
+    ):
         update_fn = _multiclass_precision_recall_curve_update_vectorized
     else:
         update_fn = _multiclass_precision_recall_curve_update_loop
--- a/src/torchmetrics/functional/classification/precision_recall_curve.py
+++ b/src/torchmetrics/functional/classification/precision_recall_curve.py
@@ -487,10 +487,7 @@ def _multiclass_precision_recall_curve_update(
         return preds, target
     if average == "micro":
         return _binary_precision_recall_curve_update(preds, target, thresholds)
-    if (
-        preds.numel() * num_classes <= 1_000_000
-        and preds.numel() * len(thresholds) <= _MAX_VECTORIZED_ELEMENTS
-    ):
+    if preds.numel() * num_classes <= 1_000_000 and preds.numel() * len(thresholds) <= _MAX_VECTORIZED_ELEMENTS:
         update_fn = _multiclass_precision_recall_curve_update_vectorized
     else:
         update_fn = _multiclass_precision_recall_curve_update_loop
--- a/tests/unittests/classification/test_precision_recall_curve.py
+++ b/tests/unittests/classification/test_precision_recall_curve.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from functools import partial
+from unittest import mock
 
 import numpy as np
 import pytest
@@ -31,6 +32,14 @@ from torchmetrics.functional.classification.precision_recall_curve import (
     binary_precision_recall_curve,
     multiclass_precision_recall_curve,
     multilabel_precision_recall_curve,
+)
+from torchmetrics.functional.classification.precision_recall_curve import (
+    _MAX_VECTORIZED_ELEMENTS,
+    _binary_precision_recall_curve_update_loop,
+    _binary_precision_recall_curve_update_vectorized,
+    _multiclass_precision_recall_curve_update,
+    _multiclass_precision_recall_curve_update_loop,
+    _multiclass_precision_recall_curve_update_vectorized,
 )
 from torchmetrics.metric import Metric
 from torchmetrics.utilities.imports import _TORCH_GREATER_EQUAL_2_1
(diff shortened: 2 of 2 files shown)


Borda (commented):
can we rather solve it with batching?

Kayvan-Zahiri:
Yes. The state is a confusion matrix accumulated over
samples, so summing the chunks equals computing it whole.

I sent the dispatch fix first because the vectorized update allocates
`preds.numel() * len(thresholds)` in one go, while the guards choosing the path
looked only at input size, with no term for the threshold count. So passing
`thresholds`, documented as the way to bound memory, could raise it instead.
Batching then makes the fast path reachable where the guard currently sends it
to the loop. 

Can chunk over thresholds, over samples, or size it from a byte budget?

Kayvan-Zahiri:
@Borda done, it now batches instead of falling back to the loop. The state is a sum over samples, so each chunk runs the existing vectorized update unchanged and the results are added, with chunks sized to stay under the element budget. On a (60000, 3) input with 200 thresholds it is about 3x faster than the loop and peaks at 8 MB instead of 288 MB; numbers are in the description.

PR #2753, opened 2024-09-17
Ignore the `NoUtterancesError` when calculating pesq for a batch

veera-puthiran-14082:
## What does this PR do?

Fixes #2752 

- Ignores the `NoUtterancesError` and create tensors with the available results, if a batch of size 32 is passed and 2 of them encountered an error, then for 30 pesq scores avg will be calculated and returned.

- introduced a simple lambda function which filter out exception messages and keep the pesq values calculated.
- since the np.array contained error messages and pesq scores, the result was in np.object dtype. it is converted to np.float32 since pesq is also calculated in float32

### Checklist

- [ ] Was this **discussed/agreed** via a Github issue? (no need for typos and docs improvements)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/torchmetrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to **update the docs**?
- [ ] Did you write any new **necessary tests**?

<details>
  <summary>PR review</summary>

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

</details>

----
📚 Documentation preview 📚: https://torchmetrics--2753.org.readthedocs.build/en/2753/

PR #3455, opened 2026-08-13
Fix PESQ aborting a whole batch on one unscorable sample

### LINKED ISSUE #3304: PerceptualEvaluationSpeechQuality - NoUtterancesError: b'No utterances detected'
## 🐛 Bug

PerceptualEvaluationSpeechQuality raises for some samples `NoUtterancesError: b'No utterances detected'.

### To Reproduce

Sample code and steps to reproduce the behavior with expected result...

<details>

```python
import torch
from torchmetrics.audio.pesq import PerceptualEvaluationSpeechQuality

target = torch.load("target.pt")
print(f"{target.shape=} {target.dtype=} {target.min()=} {target.max()=} {target.mean()=} {target.std()=}")
pred = torch.load("pred.pt")
print(f"{pred.shape=} {pred.dtype=} {pred.min()=} {pred.max()=} {pred.mean()=} {pred.std()=}")


pesq = PerceptualEvaluationSpeechQuality(fs=8000, mode="nb")
pesq.update(pred, target)
```

</details>

<details>

- TorchMetrics version (if build from source, add commit SHA): 1.7.0
- Python & PyTorch Version (e.g., 1.0): 3.11.10 & 2.6.0+cpu (also CUDA version)
- Any other relevant information such as OS (e.g., Linux): Windows & Linux

</details>

### Additional context

Attached a data sample to reproduce the issue.

[sample_data.zip](https://github.com/user-attachments/files/23396920/sample_data.zip)

Kayvan-Zahiri:
Fixes #3304.

One unscorable sample takes down the whole batch. `pesq` and `pesq_batch` both
default to `on_error=PesqError.RAISE_EXCEPTION` and torchmetrics never passes
`on_error`, so what happens depends on `n_processes`:

- `n_processes=1` (the default): `pesq()` raises `NoUtterancesError` and it
  propagates out of `update()`. The `_filter_error_msg` guard added in #2753 is
  unreachable on this path.
- `n_processes != 1`: `pesq_batch()` catches worker exceptions and returns them
  in the result list, so the guard *is* reached, but it **drops** the failed
  entry. A 3-sample batch silently returns 2 values.

Measured on a 3-sample batch with `target[1]` silent:

```
functional,   n_processes=1  -> raises NoUtterancesError
functional,   n_processes=2  -> shape (2,)  values [2.5199, 2.1842]   <- one sample vanished
class metric, n_processes=2  -> compute()=2.3521  total=2             <- 3 in, 2 counted
```

After:

```
functional,   n_processes=1  -> shape (3,)  values [2.5199, nan, 2.1842]
functional,   n_processes=2  -> shape (3,)  values [2.5199, nan, 2.1842]
class metric, either         -> compute()=2.3521  total=2
```

## Change

Pass `on_error=PesqError.RETURN_VALUES` at all three call sites and map every
failure report onto `nan`, keeping one entry per input sample. Failures arrive
two ways and both are handled: negative error codes (`-1..-7`, which cannot
collide with a real score since valid PESQ is `>= -0.5`) and the exception
objects `pesq_batch` collects from its workers.

The class metric excludes `nan` samples from both the running sum and `total`,
so a degenerate sample does not poison an epoch's average. That matches what the
multiprocessing path already did by dropping them, so it is not a new policy.

Scores for scorable samples are unchanged: `RETURN_VALUES` and
`RAISE_EXCEPTION` return the same value on success (verified,
2.158228635787964 both ways).

## Also in this PR

#2753 replaced `pesq_val.reshape(preds.shape[:-1])` with
`reshape(len(pesq_val))`, so a `(2, 3, 2100)` input returned `(6,)` instead of
`(2, 3)`, contradicting the documented `(...,)` shape. Restored, which is safe
now that no entries are dropped. **This is a behaviour change for `ndim > 2`
inputs** relative to 1.4.x-1.9.0. Happy to split it into its own PR if you would
rather keep this one minimal.

## Tests

Four cases in `tests/unittests/audio/test_pesq.py`, all failing on main:

```
test_degenerate_sample_does_not_abort_batch[1]  NoUtterancesError
test_degenerate_sample_does_not_abort_batch[2]  assert torch.Size([2]) == (3,)
test_degenerate_sample_returns_nan_for_single_sample  NoUtterancesError
test_multidim_input_keeps_shape                 assert torch.Size([6]) == (2, 3)
```

The two different failure modes for `[1]` and `[2]` are the two paths above.

`tests/unittests/audio/test_pesq.py`: 20 -> 24 passed, same 6 skipped and 3
xfailed. Doctests pass. Wider audio dir: 88 passed. ruff, format and mypy clean.

### DIFF
--- a/src/torchmetrics/functional/audio/pesq.py
+++ b/src/torchmetrics/functional/audio/pesq.py
@@ -55,8 +55,15 @@ def perceptual_evaluation_speech_quality(
         n_processes: integer specifying the number of processes to run in parallel for the metric calculation.
             Only applies to batches of data and if ``multiprocessing`` package is installed.
 
+    .. note::
+        Samples that the ``pesq`` backend cannot score, e.g. a silent reference for which no utterance can be
+        detected, are returned as ``nan`` instead of raising an error, so that a single degenerate sample does not
+        abort the calculation for the rest of the batch. The class based metric
+        :class:`~torchmetrics.audio.pesq.PerceptualEvaluationSpeechQuality` excludes such samples from its average.
+
     Returns:
-        Float tensor with shape ``(...,)`` of PESQ values per sample
+        Float tensor with shape ``(...,)`` of PESQ values per sample, with ``nan`` for samples the backend
+        could not score
 
     Raises:
         ModuleNotFoundError:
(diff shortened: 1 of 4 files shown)


ethanwharris (approved):

src/torchmetrics/functional/audio/pesq.py
```diff
@@ -98,21 +126,40 @@ def _issubtype_number(x: Any) -> bool:
...
         target_np = target.reshape(-1, preds.shape[-1]).detach().cpu().numpy()
 
         if _MULTIPROCESSING_AVAILABLE and n_processes != 1:
-            pesq_val_np = pesq_backend.pesq_batch(fs, target_np, preds_np, mode, n_processor=n_processes)
-            pesq_val_np = np.array(pesq_val_np)
+            pesq_val_np = pesq_backend.pesq_batch(
+                fs,
+                target_np,
+                preds_np,
+                mode,
```
Borda: Since it's already multi-line, let's use kwargs

Borda (approved):

PR #3458, opened 2026-08-16
Add MeanAveragePrecision3D for 3D bounding box detection

### LINKED ISSUE #3398: mAP Metric for 3D Bounding Boxes
## 🚀 Feature

I would like to be able to compute mean-average-precision for 3D Bounding boxes to also be able to evaluate detection systems that are capable of processing 3D data, like point-clouds, voxels or other.

### Motivation

Currently whenever I want to compute the mAP with `torchmetrics.detection.mean_ap.MeanAveragePrecision` I am restricted to 2D data, meaning the parameter `num_boxes` is restricted to shape `(num_boxes, 4)`. 

### Pitch

I would recommend to add a separate mAP class that only covers 3D bounding boxes. So to separate between
3D mAP and 2D mAP computation. The reason is that in 3D we have different algorithms to compute IoU and average precision etc.

### Alternatives

- Either extend the already existent  `torchmetrics.detection.mean_ap.MeanAveragePrecision`.
- Or implement a complete new metric class.

### Additional context

Read more about mAP here:
- https://medium.com/@rohitbnsd12/3d-object-detection-evaluation-a88854534aac

OnePunchMonk:
## Summary

Closes #3398.

Adds a `MeanAveragePrecision3D` metric (and its functional counterpart `mean_average_precision_3d`) for computing mAP/mAR on axis-aligned 3D bounding boxes, e.g. for evaluating detectors that operate on point-clouds, voxel grids, or other 3D sensor data.

- `torchmetrics.detection.MeanAveragePrecision3D` / `torchmetrics.functional.detection.mean_average_precision_3d`.
- Supports `box_format="xyzwhd"` (center x, y, z + width, height, depth) and `"xyzxyz"` (corner) formats.
- The existing 2D `MeanAveragePrecision` relies on `pycocotools`/`faster_coco_eval`, which only support 2D boxes and masks, so this is implemented independently: 3D IoU is computed directly on axis-aligned boxes, and average precision follows the same COCO-style 101-point interpolation, with greedy highest-score-first matching per class/IoU-threshold, and per-image `max_detection_thresholds` truncation for mAR — mirroring the existing 2D metric's API and output keys (`map`, `map_50`, `map_75`, `mar_{max_det}`, `map_per_class`, `mar_{max_det}_per_class`, `classes`) as closely as possible.
- No new required dependencies (`torch` only).

### Scope / follow-ups

This first version supports axis-aligned 3D boxes only (no yaw/rotation), which covers the common case for e.g. LiDAR-derived detections but not fully-oriented boxes (KITTI/nuScenes-style rotated boxes). Rotated 3D IoU would need convex-polygon intersection and is left as a natural follow-up if there's interest, potentially as an additional `box_format` option.

## Test plan

- [x] Added `tests/unittests/detection/test_map_3d.py` covering: pairwise 3D IoU correctness, perfect-match AP/AR = 1, non-overlapping AP/AR = 0, box format equivalence, per-class metrics, empty-prediction and empty-ground-truth edge cases, and module/functional parity.
- [x] `ruff check` / `ruff format --check` clean on new files.
- [x] `mypy` clean on new files.
- [x] Doctests in both new modules pass.

## AI Usage

- [x] Claude code was used to draft the implementation and tests. 
- [x] A human submitter ( onepunchmonk) has reviewed the changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
--
+- Added `MeanAveragePrecision3D` and `mean_average_precision_3d` for computing mAP/mAR on axis-aligned 3D bounding boxes ([#3398](https://github.com/Lightning-AI/torchmetrics/issues/3398))
 
 
 ### Changed
--- a/docs/source/detection/mean_average_precision_3d.rst
+++ b/docs/source/detection/mean_average_precision_3d.rst
@@ -1,0 +1,21 @@
+.. customcarditem::
+   :header: Mean-Average-Precision (mAP) 3D
+   :image: https://pl-flash-data.s3.amazonaws.com/assets/thumbnails/object_detection.svg
+   :tags: Detection
+
+.. include:: ../links.rst
+
+###############################
+Mean-Average-Precision (mAP) 3D
+###############################
+
+Module Interface
+________________
+
+.. autoclass:: torchmetrics.detection.mean_ap_3d.MeanAveragePrecision3D
+    :exclude-members: update, compute
+
+Functional Interface
+____________________
+
+.. autofunction:: torchmetrics.functional.detection.mean_average_precision_3d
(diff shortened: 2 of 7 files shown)


OnePunchMonk:
Hi @bhimrazy, could you take a quick first pass on this when you get a chance? Thanks!

OnePunchMonk:
@Borda made fixes.

PR #3462, opened 2026-08-17
Fix nominal metrics for non-zero-based category labels

### LINKED ISSUE #3460: Functional nominal metrics raise RuntimeError when category labels do not start at 0
## 🐛 Bug

Functional nominal metrics (`cramers_v`, `pearsons_contingency_coefficient`, `theils_u`, `tschuprows_t`) raise a `RuntimeError` during confusion matrix calculation when categorical inputs contain non-zero-based or non-contiguous category labels (e.g., labels `[1, 2]` or `[5, 6]`).

### To Reproduce

<details>

```python
import torch
from torchmetrics.functional.nominal import cramers_v

preds = torch.tensor([5, 6, 5, 6, 5, 6, 5, 6])
target = torch.tensor([5, 6, 5, 5, 6, 6, 5, 6])

# Raises RuntimeError: shape '[2, 2]' is invalid for input of size 19
cramers_v(preds, target)
```

</details>

<details>

- TorchMetrics version: 1.9.0 (latest master)
- Python & PyTorch Version: Python 3.10+, PyTorch 2.x
- Any other relevant information such as OS: All platforms (Linux, macOS, Windows)

</details>

### Additional context

The functional API infers `num_classes = len(torch.cat([preds, target]).unique())` but passes the unmapped raw values to `_multiclass_confusion_matrix_update`, which expects values strictly in `[0, num_classes - 1]`. Any values outside this range cause invalid bin indexing and reshape failures. Additionally, `_compute_chi_squared` performs in-place float addition on integer `confmat` tensors, causing type cast issues. Remapping categorical labels to contiguous 0-based indices before updating the confusion matrix resolves this issue.

VenishPaneliya:
## What does this PR do?

Fixes #3460

The functional nominal metrics infer `num_classes` from the number of distinct values but pass the **raw** labels to `_multiclass_confusion_matrix_update`, which requires them to be exactly `0, ..., num_classes - 1`. Any other encoding indexes outside the matrix and raises during binning:

```python
cramers_v(torch.tensor([5, 6, 5, 6]), torch.tensor([5, 6, 5, 5]))
# RuntimeError: shape '[2, 2]' is invalid for input of size 19
```

### Second bug found while reproducing

`_compute_chi_squared` applies the bias correction **in place**:

```python
confmat += direction * torch.minimum(...)
```

On the functional path `confmat` is integer-typed, so the float correction raises regardless of how the labels are encoded. This branch is reached whenever `df == 1`, i.e. for **every 2x2 contingency table**, which means `cramers_v` and `tschuprows_t` fail on binary inputs even with perfectly valid zero-based labels:

```python
cramers_v(torch.tensor([0, 1, 0, 1]), torch.tensor([0, 1, 0, 0]))
# RuntimeError: result type Float can't be cast to the desired output type Long
```

The in-place write also mutated the caller's tensor, which is why the class-based metrics — whose `confmat` state is float — were unaffected but would have had their state silently modified.

## Changes

- Add `_remap_nominal_labels` in `functional/nominal/utils.py`: handles NaNs first, then ranks the observed categories onto contiguous zero-based indices via `torch.bucketize` and returns the resulting `num_classes`. Ranking is by sorted order, so inputs that are already zero-based and contiguous are returned unchanged.
- Use it in `cramers_v`, `pearsons_contingency_coefficient`, `theils_u`, `tschuprows_t` and their four `_matrix` variants (8 call sites).
- Make the bias correction in `_compute_chi_squared` out-of-place.
- Add `tests/unittests/nominal/test_nominal_label_encoding.py`.

### Scope: functional API only

The class-based metrics (`CramersV`, `TschuprowsT`, …) take `num_classes` in `__init__`, allocate a fixed `num_classes x num_classes` state, and accumulate `self.confmat += confmat` across batches. Remapping per batch would map different raw labels onto the same index in different batches and silently corrupt the accumulated matrix, so their existing contract — callers pass zero-based labels — is deliberately left alone. This matches the issue title.

## Tests

New file parametrises all four metrics and their matrix variants across five relabellings (offset, one-based, non-contiguous, negative, permuted), plus float labels, the 2x2 bias-correction regression, non-mutation of the input confmat, and the single-category edge case.

Verified the tests actually catch the bug — against the unpatched source:

```
$ pytest tests/unittests/nominal/test_nominal_label_encoding.py -q
39 failed, 14 passed in 4.58s
```

With the fix:

```
$ pytest tests/unittests/nominal/test_nominal_label_encoding.py -q
53 passed, 2 warnings in 2.86s
```

Full nominal suite, unchanged from the pre-patch baseline of 104 (104 existing + 53 new):

```
$ pytest tests/unittests/nominal -q
157 passed, 298 warnings in 21.59s
```

Doctests, lint and types:

```
$ pytest --doctest-modules src/torchmetrics/functional/nominal -q
11 passed, 1 warning in 5.93s

$ ruff check src/torchmetrics/functional/nominal tests/unittests/nominal
All checks passed!

$ ruff format --check src/torchmetrics/functional/nominal tests/unittests/nominal
14 files already formatted

$ mypy src/torchmetrics/functional/nominal/utils.py src/torchmetrics/functional/nominal/cramers.py
Success: no issues found in 2 source files
```

Values are unchanged for already-valid inputs — the existing 104 tests compare against the `dython` reference implementation and still pass, and the new invariance tests assert that a relabelled input produces exactly the same statistic as its zero-based original.

## Note

`theils_u` is the only one of the four that never calls `_nominal_input_validation`, so an invalid `nan_strategy` passes through it silently. Left alone as out of scope — happy to fix it here or in a follow-up if you'd like.

## Before submitting

- [x] Was this **discussed/agreed** via a GitHub issue? (#3460)
- [x] Did you make sure to **update the docs**? (docstring for the new helper)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed malformed LaTeX in `CLIPScore` and `HausdorffDistance` docstring math so it renders correctly ([#3427](https://github.com/Lightning-AI/torchmetrics/pull/3427))
 
 
+- Fixed functional nominal metrics raising when category labels are not contiguous zero-based integers ([#3462](https://github.com/Lightning-AI/torchmetrics/pull/3462))
+
+
+- Fixed `cramers_v` and `tschuprows_t` raising on 2x2 inputs by no longer applying the bias correction in place ([#3462](https://github.com/Lightning-AI/torchmetrics/pull/3462))
+
+
 ---
 
 ## [1.9.0] - 2026-03-05
--- a/src/torchmetrics/functional/nominal/cramers.py
+++ b/src/torchmetrics/functional/nominal/cramers.py
@@ -25,6 +25,7 @@ from torchmetrics.functional.nominal.utils import (
     _drop_empty_rows_and_cols,
     _handle_nan_in_data,
     _nominal_input_validation,
+    _remap_nominal_labels,
     _unable_to_use_bias_correction_warning,
 )
 
@@ -133,7 +134,7 @@ def cramers_v(
 
     """
     _nominal_input_validation(nan_strategy, nan_replace_value)
-    num_classes = len(torch.cat([preds, target]).unique())
+    preds, target, num_classes = _remap_nominal_labels(preds, target, nan_strategy, nan_replace_value)
     confmat = _cramers_v_update(preds, target, num_classes, nan_strategy, nan_replace_value)
     return _cramers_v_compute(confmat, bias_correction)
 
@@ -177,7 +178,7 @@ def cramers_v_matrix(
     cramers_v_matrix_value = torch.ones(num_variables, num_variables, device=matrix.device)
     for i, j in itertools.combinations(range(num_variables), 2):
         x, y = matrix[:, i], matrix[:, j]
-        num_classes = len(torch.cat([x, y]).unique())
+        x, y, num_classes = _remap_nominal_labels(x, y, nan_strategy, nan_replace_value)
         confmat = _cramers_v_update(x, y, num_classes, nan_strategy, nan_replace_value)
         cramers_v_matrix_value[i, j] = cramers_v_matrix_value[j, i] = _cramers_v_compute(confmat, bias_correction)
     return cramers_v_matrix_value
--- a/src/torchmetrics/functional/nominal/pearson.py
+++ b/src/torchmetrics/functional/nominal/pearson.py
@@ -24,6 +24,7 @@ from torchmetrics.functional.nominal.utils import (
     _drop_empty_rows_and_cols,
     _handle_nan_in_data,
     _nominal_input_validation,
+    _remap_nominal_labels,
 )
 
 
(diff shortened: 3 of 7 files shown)


VenishPaneliya:
Apologies — I opened this without first checking the open PR list, and I see now that #3447, #3454 and #3461 already address the same issue (and #3446 looks like a duplicate report of #3460). #3447 predates mine by over a week. That's my mistake, and I don't want to add review load for something already covered.

For whoever triages this, the differences I can see, so you can pick one and close the rest:

- #3461 and this PR both also make the bias correction in `_compute_chi_squared` out-of-place. That part matters independently of the label encoding: the branch fires for every 2x2 table, so `cramers_v` and `tschuprows_t` raise `RuntimeError: result type Float can't be cast to the desired output type Long` on any binary input, even with valid zero-based labels. #3454 and #3447 don't appear to touch it.
- This PR's tests parametrise all four metrics **and** their four `_matrix` variants across five relabellings (offset, one-based, non-contiguous, negative, and a non-monotonic permutation), plus float labels, the 2x2 regression, non-mutation of the caller's confmat, and the single-category edge case. I verified they fail on unpatched `master` (39 failed) and pass with the fix (53 passed), so they pin the behaviour rather than just passing alongside it.
- It deliberately leaves the class-based metrics alone: they accumulate a fixed-size `confmat` across batches, so per-batch remapping would map different raw labels to the same index in different batches and corrupt the running total.

Happy to close this in favour of #3447 as the earliest, and I'd be glad to contribute the extra test cases to whichever PR you prefer instead. Just say which and I'll do that.

VenishPaneliya:
Closing this — #3447 and #3461 both predate it and cover the same fix for #3460, so there is no reason for a third PR to sit in the queue. Sorry for the noise.

Issue #3457, opened 2026-08-15
Import time: `scipy.signal`, `torchvision` and `matplotlib` are imported eagerly by `import torchmetrics`


frazane:
## 🚀 Feature

Don't import optional dependencies (`scipy.signal`, `torchvision`, `matplotlib`) at `import torchmetrics` time just because they happen to be installed.

### Motivation

`python -X importtime -c "import torchmetrics"` on 1.9.0 (torch 2.10, Python 3.12, warm cache), cumulative:

```
torchmetrics                   3.30s
  torch                        1.22s
  scipy.signal                 0.66s
  torchvision                  0.85s
  torchmetrics.utilities.plot  0.31s
```

Around 1.8 s of the 2.1 s that is not torch goes to three optional packages that nothing in the import path uses. Lightning pays it too, since `lightning_fabric/utilities/imports.py` calls `compare_version("torchmetrics", ...)` and that does `importlib.import_module`.

Sources, verified on `master` and 1.9.0:

`scipy.signal` comes from the SRMRpy `hamming` back-compat patch, which sits in `torchmetrics/__init__.py`, `torchmetrics/audio/__init__.py` and `torchmetrics/functional/audio/__init__.py`. All three run eagerly, so all three have to move. `srmrpy` appears only in `requirements/audio_test.txt` and `tests/unittests/audio/test_srmr.py`; the library never imports it.

`torchvision` comes from `functional/image/arniqa.py`, which imports `transforms` and `models.resnet50` at module level under `if _TORCHVISION_AVAILABLE:`. Already tracked in #3314, with open PR #3432; mentioned here only for completeness.

`matplotlib` comes from `utilities/plot.py`, which imports `matplotlib`, `matplotlib.axes` and `matplotlib.pyplot` under `if _MATPLOTLIB_AVAILABLE:`. Around 140 metric modules import it for `_AX_TYPE`, `_PLOT_OUT_TYPE`, `_CMAP_TYPE` and `style_change`.

### Pitch

Move the imports to first use, as `functional/audio/srmr.py` and `functional/image/lpips.py` already do:

- Drop the `scipy.signal` patch from the three `__init__.py` files; apply it in the audio tests or in the `Lightning-Sandbox/SRMRpy` fork if it is still needed.
- `utilities/plot.py`: make the type aliases `TYPE_CHECKING`-only and resolve `plt.style.context` inside `style_change`.
- `arniqa`: #3432.

### Alternatives

[...]

PR #3432, opened 2026-07-22
Fix eager torchvision import causing circular import

youdie006:
## What does this PR do?

Fixes #3314

`arniqa` and `dists` imported `torchvision` at module level (inside the `_TORCHVISION_AVAILABLE` guard), which loaded torchvision as soon as `torchmetrics` was imported and could trigger the `partially initialized module 'torchvision'` circular import in the issue traceback. This moves those imports into the metric bodies (lazy import), matching the pattern already used by `lpips` and the detection functionals, so torchvision loads only when a metric is instantiated.

Behavior is unchanged when torchvision is installed, and the missing-torchvision `ModuleNotFoundError` still fires (the `_TORCHVISION_AVAILABLE` guards are untouched).

Verified: after the change, `import torchmetrics` and the affected modules no longer pull torchvision into `sys.modules`; a real `dists` compute still produces valid output; `pytest tests/unittests/image/test_arniqa.py test_dists.py` gives 11 passed / 10 skipped (including 2 new regression tests asserting no eager torchvision load); `ruff` clean.

<details>
  <summary>Before submitting</summary>

- [x] Was this **discussed/agreed** via a Github issue? — yes, #3314 (maintainer-authored, help wanted)
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/torchmetrics/blob/master/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to **update the docs**? — N/A, internal import-timing fix, no docs/API change
- [x] Did you write any new **necessary tests**? — a regression test per module

</details>

## Did you have fun?

Yes.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). I verified the lazy-import fix (torchvision no longer eagerly loaded), that compute and the missing-dependency error path are unchanged, and that the image tests pass.*

PR #3463, opened 2026-08-17
Do not import scipy.signal and matplotlib at package import time

VenishPaneliya:
## What does this PR do?

Refs #3457 — the `scipy.signal` and `matplotlib` halves. Leaves the `torchvision`/arniqa half to #3432 to avoid conflicting.

`import torchmetrics` pulled in two optional stacks that nothing on the import path uses. This sits on the critical path for downstream libraries — Lightning imports torchmetrics only to compare versions — so everyone paid for it, including users who never plot or touch audio metrics.

### `scipy.signal`

Came from the SRMRpy `hamming` back-compat patch, duplicated across three eagerly executed `__init__.py` files. `srmrpy` is only ever imported by `tests/unittests/audio/test_srmr.py` — the library reimplements SRMR in pure PyTorch and never imports it — so the patch moves into that test module.

### `matplotlib`

Came from `utilities/plot.py`, which imported it at module level purely to build the `_AX_TYPE` / `_PLOT_OUT_TYPE` / `_CMAP_TYPE` annotation aliases and to alias `style_change = plt.style.context`. The aliases now resolve to the real classes under `TYPE_CHECKING` and to `object` at runtime, and `style_change` imports `pyplot` on first use.

I kept runtime `object` fallbacks rather than making the aliases `TYPE_CHECKING`-only, because that would require `from __future__ import annotations` in the ~140 modules that annotate with them. This way the diff stays inside `plot.py` (plus the two call sites below).

### The subtle part

Once the aliases stop being real classes at runtime, `isinstance(x, _AX_TYPE)` becomes `isinstance(x, object)` — **always true**. Four places relied on that check:

| Location | What silently broke |
|---|---|
| `plot.py` `trim_axs` | returned the untrimmed axes array instead of flattening and trimming |
| `collections.py` ×2 | `MetricCollection.plot` argument validation stopped raising |
| `wrappers/multitask.py` | `MultitaskWrapper.plot` argument validation stopped raising |

The `trim_axs` one was caught by `test_confusion_matrix_plotter[multilabel confusion matrix]` failing with `AttributeError: 'numpy.ndarray' object has no attribute 'set_title'`. All four now go through a new `_is_axes` helper that resolves the actual class, and there is a regression test pinning the behaviour.

## Measurements

`import torchmetrics`, median of 5 cold subprocesses, Python 3.13 / Windows:

**With only scipy and matplotlib installed** — the configuration this PR targets:

| | master | this PR |
|---|---|---|
| wall time | 5559 ms | **3571 ms** (−36%) |
| `sys.modules` | 2158 | **1371** (−787) |
| heavy modules loaded | `scipy.signal`, `matplotlib.pyplot` | none |

**With `transformers` and `torchvision` also installed**, where other eager imports dominate:

| | master | this PR |
|---|---|---|
| wall time | 15686 ms | 14426 ms (−8%) |
| `sys.modules` | 3947 | 3776 (−171) |
| heavy modules loaded | `scipy.signal`, `matplotlib.pyplot`, `torchvision` | `torchvision` |

## Tests

New `tests/unittests/utilities/test_lazy_imports.py` checks in a subprocess that a bare `import torchmetrics` leaves `scipy.signal` and `matplotlib` out of `sys.modules`, that the aliases still resolve at runtime, that `style_change` works as both a context manager and a repeatable decorator, and that `_is_axes` discriminates where the alias cannot.

```
$ MPLBACKEND=Agg pytest tests/unittests/utilities -q
6 failed, 371 passed, 8 skipped in 171.62s
```

All 6 failures are missing optional packages locally (`pesq`, `gammatone`, `pycocotools`) and are identical on unpatched `master`:

```
$ git stash && MPLBACKEND=Agg pytest tests/unittests/utilities -q
9 failed, 368 passed, 8 skipped in 110.95s
```

— the 3 extra being this PR's own new tests, which correctly fail without the fix.

Lint and types, compared against `master` for the same files:

```
$ ruff check <changed files>
3 errors  (RUF059 ×3 — all pre-existing, identical on master)

$ ruff format --check <changed files>
5 files already formatted

$ mypy src/torchmetrics/utilities/plot.py src/torchmetrics/collections.py src/torchmetrics/wrappers/multitask.py
Found 11 errors   (master: 16 — the TYPE_CHECKING types let mypy resolve more, net -5)
```

## One thing found along the way, not fixed here

When `transformers` is installed, `import torchmetrics` still pulls in `scipy` — but via a different path than this issue describes:

```
torchmetrics/__init__.py
  → functional/__init__.py:129
    → functional/text/__init__.py:50
      → functional/text/bert.py:56  →  from transformers import AutoModel, AutoTokenizer
```

`bert.py` imports `transformers` at module level, which pulls in `scipy.sparse` (and a lot more). That is a separate eager import from the SRMRpy one, so I left it alone and scoped the regression test to `scipy.signal` accordingly — asserting on the `scipy` root would make the test fail for an unrelated reason. Happy to open a follow-up issue if that is worth tracking.

## Before submitting

- [x] Was this **discussed/agreed** via a GitHub issue? (#3457)
- [x] Did you make sure to **update the docs**? (docstrings on the new helpers)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
--
+- Stopped importing `scipy.signal` and `matplotlib` at `import torchmetrics` time, resolving both on first use instead ([#3463](https://github.com/Lightning-AI/torchmetrics/pull/3463))
 
 
 ### Deprecated
--- a/src/torchmetrics/__init__.py
+++ b/src/torchmetrics/__init__.py
@@ -27,13 +27,6 @@ if package_available("PIL"):
     if not hasattr(PIL, "PILLOW_VERSION"):
         PIL.PILLOW_VERSION = PIL.__version__
 
-if package_available("scipy"):
-    import scipy.signal
-
-    # back compatibility patch due to SMRMpy using scipy.signal.hamming
-    if not hasattr(scipy.signal, "hamming"):
-        scipy.signal.hamming = scipy.signal.windows.hamming
-
 from torchmetrics import functional  # noqa: E402
 from torchmetrics.aggregation import (  # noqa: E402
     CatMetric,
--- a/src/torchmetrics/audio/__init__.py
+++ b/src/torchmetrics/audio/__init__.py
@@ -29,16 +29,8 @@ from torchmetrics.utilities.imports import (
     _PESQ_AVAILABLE,
     _PYSTOI_AVAILABLE,
     _REQUESTS_AVAILABLE,
-    _SCIPI_AVAILABLE,
     _TORCHAUDIO_AVAILABLE,
 )
-
-if _SCIPI_AVAILABLE:
-    import scipy.signal
-
-    # back compatibility patch due to SMRMpy using scipy.signal.hamming
-    if not hasattr(scipy.signal, "hamming"):
-        scipy.signal.hamming = scipy.signal.windows.hamming
 
 __all__ = [
     "ComplexScaleInvariantSignalNoiseRatio",
--- a/src/torchmetrics/collections.py
+++ b/src/torchmetrics/collections.py
@@ -26,7 +26,7 @@ from torchmetrics.metric import Metric
 from torchmetrics.utilities import rank_zero_warn
 from torchmetrics.utilities.data import _flatten, _flatten_dict, allclose
 from torchmetrics.utilities.imports import _MATPLOTLIB_AVAILABLE
-from torchmetrics.utilities.plot import _AX_TYPE, _PLOT_OUT_TYPE, plot_single_or_multi_val
+from torchmetrics.utilities.plot import _AX_TYPE, _PLOT_OUT_TYPE, _is_axes, plot_single_or_multi_val
 
 if not _MATPLOTLIB_AVAILABLE:
     __doctest_skip__ = ["MetricCollection.plot", "MetricCollection.plot_all"]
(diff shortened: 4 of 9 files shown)


VenishPaneliya:
Apologies — I should have checked the open PR list before opening this. #3459 already covers the `scipy.signal` half of #3457, and #3432 covers the `torchvision` half. That's my mistake.

Flagging the overlap so you can decide rather than review both:

- The `scipy.signal` change here is the same idea as #3459 — drop the SRMRpy `hamming` patch from the three eager `__init__.py` files and move it into `tests/unittests/audio/test_srmr.py`, which is the only place `srmrpy` is ever imported.
- What isn't covered elsewhere is the **matplotlib** half: `utilities/plot.py` imported matplotlib at module level purely to build the `_AX_TYPE` / `_PLOT_OUT_TYPE` / `_CMAP_TYPE` aliases and to alias `plt.style.context`. That's the larger share of the win here (5559 ms -> 3571 ms, 2158 -> 1371 modules, measured with only scipy and matplotlib installed).
- One thing worth keeping regardless of which PR you take: once those aliases stop being real classes at runtime, `isinstance(x, _AX_TYPE)` becomes `isinstance(x, object)` and is **always true**. Four call sites rely on that check — `trim_axs` in `plot.py`, two in `MetricCollection.plot`, one in `MultitaskWrapper.plot` — so they silently stop trimming and stop validating, with no error raised. `test_confusion_matrix_plotter[multilabel confusion matrix]` catches it. This PR routes them through a small `_is_axes` helper and adds a regression test.

If you'd prefer, I'm happy to close this and rebase just the matplotlib half on top of #3459 once that lands, so the two don't conflict. Let me know which you'd rather have.

VenishPaneliya:
Closing this — #3459 covers the `scipy.signal` half of #3457 and #3432 covers the `torchvision` half, both ahead of this one. Sorry for the noise.

PR #3466, opened 2026-08-18
Preserve input dtype in spearman_corrcoef and kendall_rank_corrcoef

### LINKED ISSUE #3465: spearman_corrcoef and kendall_rank_corrcoef silently downcast float64 to float32
## 🐛 Bug

`spearman_corrcoef` and `kendall_rank_corrcoef` silently return `float32` for `float64` inputs. Every other regression metric preserves the input dtype, and `kendall_rank_corrcoef` itself preserves it for `variant="c"` but not for `"a"` or `"b"`, so this looks unintended rather than deliberate.

### To Reproduce

```python
import torch
from torchmetrics.functional import pearson_corrcoef, spearman_corrcoef, kendall_rank_corrcoef

preds = torch.rand(100, dtype=torch.float64)
target = torch.rand(100, dtype=torch.float64)

print(pearson_corrcoef(preds, target).dtype)       # torch.float64
print(spearman_corrcoef(preds, target).dtype)      # torch.float32  <-- 
print(kendall_rank_corrcoef(preds, target).dtype)  # torch.float32  <-- 
```

The class-based `SpearmanCorrCoef` and `KendallRankCorrCoef` return `float32` as well.

`kendall_rank_corrcoef` is inconsistent across its own variants:

| `variant` | output dtype for `float64` input |
|---|---|
| `"a"` | `float32` |
| `"b"` (default) | `float32` |
| `"c"` | `float64` |

Surveying the functional regression metrics with `float64` inputs, these two are the only ones that downcast:

| metric | float64 in -> out |
|---|---|
| `mean_absolute_error`, `mean_squared_error`, `mean_absolute_percentage_error`, `symmetric_mean_absolute_percentage_error`, `mean_squared_log_error`, `weighted_mean_absolute_percentage_error`, `r2_score`, `explained_variance`, `pearson_corrcoef`, `concordance_corrcoef`, `log_cosh_error`, `relative_squared_error`, `tweedie_deviance_score`, `minkowski_distance`, `normalized_root_mean_squared_error` | `float64` |
| **`spearman_corrcoef`** | **`float32`** |
| **`kendall_rank_corrcoef`** | **`float32`** |

### Why it matters beyond the dtype label

Spearman is rank-based, so two identical strictly-increasing vectors must give exactly `1.0` regardless of the spacing between values. When the differences are below `float32` resolution the result drifts:

```python
import torch
from torchmetrics.functional import spearman_corrcoef, pearson_corrcoef

a = torch.tensor([1.0, 1.0 + 1e-10, 1.0 + 2e-10, 1.0 + 3e-10], dtype=torch.float64)

print(spearman_corrcoef(a, a).item())  # 0.9999992251396179
print(pearson_corrcoef(a, a).item())   # 0.9999999999999999
```

Users who deliberately opt into `float64` — common when correlating values with a small dynamic range — silently lose that precision.

### Cause

Both come from an integer intermediate being divided, which falls back to the **global default dtype** (`float32`) rather than the input dtype.

**`spearman_corrcoef`** — `_rank_data` in `functional/regression/spearman.py`:

```python
sum_ranks = torch.zeros_like(uniq, dtype=torch.int32)
sum_ranks.scatter_add_(0, inv, rank.to(torch.int32))
mean_ranks = sum_ranks / counts   # int32 / int64 -> float32, regardless of data.dtype
return mean_ranks[inv]
```

The ranks therefore come back `float32` and everything downstream inherits it.

**`kendall_rank_corrcoef`** — `_get_metric_metadata` in `functional/regression/kendall.py`:

```python
n_total = torch.tensor(preds.shape[0], device=preds.device)   # int64, no dtype from preds
```

The concordant/discordant/tie counts are `int64` too, so in `_calculate_tau` the divisions for variants `"a"` and `"b"` produce `float32`. Variant `"c"` escapes it only because it builds `preds_unique` with `dtype=preds.dtype`, which promotes the expression.

### Expected behavior

The output dtype should follow the input, as the other regression metrics do — `float64` in, `float64` out. For `float16`/`bfloat16` inputs, promoting the rank/count arithmetic to `float32` is still appropriate, since ranks are integers that need more mantissa than `float16` offers for larger inputs.

### Environment

- TorchMetrics: `master` (commit 37a80809)
- PyTorch: 2.13.0+cpu, Python 3.13, Windows

I have a fix and regression tests ready and am happy to open a PR if this is welcome.

VenishPaneliya:
## What does this PR do?

Fixes #3465

`spearman_corrcoef` and `kendall_rank_corrcoef` returned `float32` for `float64` inputs. Every other functional regression metric preserves the input dtype, and `kendall_rank_corrcoef` preserved it for `variant="c"` but not for `"a"` or `"b"` — so the metric disagreed with itself depending on which variant you picked.

```python
preds = torch.rand(100, dtype=torch.float64)
target = torch.rand(100, dtype=torch.float64)

pearson_corrcoef(preds, target).dtype       # torch.float64
spearman_corrcoef(preds, target).dtype      # torch.float32  <- before
kendall_rank_corrcoef(preds, target).dtype  # torch.float32  <- before
```

## Cause

Both metrics build their statistics from **integer** intermediates, and dividing two integer tensors in PyTorch yields the *global default* dtype rather than anything derived from the input.

`_rank_data` in `functional/regression/spearman.py`:

```python
sum_ranks = torch.zeros_like(uniq, dtype=torch.int32)
mean_ranks = sum_ranks / counts   # int32 / int64 -> float32, whatever data.dtype was
```

`_calculate_tau` in `functional/regression/kendall.py` divides the `int64` concordant/discordant/tie counts. Variant `"c"` escaped it only because it builds `preds_unique` with `dtype=preds.dtype`, which promoted the surrounding expression.

## Fix

Both now compute in `torch.promote_types(input.dtype, torch.float32)`:

| input | output |
|---|---|
| `float64` | `float64` (was `float32`) |
| `float32` | `float32` |
| `float16` / `bfloat16` | `float32` |

`float16`/`bfloat16` are deliberately **promoted rather than preserved** — ranks and pair counts are integers, and `float16` only represents integers exactly up to 2048, so narrowing the accumulation to the input dtype would be worse than the bug. This also matches the existing behaviour for those dtypes, so nothing changes for them.

## Scope — two adjacent things deliberately left alone

**1. The `eps` bias in `_spearman_corrcoef_compute`.** `corrcoef = cov / (preds_std * target_std + eps)` with `eps=1e-6` biases every result away from an exact ±1. On exact integer ranks, where dtype plays no part:

```
_spearman_corrcoef_compute(a, a)             # eps=1e-6 -> 0.9999992000006399
_spearman_corrcoef_compute(a, a, eps=0.0)    #          -> 0.9999999999999998
```

That is independent of this bug and changing it would move results for every existing user, so it is a call for the maintainers. I noted it in #3465; happy to open a separate issue if you want it tracked.

**2. `SpearmanCorrCoef.update` casts explicitly:**

```python
self.preds.append(preds.to(self.dtype))   # module dtype, float32 by default
```

`KendallRankCorrCoef.update` has no such cast, which is why the class-based Kendall now returns `float64` while the class-based Spearman still returns `float32`. That cast may well be deliberate — halving the stored footprint is a reasonable motive given the metric already warns about buffering every prediction — so I have not touched it. Say the word and I'll align it.

## Tests

New `tests/unittests/regression/test_corrcoef_dtype.py` covers Spearman, all three Kendall variants, `_rank_data` directly, the multioutput path, the tie-averaging branch (the exact division that downcast), and asserts the three correlation metrics agree with each other on what a `float64` input deserves. It also pins that widening does not move the value for `float32` inputs.

Verified the tests actually catch the bug — against unpatched `src/`:

```
$ pytest tests/unittests/regression/test_corrcoef_dtype.py -q
9 failed, 21 passed in 6.44s
```

With the fix:

```
$ pytest tests/unittests/regression/test_corrcoef_dtype.py -q
30 passed in 6.66s
```

Full regression suite, compared against `master` in the same environment:

```
master : 27 failed, 620 passed, 155 skipped, 24 xfailed
this PR: 18 failed, 629 passed, 155 skipped, 24 xfailed
```

The 9-test difference is exactly this PR's own tests. The remaining 18 are pre-existing `nrmse` mismatches against the `permetrics` reference in `test_mean_error.py`, identical on `master` and untouched here. No failure in `test_spearman.py`, `test_kendall.py`, `test_pearson.py`, or the new file:

```
$ pytest tests/unittests/regression/test_spearman.py tests/unittests/regression/test_kendall.py \
         tests/unittests/regression/test_pearson.py tests/unittests/regression/test_corrcoef_dtype.py -q
151 passed, 9 skipped in 17.07s
```

Values cross-checked against SciPy on `float32`, with and without ties:

```
spearman  tm=-0.0387489684  scipy=-0.0387489687
kendall   tm=-0.0272361804  scipy=-0.0272361809
ties: spearman  tm=0.0057996665  scipy=0.0057996669
ties: kendall   tm=0.0045146765  scipy=0.0045146766
```

Lint, types and doctests:

```
$ ruff check <changed files>        All checks passed!
$ ruff format --check <changed>     3 files already formatted
$ mypy <changed source files>       Success: no issues found in 2 source files
$ pytest --doctest-modules <changed source files>   3 passed
```

## Before submitting

- [x] Was this **discussed/agreed** via a GitHub issue? (#3465)
- [x] Did you make sure to **update the docs**? (n/a — no public API or signature change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed `Metric` ignoring an active `torch.device` context manager on torch 2.3-2.7 ([#3448](https://github.com/Lightning-AI/torchmetrics/pull/3448))
 
 
+- Fixed `spearman_corrcoef` and `kendall_rank_corrcoef` silently returning `float32` for `float64` inputs ([#3466](https://github.com/Lightning-AI/torchmetrics/pull/3466))
+
+
 ---
 
 ## [1.9.0] - 2026-03-05
--- a/src/torchmetrics/functional/regression/kendall.py
+++ b/src/torchmetrics/functional/regression/kendall.py
@@ -71,8 +71,7 @@ def _count_concordant_pairs(preds: Tensor, target: Tensor) -> Tensor:
 def _discordant_element_sum(x: Tensor, y: Tensor, i: int) -> Tensor:
     """Count a total number of discordant pairs in a single sequences."""
     return (
-        torch
-        .logical_or(
+        torch.logical_or(
             torch.logical_and(x[i] > x[(i + 1) :], y[i] < y[(i + 1) :]),
             torch.logical_and(x[i] < x[(i + 1) :], y[i] > y[(i + 1) :]),
         )
--- a/src/torchmetrics/functional/regression/kendall.py
+++ b/src/torchmetrics/functional/regression/kendall.py
@@ -71,7 +71,8 @@ def _count_concordant_pairs(preds: Tensor, target: Tensor) -> Tensor:
 def _discordant_element_sum(x: Tensor, y: Tensor, i: int) -> Tensor:
     """Count a total number of discordant pairs in a single sequences."""
     return (
-        torch.logical_or(
+        torch
+        .logical_or(
             torch.logical_and(x[i] > x[(i + 1) :], y[i] < y[(i + 1) :]),
             torch.logical_and(x[i] < x[(i + 1) :], y[i] > y[(i + 1) :]),
         )
@@ -161,21 +162,24 @@ def _calculate_tau(
     variant: _MetricVariant,
 ) -> Tensor:
     """Calculate Kendall's tau from metric metadata."""
+    # the pair and tie counts are integer tensors, so dividing them would fall back to the global default dtype and
+    # silently downcast a float64 input; float16/bfloat16 are promoted to float32 for the accumulation
+    dtype = torch.promote_types(preds.dtype, torch.float32)
+    con_min_dis_pairs = con_min_dis_pairs.to(dtype)
+
     if variant == _MetricVariant.A:
-        return con_min_dis_pairs / (concordant_pairs + discordant_pairs)
+        return con_min_dis_pairs / (concordant_pairs + discordant_pairs).to(dtype)
     if variant == _MetricVariant.B:
-        total_combinations: Tensor = n_total * (n_total - 1) // 2
-        if preds_ties is None:
-            preds_ties = torch.tensor(0.0, dtype=total_combinations.dtype, device=total_combinations.device)
-        if target_ties is None:
-            target_ties = torch.tensor(0.0, dtype=total_combinations.dtype, device=total_combinations.device)
+        total_combinations: Tensor = (n_total * (n_total - 1) // 2).to(dtype)
+        preds_ties = total_combinations.new_zeros(()) if preds_ties is None else preds_ties.to(dtype)
+        target_ties = total_combinations.new_zeros(()) if target_ties is None else target_ties.to(dtype)
         denominator = (total_combinations - preds_ties) * (total_combinations - target_ties)
         return con_min_dis_pairs / torch.sqrt(denominator)
 
-    preds_unique = torch.tensor([len(p.unique()) for p in preds.T], dtype=preds.dtype, device=preds.device)
-    target_unique = torch.tensor([len(t.unique()) for t in target.T], dtype=target.dtype, device=target.device)
+    preds_unique = torch.tensor([len(p.unique()) for p in preds.T], dtype=dtype, device=preds.device)
+    target_unique = torch.tensor([len(t.unique()) for t in target.T], dtype=dtype, device=target.device)
     min_classes = torch.minimum(preds_unique, target_unique)
-    return 2 * con_min_dis_pairs / ((min_classes - 1) / min_classes * n_total**2)
+    return 2 * con_min_dis_pairs / ((min_classes - 1) / min_classes * n_total.to(dtype) ** 2)
 
 
 def _get_p_value_for_t_value_from_dist(t_value: Tensor) -> Tensor:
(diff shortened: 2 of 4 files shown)


VenishPaneliya:
CI fix pushed — the torch 2.0 job was failing on my new tests, not on the change itself.

The new parametrisation exercises `float16`/`bfloat16` on CPU, which torch below 2.1 cannot do for these metrics: `_rank_data` builds its ranks with `torch.arange`, and that has no half support on CPU before 2.1. The existing `test_spearman_corrcoef_half_cpu` already carries a `_TORCH_GREATER_EQUAL_2_1` guard for exactly this reason, with the comment *"Spearman half + cpu does not work due to missing support in torch.arange"* — I'd simply missed it.

Reused that same guard on the two half-precision parameters rather than dropping the coverage on newer torch, so `float16`/`bfloat16` are still asserted everywhere they can run. `float32` and `float64` — which is what the bug is actually about — stay covered on every version.

Also worth flagging: my local `ruff format` reflowed `_discordant_element_sum`, a function this PR doesn't otherwise touch. pre-commit.ci reverted it in dc0c11d, so the diff is back to only the intended changes. My mistake for running a different ruff version than the pinned hook.

PR #3468, opened 2026-08-19
Fix `davies_bouldin_score` returning a wrong value for off-origin float64 data

### LINKED ISSUE #3467: davies_bouldin_score returns a wrong value for off-origin float64 data (float32 buffers)
## 🐛 Bug

`davies_bouldin_score` and `calinski_harabasz_score` allocate their accumulators without a `dtype`, so the buffers are `float32` regardless of the input. `dunn_index` — same family, same input signature — preserves the input dtype, so the clustering metrics disagree with each other.

For `davies_bouldin_score` this is not only a dtype-label problem. The `float32` centroid buffer causes catastrophic cancellation for data that is not centred near zero, and the returned score becomes meaningless.

### To Reproduce

```python
import torch
from torchmetrics.functional.clustering import calinski_harabasz_score, davies_bouldin_score, dunn_index

data = torch.rand(300, 3, dtype=torch.float64)
labels = torch.arange(300) % 4

print(calinski_harabasz_score(data, labels).dtype)  # torch.float32  <-- 
print(davies_bouldin_score(data, labels).dtype)     # torch.float32  <-- 
print(dunn_index(data, labels).dtype)               # torch.float64
```

### The serious part: `davies_bouldin_score` on off-origin data

Same `float64` data, shifted away from the origin, compared against the identical algorithm run with `float64` buffers:

```python
data = torch.rand(300, 3, dtype=torch.float64) + offset
labels = torch.arange(300) % 4
```

| `offset` | `davies_bouldin_score` | `float64` reference | relative error |
|---|---|---|---|
| `0` | 27.0118026733 | 27.0118054968 | 1.0e-07 |
| `1e3` | 27.0143089294 | 27.0118054967 | 9.3e-05 |
| `1e5` | 29.3704795837 | 27.0118054947 | **8.7e-02** |
| `1e7` | **1.7268073559** | 27.0118055903 | **9.4e-01** |

At `offset=1e7` the metric returns `1.73` where the correct value is `27.01`. The clustering structure is identical in every row — only the origin moved.

Off-origin coordinates are ordinary: geographic positions, epoch timestamps, unnormalised sensor readings. A user who reaches for `float64` precisely because their data has a large dynamic range currently gets it silently discarded.

### Cause

Buffers are allocated without `dtype`, so they default to `float32`:

`functional/clustering/davies_bouldin_score.py`:

```python
intra_dists = torch.zeros(num_labels, device=data.device)          # float32
centroids = torch.zeros((num_labels, dim), device=data.device)     # float32
...
centroids[k] = cluster_k.mean(dim=0)                               # float64 mean, narrowed on assignment
intra_dists[k] = (cluster_k - centroids[k]).pow(2.0).sum(dim=1).sqrt().mean()
```

Writing a `float64` centroid into the `float32` buffer rounds it to about 7 significant digits. At `offset=1e7` the stored centroid is wrong by roughly `1e0`, while the true intra-cluster spread is roughly `0.3` — so `cluster_k - centroids[k]` is dominated by the rounding error rather than by the data.

`functional/clustering/calinski_harabasz_score.py`:

```python
between_cluster_dispersion = torch.tensor(0.0, device=data.device)  # float32
within_cluster_dispersion = torch.tensor(0.0, device=data.device)   # float32
```

Here only the running accumulation is narrowed — the per-cluster terms are still computed in the input dtype — so the damage is bounded at about `1e-7` relative rather than being catastrophic. It is still a silent `float32` ceiling on a `float64` input.

Both also return `torch.tensor(..., dtype=torch.float32)` explicitly from their degenerate branches, which would stay inconsistent even after the buffers are fixed.

### Expected behavior

The accumulators should be allocated in the input dtype, so `float64` in gives `float64` out and the computation is actually carried out at the requested precision — matching `dunn_index`.

### Environment

- TorchMetrics: `master` (commit 37a80809)
- PyTorch: 2.13.0+cpu, Python 3.13, Windows

### Related

Same class of problem as #3465 (`spearman_corrcoef` / `kendall_rank_corrcoef`) but a different mechanism — that one is integer division falling back to the default dtype, this one is untyped buffer allocation. Reporting separately since the fix and the blast radius are different.

I have a fix and regression tests ready and am happy to open a PR.

VenishPaneliya:
## What does this PR do?

Fixes #3467

`davies_bouldin_score` and `calinski_harabasz_score` allocated their accumulators without a `dtype`, so the buffers were `float32` whatever the input was. `dunn_index` — same family, same signature — preserves the input dtype, so the clustering metrics disagreed with each other.

For **Davies-Bouldin this is a correctness bug, not a dtype-label one.** Writing a `float64` centroid into a `float32` buffer rounds it to about seven significant digits, so `cluster_k - centroids[k]` ends up dominated by that rounding as soon as the data sits away from the origin:

| data offset | returned | correct | relative error |
|---|---|---|---|
| `0` | 27.0118026733 | 27.0118054968 | 1.0e-07 |
| `1e3` | 27.0143089294 | 27.0118054967 | 9.3e-05 |
| `1e5` | 29.3704795837 | 27.0118054947 | **8.7e-02** |
| `1e7` | **1.7268073559** | 27.0118055903 | **9.4e-01** |

The clustering structure is identical in every row — only the origin moved. At `offset=1e7` the metric returns `1.73` where the answer is `27.01`.

Off-origin coordinates are ordinary: geographic positions, epoch timestamps, unnormalised sensor readings. A user reaching for `float64` *because* their data has a large dynamic range had it silently discarded.

Calinski-Harabasz narrows only the running accumulation — the per-cluster terms are still computed in the input dtype — so its error stayed near `float32` epsilon (~1e-7 relative). Still a silent `float32` ceiling on a `float64` input, but not catastrophic.

## Changes

```python
# davies_bouldin_score.py
- intra_dists = torch.zeros(num_labels, device=data.device)
- centroids = torch.zeros((num_labels, dim), device=data.device)
+ intra_dists = torch.zeros(num_labels, device=data.device, dtype=data.dtype)
+ centroids = torch.zeros((num_labels, dim), device=data.device, dtype=data.dtype)

# calinski_harabasz_score.py
- between_cluster_dispersion = torch.tensor(0.0, device=data.device)
- within_cluster_dispersion = torch.tensor(0.0, device=data.device)
+ between_cluster_dispersion = torch.tensor(0.0, device=data.device, dtype=data.dtype)
+ within_cluster_dispersion = torch.tensor(0.0, device=data.device, dtype=data.dtype)
```

Both degenerate branches also returned an explicit `dtype=torch.float32` scalar; those now follow the input dtype too, so the return type is consistent on every path.

**`float32` inputs are unaffected** — they already allocated `float32` buffers, so no value moves.

## Tests

New `tests/unittests/clustering/test_clustering_dtype.py`:

- dtype preservation across all three data-based clustering metrics;
- **translation invariance** for Davies-Bouldin and Calinski-Harabasz — shifting the data must not change a score defined purely by relative geometry. This is the test that pins the real regression;
- Davies-Bouldin matched against an independent `float64` implementation of the same formula;
- class-based wrappers agreeing with the functional results;
- `float32` behaviour explicitly unchanged.

On the tolerances: subtracting a centroid from off-origin data cancels leading digits, so exact translation invariance is **not** achievable even in `float64`. Measured limits are `2.6e-12` at `offset=1e3`, `2.9e-10` at `1e5` and `5.2e-08` at `1e7`, so the tolerance scales with the offset rather than pretending the error is zero. The bug produced a 94% error at `1e7`, orders of magnitude outside those bounds, so the test still catches it decisively.

Verified the tests actually catch the bug — against unpatched `src/`:

```
$ pytest tests/unittests/clustering/test_clustering_dtype.py -q
10 failed, 12 passed in 6.13s
```

With the fix:

```
$ pytest tests/unittests/clustering/test_clustering_dtype.py -q
22 passed in 7.12s
```

Full clustering suite, compared against `master` in the same environment:

```
master : 10 failed, 187 passed, 8 skipped
this PR:  0 failed, 197 passed, 8 skipped
```

The 10 are exactly this PR's own tests. Nothing else in the suite moves.

Lint, types and doctests:

```
$ ruff check <changed files>                        All checks passed!
$ ruff format --check <changed files>               3 files already formatted
$ mypy <changed source files>                       Success: no issues found in 2 source files
$ pytest --doctest-modules <changed source files>   2 passed
```

## Related

Same *class* of problem as #3465 / #3466 (`spearman_corrcoef` and `kendall_rank_corrcoef`), but a different mechanism — that one is integer division falling back to the default dtype, this one is untyped buffer allocation. Kept separate because the fix and the blast radius differ.

## Before submitting

- [x] Was this **discussed/agreed** via a GitHub issue? (#3467)
- [x] Did you make sure to **update the docs**? (n/a — no public API or signature change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed `Metric` ignoring an active `torch.device` context manager on torch 2.3-2.7 ([#3448](https://github.com/Lightning-AI/torchmetrics/pull/3448))
 
+- Fixed `davies_bouldin_score` returning an incorrect value for `float64` data far from the origin, and `calinski_harabasz_score` accumulating at `float32`, by allocating their buffers in the input dtype ([#3468](https://github.com/Lightning-AI/torchmetrics/pull/3468))
+
 
 ---
 
--- a/src/torchmetrics/functional/clustering/calinski_harabasz_score.py
+++ b/src/torchmetrics/functional/clustering/calinski_harabasz_score.py
@@ -48,8 +48,9 @@ def calinski_harabasz_score(data: Tensor, labels: Tensor) -> Tensor:
     _validate_intrinsic_labels_to_samples(num_labels, num_samples)
 
     mean = data.mean(dim=0)
-    between_cluster_dispersion = torch.tensor(0.0, device=data.device)
-    within_cluster_dispersion = torch.tensor(0.0, device=data.device)
+    # allocated in the input dtype so a float64 input is not silently accumulated at float32 precision
+    between_cluster_dispersion = torch.tensor(0.0, device=data.device, dtype=data.dtype)
+    within_cluster_dispersion = torch.tensor(0.0, device=data.device, dtype=data.dtype)
     for k in range(num_labels):
         cluster_k = data[labels == k, :]
         mean_k = cluster_k.mean(dim=0)
@@ -57,5 +58,5 @@ def calinski_harabasz_score(data: Tensor, labels: Tensor) -> Tensor:
         within_cluster_dispersion += ((cluster_k - mean_k) ** 2).sum()
 
     if within_cluster_dispersion == 0:
-        return torch.tensor(1.0, device=data.device, dtype=torch.float32)
+        return torch.tensor(1.0, device=data.device, dtype=data.dtype)
     return between_cluster_dispersion * (num_samples - num_labels) / (within_cluster_dispersion * (num_labels - 1.0))
--- a/src/torchmetrics/functional/clustering/calinski_harabasz_score.py
+++ b/src/torchmetrics/functional/clustering/calinski_harabasz_score.py
@@ -41,6 +41,11 @@ def calinski_harabasz_score(data: Tensor, labels: Tensor) -> Tensor:
     """
     _validate_intrinsic_cluster_data(data, labels)
 
+    # the squared accumulations below overflow/lose precision in half, so promote those to float32;
+    # float32/float64 are kept as-is
+    if data.dtype not in (torch.float32, torch.float64):
+        data = data.to(torch.float32)
+
     # convert to zero indexed labels
     unique_labels, labels = torch.unique(labels, return_inverse=True)
     num_labels = len(unique_labels)
--- a/src/torchmetrics/functional/clustering/davies_bouldin_score.py
+++ b/src/torchmetrics/functional/clustering/davies_bouldin_score.py
@@ -47,8 +47,10 @@ def davies_bouldin_score(data: Tensor, labels: Tensor) -> Tensor:
     num_samples, dim = data.shape
     _validate_intrinsic_labels_to_samples(num_labels, num_samples)
 
-    intra_dists = torch.zeros(num_labels, device=data.device)
-    centroids = torch.zeros((num_labels, dim), device=data.device)
+    # allocated in the input dtype: a float32 centroid buffer rounds the stored centre to ~7 significant digits, so
+    # `cluster_k - centroids[k]` catastrophically cancels for data that is not centred near the origin
+    intra_dists = torch.zeros(num_labels, device=data.device, dtype=data.dtype)
+    centroids = torch.zeros((num_labels, dim), device=data.device, dtype=data.dtype)
     for k in range(num_labels):
         cluster_k = data[labels == k, :]
         centroids[k] = cluster_k.mean(dim=0)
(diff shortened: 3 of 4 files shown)


VenishPaneliya:
You were right to flag this, and it was worse than a dtype change — it was a hard regression. On `master`, half inputs are promoted to float32 by the default buffer dtype and work fine; with my change `davies_bouldin_score` raised outright, because `torch.cdist` has no half kernel:

```
>>> davies_bouldin_score(torch.randn(60, 4, dtype=torch.float16), labels)
NotImplementedError: "cdist" not implemented for 'Half'
```

against `master`:

```
BASE torch.float16:  OK -> torch.float32 8.0390
BASE torch.bfloat16: OK -> torch.float32 10.4815
```

So I took your second option: both metrics now promote anything that isn't float32/float64 up to float32, which restores the old behaviour for half while keeping the float64 preservation this PR is actually for.

```
dtype        davies_bouldin        calinski_harabasz
float16      float32 8.0391         float32 0.5619
bfloat16     float32 10.4840        float32 0.5797
float32      float32 9.3202         float32 1.0359
float64      float64 6.3424         float64 1.0680
```

float16 lands on 8.0391 against master's 8.0390, so the promoted path reproduces the previous result.

Added `test_half_precision_is_computed_in_float32` covering both dtypes on both metrics. I scoped it to the two functions this PR touches — `dunn_index` returns the input dtype for half and I haven't changed it, so I didn't want to assert a policy for it here. If you'd rather these metrics reject half instead of promoting it, say so and I'll swap the promotion for an explicit error, but promoting matches what users get on master today.

`tests/unittests/clustering/` passes (24 in the dtype file, 12 in the two metric files) and ruff is clean.

PR #3470, opened 2026-08-20
Fix NMI/AMI returning 0.0 instead of 1.0 for two single-cluster labelings

### LINKED ISSUE #3469: NMI/AMI return 0.0 instead of 1.0 for two single-cluster labelings
## Bug

`normalized_mutual_info_score` and `adjusted_mutual_info_score` return `0.0` for two identical single-cluster labelings, where scikit-learn (the documented reference for these metrics) returns `1.0`, a perfect match.

## Repro

```python
import torch, numpy as np
import sklearn.metrics as skm
import torchmetrics.functional.clustering as tmc

a = np.array([0, 0, 0, 0]); ta = torch.tensor(a)
print(float(tmc.normalized_mutual_info_score(ta, ta)), skm.normalized_mutual_info_score(a, a))  # 0.0 vs 1.0
print(float(tmc.adjusted_mutual_info_score(ta, ta)), skm.adjusted_mutual_info_score(a, a))        # 0.0 vs 1.0
```

## Why

Both labelings have zero entropy, so they trivially agree. scikit-learn treats this limit case as a perfect match and returns `1.0`. torchmetrics falls through to the `mutual_info == 0` short-circuit and returns `0.0`. The asymmetric case (one single cluster, the other not) already returns `0.0` in both libraries and is correct.

Environment: torchmetrics 1.9.0 / current `master`.

I have a fix ready and will open a PR referencing this issue.

AlejandroCoronadoN:
Fixes #3469.

`normalized_mutual_info_score` and `adjusted_mutual_info_score` return `0.0` when both `preds` and `target` are a single cluster (two zero-entropy labelings). scikit-learn, the documented reference for these metrics, returns `1.0` for this limit case because the two labelings trivially agree.

Repro:

```python
import torch, numpy as np
import sklearn.metrics as skm
import torchmetrics.functional.clustering as tmc

a = np.array([0, 0, 0, 0]); ta = torch.tensor(a)
print(float(tmc.normalized_mutual_info_score(ta, ta)), skm.normalized_mutual_info_score(a, a))  # 0.0 vs 1.0
print(float(tmc.adjusted_mutual_info_score(ta, ta)), skm.adjusted_mutual_info_score(a, a))        # 0.0 vs 1.0
```

Fix: return `1.0` early when both labelings have exactly one unique value, matching sklearn. For `adjusted_mutual_info_score` I also added the `check_cluster_labels(preds, target)` call that the other clustering metrics already use, so the early return does not bypass input validation (float inputs still raise, verified). No non-degenerate result changes.

Tests: added a `both_single_cluster` case parametrized over every `average_method` to both functional test files, using sklearn as the reference oracle. `ruff` is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed `normalized_mutual_info_score` and `adjusted_mutual_info_score` returning `0.0` instead of `1.0` for two single-cluster labelings ([#3470](https://github.com/Lightning-AI/torchmetrics/pull/3470))
 - Fixed malformed LaTeX in `CLIPScore` and `HausdorffDistance` docstring math so it renders correctly ([#3427](https://github.com/Lightning-AI/torchmetrics/pull/3427))
 
 
--- a/src/torchmetrics/functional/clustering/adjusted_mutual_info_score.py
+++ b/src/torchmetrics/functional/clustering/adjusted_mutual_info_score.py
@@ -21,6 +21,7 @@ from torchmetrics.functional.clustering.utils import (
     _validate_average_method_arg,
     calculate_entropy,
     calculate_generalized_mean,
+    check_cluster_labels,
 )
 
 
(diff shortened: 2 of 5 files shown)


justusschock (approved):

AlejandroCoronadoN:
Thanks for the review, @justusschock. This has drifted a bit behind master; happy to rebase it if that helps it land, and glad to make any other changes you need.

Borda (approved):

PR #3473, opened 2026-08-20
fix(wrappers): validate `ClasswiseWrapper` labels

### LINKED ISSUE #3472: ClasswiseWrapper silently drops values for invalid labels
## Bug

`ClasswiseWrapper` converts a metric tensor into a dictionary with `zip(self.labels, x)`. If the number of labels does not match the number of metric values, `zip` silently truncates the longer input. Duplicate labels also overwrite earlier dictionary entries. Both cases can make class metrics disappear without any error.

## Reproduction

```python
import torch

from torchmetrics.classification import MulticlassAccuracy
from torchmetrics.wrappers import ClasswiseWrapper

metric = ClasswiseWrapper(
    MulticlassAccuracy(num_classes=3, average=None),
    labels=["cat", "dog"],
)

result = metric(
    torch.tensor([0, 1, 2]),
    torch.tensor([0, 1, 2]),
)
print(result)
```

The wrapped metric produces three class values, but the result contains only two:

```text
{"multiclassaccuracy_cat": tensor(1.), "multiclassaccuracy_dog": tensor(1.)}
```

Likewise, labels such as `["cat", "cat", "dog"]` produce only two dictionary keys because the second `cat` overwrites the first.

## Expected behavior

Custom labels should define a one-to-one mapping to the wrapped metric values. Mismatched label counts and duplicate labels should raise a clear `ValueError` instead of silently losing results.

## Environment

- TorchMetrics: current `master` at `e63e92a6`
- PyTorch: 2.13.0
- Python: 3.12.13
- macOS

tandede:
## What does this PR do?

Fixes #3472.

`ClasswiseWrapper` previously built its output with `zip(self.labels, x)`. A shorter or longer label list therefore silently truncated one side, while duplicate labels overwrote dictionary entries. In both cases class-level metric values could disappear without an error.

This change:

- rejects duplicate custom labels during initialization;
- checks the label count against the wrapped metric output before constructing the result dictionary;
- documents the one-to-one label requirement;
- adds regressions for shorter, longer, and duplicate label lists.

The count check lives in the shared output conversion path because generic wrapped metrics do not necessarily expose their output length during initialization. It therefore protects both `forward()` and `compute()` without restricting which metric types can be wrapped.

<details>

- [x] Was this discussed/agreed via a Github issue? (#3472)
- [x] Did you read the contributor guideline?

</details>

## Validation

- Before the source fix, the three new regression cases failed because no `ValueError` was raised.
- `pytest -q tests/unittests/wrappers/test_classwise.py tests/unittests/wrappers/test_tracker.py` — 39 passed, 1 skipped.
- Wrapper suite excluding the optional feature-sharing dependency — 124 passed, 10 skipped.
- `pre-commit run --files src/torchmetrics/wrappers/classwise.py tests/unittests/wrappers/test_classwise.py` — all hooks passed.

## Did you have fun?

Yes.

### DIFF
--- a/src/torchmetrics/wrappers/classwise.py
+++ b/src/torchmetrics/wrappers/classwise.py
@@ -38,7 +38,8 @@ class ClasswiseWrapper(WrapperMetric):
     Args:
         metric: base metric that should be wrapped. It is assumed that the metric outputs a single
             tensor that is split along the first dimension.
-        labels: list of strings indicating the different classes.
+        labels: list of unique strings indicating the different classes. The number of labels must match the number of
+            values returned by the wrapped metric.
         prefix: string that is prepended to the metric names.
         postfix: string that is appended to the metric names.
 
@@ -130,6 +131,8 @@ class ClasswiseWrapper(WrapperMetric):
 
         if labels is not None and not (isinstance(labels, list) and all(isinstance(lab, str) for lab in labels)):
             raise ValueError(f"Expected argument `labels` to either be `None` or a list of strings but got {labels}")
+        if labels is not None and len(set(labels)) != len(labels):
+            raise ValueError(f"Expected argument `labels` to contain unique values, but got {labels}")
         self.labels = labels
 
         if prefix is not None and not isinstance(prefix, str):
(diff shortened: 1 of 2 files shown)


justusschock (approved):

tandede:
I retriggered CI because the previous package workflow was cancelled immediately and linkcheck only hit external-site timeouts. The four new fork workflows are now waiting for maintainer approval to run (`action_required`); I do not have repository permission to approve them.

Borda (approved):

PR #3482, opened 2026-08-31
Forward `batch_size` in `PerceptualPathLength.compute`

VenishPaneliya:
## What does this PR do?

`PerceptualPathLength.__init__` documents `batch_size`, validates it, and stores it as `self.batch_size` — but that attribute is never read again. `compute()` forwards ten arguments to the functional `perceptual_path_length` and `batch_size` is not one of them, so the functional default of 64 always wins over the class default of 128.

Measured with a counting `sim_net` at `num_samples=256`:

| requested `batch_size` | class: passes / effective | functional: passes / effective |
|---|---|---|
| 128 | 4 / **64** | 2 / 128 |
| 256 | 4 / **64** | 1 / 256 |

The class ignores the argument entirely. The fix is to forward it:

```python
 return perceptual_path_length(
     generator=self.generator,
     num_samples=self.num_samples,
     conditional=self.conditional,
+    batch_size=self.batch_size,
     interpolation_method=self.interpolation_method,
```

`batch_size` only chunks the sampling loop before `torch.cat`, so the returned values are unchanged — what changes is how many generator and LPIPS forward passes run. At the library default `num_samples=10_000`, a user asking for 1024 currently gets 157 batches instead of 10.

Two things worth flagging:

- The default *effective* batch size goes 64 → 128, so peak memory at default settings roughly doubles. 128 is the documented and validated default, so honouring it looks like the correct behaviour, but it is a real change for anyone relying on the current default.
- Different batch shapes reassociate the float32 sum, so results shift by ~6e-7 relative. That is well inside the existing reference-test tolerance and the suite is unchanged.

## Testing

Added `test_batch_size_is_forwarded`, which drives the metric with a `sim_net` that records the batch dimension it receives and asserts the largest matches the requested `batch_size`.

- before the fix: `assert 64 == 32` — fails
- after the fix: passes
- full `tests/unittests/image/test_perceptual_path_length.py`: 17 passed, 1 skipped (unchanged from master)

<details>

- [x] Did you make sure to **update the docs**? (no doc change needed — the docstring already describes the intended behaviour)

</details>

<details>
  <summary>PR review</summary>

</details>

## Did you have fun?

Yes — chasing an attribute that was assigned and then quietly forgotten is a fun kind of bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed malformed LaTeX in `CLIPScore` and `HausdorffDistance` docstring math so it renders correctly ([#3427](https://github.com/Lightning-AI/torchmetrics/pull/3427))
 
 
+- Fixed `PerceptualPathLength` ignoring the `batch_size` argument, which left it running at the functional default of 64 ([#3482](https://github.com/Lightning-AI/torchmetrics/pull/3482))
+
+
 - Fixed `Metric` ignoring an active `torch.device` context manager on torch 2.3-2.7 ([#3448](https://github.com/Lightning-AI/torchmetrics/pull/3448))
 
 
--- a/src/torchmetrics/image/perceptual_path_length.py
+++ b/src/torchmetrics/image/perceptual_path_length.py
@@ -172,6 +172,7 @@ class PerceptualPathLength(Metric):
             generator=self.generator,
             num_samples=self.num_samples,
             conditional=self.conditional,
+            batch_size=self.batch_size,
             interpolation_method=self.interpolation_method,
             epsilon=self.epsilon,
             resize=self.resize,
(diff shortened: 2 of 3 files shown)


VenishPaneliya:
Good catches, all three applied.

- Added a `_TORCHVISION_AVAILABLE` skip — `PerceptualPathLength.__init__` raises `ModuleNotFoundError` without torchvision, so the test would have hard-failed rather than skipped on a minimal install.
- `_CountingSimNet` now returns `(N, 1, 1, 1)` on `img1.device` with `img1.dtype`, matching what LPIPS actually hands back.
- `_TinyGenerator` now emits `255 * torch.rand(...)`. That one was a real contract violation on my side: the functional documents generator output as scaled to [0, 255] and immediately does `2 * (out / 255) - 1`, so my [0, 1] generator was feeding it values it would have squashed to roughly [-1, -0.99].

Re-checked the regression still holds both ways after the changes: with the fix the test passes and `batch_sizes` is all 32s; with `batch_size=self.batch_size` removed from `compute()` it fails with `[64]`, i.e. one unbatched pass.

PR #3483, opened 2026-09-01
refactor: drop the dead torch<2.6 compatibility paths

bhimrazy:
## What does this PR do?

Follow-up to #3449. With the floor at torch **2.6**, five version flags in `utilities/imports.py` are now constants. This removes them and the branches they guarded — no behaviour change on any supported torch.

- `Metric` no longer caches `self._TORCH_GREATER_EQUAL_2_1`; the `_sync_dist` empty-list guard becomes unconditional.
- `_cumsum` had degenerated into a pass-through, so it is removed and its 8 call sites use `torch.cumsum`.
- ARNIQA drops its unreachable `requires PyTorch >= 2.2.0` guard.
- ~60 test `skipif`/`xfail` guards that can no longer fire are gone.

<details>
<summary><b>Flags removed</b></summary>

| Flag | Evaluated to | Sites |
|---|---|---|
| `_TORCH_GREATER_EQUAL_2_1` | always `True` | 96 |
| `_TORCH_GREATER_EQUAL_2_2` | always `True` | 18 |
| `_TORCH_GREATER_EQUAL_2_3` | always `True` | 0 — defined, never referenced |
| `_TORCH_GREATER_EQUAL_2_5` | always `True` | 5 |
| `_TORCH_LESS_THAN_2_6` | always `False` | 5 |

`_TORCH_GREATER_EQUAL_2_8` stays — still live in `Metric.__init__`.

</details>

<details>
<summary><b>Two tests removed rather than un-gated</b></summary>

Each asserted that torch still *lacks* a feature, and each had become permanently skipped/xfailed on 2.6:

- `test_cumsum_still_not_supported` — "If this test begins to pass, it means newer Pytorch versions support this and we can drop internal support." It does, so this drops it.
- `test_half_precision_top_k_cpu_raises_error` — `skipif(_TORCH_GREATER_EQUAL_2_2)`, i.e. never ran. cpu+half `topk` works on current torch.

</details>

<details>
<summary><b>Left in place deliberately</b></summary>

- **`TorchMetricsUserWarning`** is now unused inside `src/` — its only warn site was in `_cumsum`. Kept, since it is public in `torchmetrics.utilities.exceptions`.
- **`_top_k_with_half_precision_support`** looks dead by the same argument, but its `argsort(stable=True).flip()` tie-breaking differs from `topk`, so removing it would change half-precision results rather than just delete dead code. Worth its own PR.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed support for PyTorch 2.0-2.5, the minimum required version is now 2.6 ([#3449](https://github.com/Lightning-AI/torchmetrics/pull/3449))
 
 
+- Removed the dead `torch<2.6` compatibility branches left behind by the 2.6 floor, including the internal `_cumsum` workaround ([#3450](https://github.com/Lightning-AI/torchmetrics/pull/3450))
+
+
 ### Fixed
 
 - Fixed malformed LaTeX in `CLIPScore` and `HausdorffDistance` docstring math so it renders correctly ([#3427](https://github.com/Lightning-AI/torchmetrics/pull/3427))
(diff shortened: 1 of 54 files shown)


bhimrazy (commented):
Checked

ethanwharris (approved):

tchaton (approved):

justusschock (approved):

PR #3485, opened 2026-09-08
Return zero V-measure for zero-information clusterings

### LINKED ISSUE #3484: V-measure returns 1 for independent nontrivial clusterings
For two independent, nontrivial clusterings, both homogeneity and completeness are zero, but `v_measure_score` currently returns `1.0`.

```python
import torch
from torchmetrics.functional.clustering import v_measure_score

preds = torch.tensor([0, 1, 0, 1])
target = torch.tensor([0, 0, 1, 1])

print(v_measure_score(preds, target))  # tensor(1.)
```

scikit-learn returns `0.0` for the same labels. That also matches the weighted harmonic mean definition: when both component scores are zero, V-measure should be zero.

This reproduces on current `master` (`8d008de1`) with Python 3.12.13 and PyTorch 2.13.0. I have a small fix and regression test ready. I also checked the edge case with a small Lean formalization and can attach it if useful :)

Robertboy18:
## What does this PR do?

Fixes #3484

I was checking V-measure edge cases and noticed that the zero-information branch returned `1`, even though both homogeneity and completeness are zero. This changes that branch to return `0`, matching scikit-learn and the metric definition.

I added regression coverage for both the functional and class APIs with `beta` values `0.5`, `1.0`, and `2.0`. I also checked every partition pair through five samples against scikit-learn (8,877 cases); the full clustering suite passes (`178 passed, 8 skipped`), and Ruff is clean.

I also wrote a small Lean formalization of this edge case and can attach it if useful :)

### DIFF
--- a/src/torchmetrics/functional/clustering/homogeneity_completeness_v_measure.py
+++ b/src/torchmetrics/functional/clustering/homogeneity_completeness_v_measure.py
@@ -110,5 +110,5 @@ def v_measure_score(preds: Tensor, target: Tensor, beta: float = 1.0) -> Tensor:
     """
     completeness, homogeneity = _completeness_score_compute(preds, target)
     if homogeneity + completeness == 0.0:
-        return torch.ones_like(homogeneity)
+        return torch.zeros_like(homogeneity)
     return (1 + beta) * homogeneity * completeness / (beta * homogeneity + completeness)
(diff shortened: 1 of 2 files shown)


justusschock (approved):

Robertboy18:
Thanks for reviewing this! I refreshed the branch against master and reran the clustering suite: 178 passed, 8 skipped. Ruff is clean too. Could you take another look at merging once the new checks finish? :)

PR #3491, opened 2026-09-08
Preserve group IDs in fairness metrics

### LINKED ISSUE #3489: Group fairness metrics can relabel missing group IDs
When a batch does not contain every configured group, `BinaryGroupStatRates` currently compresses the group IDs.

```python
import torch
from torchmetrics.classification import BinaryGroupStatRates
from torchmetrics.functional.classification import binary_groups_stat_rates

preds = torch.tensor([1, 0])
target = torch.tensor([1, 0])
groups = torch.tensor([2, 2])

print(binary_groups_stat_rates(preds, target, groups, num_groups=3))
# {'group_0': tensor([0.5000, 0.0000, 0.5000, 0.0000])}

metric = BinaryGroupStatRates(num_groups=3)
metric.update(preds, target, groups)
print(metric.tp)  # tensor([1, 0, 0])
```

The observations belong to group `2`, but they are returned and accumulated as group `0`. Negative and out-of-range IDs are similarly relabeled, and documented multidimensional group tensors fail in `bincount`.

This reproduces on current `master` (`8d008de1`) with Python 3.12.13 and PyTorch 2.13.0. I have a fix and regression tests ready :)

Robertboy18:
## What does this PR do?

Fixes #3489

I was checking `BinaryGroupStatRates` with batches that do not contain every configured group and noticed the IDs get compressed. For example, a batch containing only group `2` updates `group_0`. The same path accepts invalid IDs and fails on documented multidimensional group tensors.

This keeps count slots aligned with `num_groups`, validates the configured ID range, and handles both per-sample and elementwise group tensors. I added regression coverage for the functional and class APIs.

The full group-fairness test file passes, along with pre-commit, deterministic CPU/CUDA checks, and 500 randomized partition cases. I can also add a small Lean/TorchLean model of the partition invariant if useful :)

### DIFF
--- a/src/torchmetrics/functional/classification/group_fairness.py
+++ b/src/torchmetrics/functional/classification/group_fairness.py
@@ -24,24 +24,20 @@ from torchmetrics.functional.classification.stat_scores import (
 )
 from torchmetrics.utilities import rank_zero_warn
 from torchmetrics.utilities.compute import _safe_divide
-from torchmetrics.utilities.data import _flexible_bincount
+from torchmetrics.utilities.data import _bincount
 
 
 def _groups_validation(groups: torch.Tensor, num_groups: int) -> None:
     """Validate groups tensor.
 
-    - The largest number in the tensor should not be larger than the number of groups. The group identifiers should
-    be ``0, 1, ..., (num_groups - 1)``.
+    - The group identifiers should be ``0, 1, ..., (num_groups - 1)``.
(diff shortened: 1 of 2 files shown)


Robertboy18:
I reran the group-fairness tests against the current branch: 46 passed, including the CUDA cases. Ruff is clean too. This is ready for review whenever someone has time :)

justusschock (approved):

